PR #35, opened 2016-03-13
Import seopostmeta from db

mrtorbert:
Added SEO meta import functionality


mrtorbert:
At the moment this is only title, description, and meta keywords but it should be tested at least for the major plugins/themes...

PR #147, opened 2016-04-07
All in one seo pack file pull into dev branch

michaeltorbert:
this needs to be tested first

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@
 .travis.yml
 .svn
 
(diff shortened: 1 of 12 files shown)


michaeltorbert:
@wpsmort Steve can you test the all_in_one_seo_pack-file branch?
It needs to be tested as both free and pro before we merge.

Basic general testing, along with being able to upgrade to pro from free and to pro from pro.

PR #240, opened 2016-04-26
Wpstandards

joelrd:
WP Standarization for admin\aioseop_module_class.php up to line 2588

### DIFF
--- a/admin/aioseop_module_class.php
+++ b/admin/aioseop_module_class.php
@@ -1504,141 +1504,223 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Module' ) ) {
 				return $image[0];
 		}
 
+		/**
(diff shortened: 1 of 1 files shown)


michaeltorbert:
@joelrd @amostajo There are still a lot of removed comments. Are these on purpose for a reason or was it just a mistake?

joelrd:
@michaeltorbert @amostajo Sorry, I just noticed that I didn't add all the previous comments missing from the last Pull-Request. I double checked and added them again. If there are other missing comments let me know. I am really sorry. Because I added a comment on every loop by my own, but I didn't know how it was structured and it was taking more time than expected. So I decided to delete them. Some of the comments were confused with those. I left this practice on a couple of files after. I will check them again and make sure that none of the comments are missing for the next files.

PR #834, opened 2017-03-22
Fixes #687. Adds compatibility class with WPML, global aioseop_home_u…

### LINKED ISSUE #687: XML Sitemap - Unsafe attempt to load URL when loading XSL file
Reported by Marco Tiggelaar via email today.

When using WPML with different TLDs and HTTPS, the XSL doesn't load because of this error:

Unsafe attempt to load URL https://www.example.com/sitemap.xsl from frame with URL https://www.example.nl/sitemap.xml. Domains, protocols and ports must match.

amostajo:
…rl() and refactores sitemap generator.

Solves compatibility issue with WPML by refactoring the usage of
"home_url" and "get_home_url" when generating sitemaps. Additionally it
extends the solution of compatibility with third party support by adding
an abtract class and integration with the plugin. Global function to
handle variations of "home_url" has been added as "aioseop_home_url".

### DIFF
--- a/all_in_one_seo_pack.php
+++ b/all_in_one_seo_pack.php
@@ -369,6 +369,11 @@ if ( ! function_exists( 'aiosp_action_links' ) ) {
 }
 
 if ( ! function_exists( 'aioseop_init_class' ) ) {
+	/**
+	 * Inits All-in-One-Seo plugin class.
+	 *
+	 * @since 2.3.13 Loads third party compatibility class.
+	 */
 	function aioseop_init_class() {
 		global $aiosp;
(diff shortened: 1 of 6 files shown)


michaeltorbert:
#687

wpsmort:
This works for the options in WPML for Different languages in directories and for A different domain per language.  However, it breaks if you select the option for Language name added as a parameter.

Neither the default language sitemap show or the other languages.  This is the URL for the XSL file for the default language:

http://stevemortiboy.comsitemap.xsl

This is the URL for the XSL file for the French language:

http://stevemortiboy.com/?lang=frsitemap.xsl

Issue #672, opened 2016-12-07
Snippet preview and post seo desc default don't filter out html


dawatorbert:
![screen shot 2016-12-06 at 7 32 09 pm](https://cloud.githubusercontent.com/assets/18234064/20950005/d5aa803e-bbea-11e6-9f9a-1e33b937eb02.png)

wpsmort:
This only affects WooCommerce product short descriptions.  It's because the Product Short Description was developed as the Post Excerpt field for Products and you can't use HTML in the Post Excerpt.  But now that they've added the TinyMCE editor to the Product Short Description you can use it like the Post Content field and put HTML and shortcodes in the Product Short Description.  

We therefore need to add code to parse any HTML added to the Product Short Description when using auto-generated meta descriptions before displaying the preview snippet.

wpsmort:
I tested this in pull request 835 and it was fixed.  I opened a new issue #837 for the same problem on the Social Settings tab.

stanislavsa:
I added fix for an issue #837 in the pull request 835. Also placed the code that filters out tags for snippet preview before the code that truncates snippet preview description so for now snippet preview and post seo description placeholder displays text with the same length.

wpsmort:
@stanislavsa I tested this further with more HTML tags but using different format buttons from the TinyMCE bar in the Product Short Description field.  I found that certain text formatting seems to break this.  I believe it's because the HTML markup is being included in the number of characters and once we reach 160 characters we don't display anything more in the Description field.  For example:

This is getting counted as 160 characters:
<span style="color: #ff0000;">Lorem</span> <span style="text-decoration: underline;">ipsum</span> dolor sit amet

But in fact, once you strip out the HTML there's only 27 characters of text there.

stanislavsa:
@wpsmort ok, I'l check one more time

stanislavsa:
@wpsmort yes, you was right. The number of symbols was incorrect. I figured out where the description truncates in the code and did some changes.

wpsmort:
@stanislavsa This works now, thanks.  Can you submit a new pull request and I'll mark it for testing.

wpsmort:
This is tested and is good.  Needs final code review.

PR #835, opened 2017-03-27
Snippet preview and post seo desc default don't filter out html #672

stanislavsa:
fixed.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -1245,7 +1245,8 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			$extra_title_len = strlen( str_replace( $replace_title, '', $title_format ) );
 		}
 
-		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( wp_strip_all_tags( $description ) ) );
+		$description = htmlspecialchars( wp_strip_all_tags( htmlspecialchars_decode( $description ) ) );
+		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( $description ) );
 		$extra_title_len = (int) $extra_title_len;
 		$args['value'] .= "<script>var aiosp_title_extra = {$extra_title_len};</script>";
 		$buf = $this->get_option_row( $args['name'], $args['options'], $args );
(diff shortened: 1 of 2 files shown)


aioseop_class.php
```diff
@@ -1245,7 +1245,8 @@ function custom_output_option( $buf, $args ) {
 			$extra_title_len = strlen( str_replace( $replace_title, '', $title_format ) );
 		}
 
-		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( wp_strip_all_tags( $description ) ) );
+		$description = htmlspecialchars( wp_strip_all_tags( htmlspecialchars_decode( $description ) ) );
+		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( $description ) );
```
aioseopbot: __[phpmd]__
* Avoid unused local variables such as '$url'.

wpsmort:
This is tested and I have confirmed that it works.  I have opened a new issue #837 for the same problem on the Social Settings tab.

aioseop_class.php
```diff
@@ -1245,7 +1246,7 @@ function custom_output_option( $buf, $args ) {
 			$extra_title_len = strlen( str_replace( $replace_title, '', $title_format ) );
 		}
 
-		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( wp_strip_all_tags( $description ) ) );
+		$args['value']   = sprintf( $args['value'], $title, esc_url( $url ), esc_attr( $description ) );
```
aioseopbot: __[phpmd]__
* Avoid unused local variables such as '$url'.

wpsmort:
I tested this further with more HTML tags but using different format buttons from the TinyMCE bar in the Product Short Description field.  I found that certain text formatting seems to break this.  I believe it's because the HTML markup is being included in the number of characters and once we reach 160 characters we don't display anything more in the Description field.  For example:

This is getting counted as 160 characters:
<span style="color: #ff0000;">Lorem</span> <span style="text-decoration: underline;">ipsum</span> dolor sit amet

But in fact, once you strip out the HTML there's only 27 characters of text there.

PR #838, opened 2017-03-29
#687 Adds WPML support for compatibility with sitemap rendering on different languages

amostajo:
Adds WPML support for compatibility with sitemap rendering on different languages. #687


wpsmort:
Please see my feedback in #687.  I tested today and this currently doesn't work for the one Language URL Format setting of Language name added as a parameter

wpsmort:
@amostajo Is this PR ready for testing or do you need to create a new PR?

wpsmort:
This is tested with all three language options in WPML and confirmed that all three now work.  This needs a code review.

michaeltorbert:
Also this fixes #865

Issue #253, opened 2016-04-28
Importer Exporter Module - Remove options updated message when compromised file is imported


michaeltorbert:
When a compromised file is attempted to be imported, it should fail, and gives this warning. However, it stills says the options have been updated. 

1) Ensure no options are updated
2) See why there's a message saying so

<img width="838" alt="screen shot 2016-04-28 at 5 28 36 pm" src="https://cloud.githubusercontent.com/assets/17364998/14902072/bdc2c3ec-0d66-11e6-8501-61e368b6a63b.png">

michaeltorbert:
This is going to need to come when we redo the options.

stanislavsa:
I checked this issue and for sure that  options are not updated because in that case triggers an exception before the code that update settings and only warning displayed.
I added a new hook for this case to prevent displaying options updated message.
Also I got next issue when message displayed  
![importer exporter web technologies wordpress](https://cloud.githubusercontent.com/assets/22232968/24507219/95f10404-1568-11e7-8622-038550e4e397.png)
That is because esc_html function was added in some later commit and all tags was displayed as html entities. So I replaced it with wp_kses.

michaeltorbert:
@stanislavsa Are you still working on this or is it ready for a PR?

stanislavsa:
@michaeltorbert It's ready for a PR.

wpsmort:
This is tested and is good.  Needs a code review. PR #846

PR #846, opened 2017-03-31
Issue #253 - Importer Exporter Module - Remove options updated message when compromised file is imported

stanislavsa:
This is for issue #253


wpsmort:
This is tested and is good.  Needs a code review.

Issue #97, opened 2016-03-30
Performance Module - Improve the debug information that is emailed to us


wpsmort:
Users should be able to send us debug info with plugins active, ALL settings, all server info, theme active, and certain WP settings, and URL

michaeltorbert:
@nishant981 Conversations, questions, comments, updates, screenshots, etc should be here.

nishant981:
Hello @michaeltorbert


I have added one button in admin panel under Performance tab in All in One SEO plugin.
Currently I am using my email ID on which the mail will come. Please share your email ID so that I can replay it.

As you have mentioned in tasks, we have display WP settings, URL, Active/Inactive Plugins and Active Theme in email.

If there is anything you need to add in email then let us know. We will add those also.
Please find the attached screenshot for reference.

Let me know in case of any concern
If everything fine then i will upload the code on github as well

Regards

![screenshot_1](https://cloud.githubusercontent.com/assets/13958896/14308893/c1a87316-fbf6-11e5-892b-d7184034e1c3.png)

wpsmort:
Hi @nishant981, can you send an example email to me steve@semperfiwebdesign.com

michaeltorbert:
@wpsmort I recommend not posting your email here for scrapers to pick up. ;)

nishant981:
Hello @wpsmort

The information received via  email mentioned in attached screenshot.
Please review the same and let me know is it fine or not.

Regards
![all_in_one_seo_screenshot](https://cloud.githubusercontent.com/assets/13958896/14344086/1f33298c-fcc3-11e5-983c-644c853aee82.png)

wpsmort:
Hi @nishant981, I don't see any difference between your email and our current debug email.  We need to improve on the current debug email by including all settings from All in One SEO Pack.  

We have the ability to export all settings using our Importer Exporter module which you can activate under Feature Manager.  If you activate that module and perform an export you'll get a file called settings_aioseop.ini.  Take a look at that file, that's what we want included in the debug email.

Here's a PDF of the settings exported from my test site:

[settings_aioseop.pdf](https://github.com/semperfiwebdesign/all-in-one-seo-pack/files/208620/settings_aioseop.pdf)

wpsmort:
@stanislavsa I tested this but when I sent the email I got this notice:

Notice: Undefined index: sfwd_update_check in all-in-one-seo-pack-issue-97/modules/aioseop_performance.php on line 291

[...]

PR #851, opened 2017-04-03
Issue #97

stanislavsa:
added attachment file to debug email with all plugin settings and posts seo data.


wpsmort:
@stanislavsa I tested this but when I sent the email I got this notice:

Notice: Undefined index: sfwd_update_check in all-in-one-seo-pack-issue-97/modules/aioseop_performance.php on line 291

I checked and this notice happens in the current live release so it's not new.  I've opened issue #864 for this and assigned it to you.

Can you also include the following information in the debug email:

Site Title
Site Language
Front Page Displays
Search Engine Visibility
Permalink Setting

PR #867, opened 2017-04-07
Issue #97 - Performance Module - Improve the debug information that is emailed to us

stanislavsa:
This is for issue #97 

Added new information in the debug email.


wpsmort:
This is tested and is good.  Needs a code review.

Issue #902, opened 2017-05-22
"Edit SEO" in admin bar shows same post ID no matter the actual post.


michaeltorbert:
Reported by energytherapy.biz via email.
I've logged in and confirmed the issue on their site, but am unable to reproduce it elsewhere. I'm creating a clone of their site for testing.

michaeltorbert:
The theme creates a query improperly, so to fix the URL I'm changing how we get the post info.

For testing, first verify the problem exists with the existing theme and plugins installed (the various woocommerce plugins have to be installed for the theme to work). The "Edit SEO" post should link to the same URL (look at the post ID) for different posts/pages.
Then verify that this commit in the issue-902 branch fixes it. 

Also, make sure we didn't break the URL for normal themes.

wpsmort:
I have tested this and confirmed that it fixes the problem on the customer's site and does not break anything with regards to a normal theme.

PR #904, opened 2017-05-22
accommodate themes making bad queries #902

michaeltorbert:


wpsmort:
I have tested this and confirmed that it fixes the problem on the customer's site and does not break anything with regards to a normal theme.

Issue #914, opened 2017-05-31
Single blank space in og:description when non-breaking space it present


wpsmort:
This relates to #901.  Whilst testing that PR, I found that when you have an excerpt that ends with a non-breaking space (&nbsp;) then the non-breaking space is stripped but replaced with a single blank space.  I notified @amostajo about this so he could investigate.  

PR #901 has now been closed and merged into the development branch hence this new issue which will require a new PR.

amostajo:
@wpsmort please test this.

wpsmort:
This is tested and is fixed.  It is ready for a code review - PR #915

Issue #910, opened 2017-05-30
Manually entered HTML markup is not stripped from OG description 


wpsmort:
When you enter HTML markup in the OG Description field on a post or page, the markup is output to the meta tag in the source code.  

This was identified whilst testing PR's #839 and #901.

Here is the HTML markup I used in the OG Description field:

```
<strong>Bold text</strong> <span style="text-decoration: underline;">Underlined text</span>
<ul><li>List item</li></ul>
<a href="https://google.com/" target="_blank" rel="noopener noreferrer">Link text</a>
```

Here's a screenshot of how it looks in PR #839, notice how the meta description tag is correct:

<img width="1286" alt="screen shot 2017-05-30 at 3 37 18 pm" src="https://cloud.githubusercontent.com/assets/17548525/26601475/efcce14a-454d-11e7-9104-785da2453c12.png">

Here's a screenshot of how it looks in PR #901, notice how the meta description tag is correct except it's missing the link text:

<img width="1264" alt="screen shot 2017-05-30 at 3 37 35 pm" src="https://cloud.githubusercontent.com/assets/17548525/26601479/f61a236e-454d-11e7-83bc-b0ae3b91996e.png">

Also, note that with PR #901, the HTML is being output to the page, see screenshot:

<img width="803" alt="screen shot 2017-05-30 at 3 38 49 pm" src="https://cloud.githubusercontent.com/assets/17548525/26601514/1aea26ee-454e-11e7-9be7-d6bbee091a30.png">

amostajo:
@wpsmort ready for your testing.

wpsmort:
@michaeltorbert This is tested and is good.  The only issue I see is that we are stripping the link text, when we should be only stripping the link HTML markup.  For example,

This markup gets removed from the og:description meta tag:
<a href="https://google.com/" target="_blank">Link text</a>

When we should really just strip the HTML and leave the text in between the anchor tags.

I suggest that we merge the code in PR #915 into the development branch for this release and I'll open a separate issue for the stripping of link text.

michaeltorbert:
@wpsmort Agreed.

PR #915, opened 2017-05-31
Fixes for #910 and #914

amostajo:
Stripping line added to `aioseip_description` filter. All descriptions
are being stripped from HTML markup.


wpsmort:
This is tested and is good.  It is ready for code review.  

See my note in issue #910 regarding stripping link text.

Issue #916, opened 2017-06-01
Manually entered links are stripped from descriptions


wpsmort:
This is related to #910 

When you manually enter a link in the SEO Description or OG Description fields, the HTML and link text is stripped.  We should only strip the HTML, not the link text.  For example:

`<a href="https://google.com/" target="_blank">Link text</a>`

should become:

Link text

amostajo:
@wpsmort ready for review.

wpsmort:
This is tested and is good.  PR #927 is ready for code review.

PR #927, opened 2017-06-02
#916 Strips links except anchor texts on descriptions.

amostajo:
This is for issue #916 

Added rule to `aioseop_description` so HTML links are stripped by their
anchor text is kept.


wpsmort:
This is tested and is good.  This PR is ready for code review.

Issue #932, opened 2017-06-09
Broken filter in version 2.3.13 


wpsmort:
Reported by Marco Ricca by email

Marco uses the filter below that we provided him back in 2014.  This filter adds the comment number to descriptions for paginated comment pages.  It's now outputting the comment number twice.  This broke in v2.3.13 and is probably related to #899.

```
add_filter( 'aioseop_description', 'sfwd_description_comment_page_num' );
function sfwd_description_comment_page_num( $description ) {
$cpage = get_query_var( 'cpage' );
if ( !empty( $cpage ) ) $description = "Pag. n.$cpage - " . $description;
return $description;
}
```

michaeltorbert:
This was most likely adding the extra apply_filter aioseop_description in aioseop_class.php:2571

Since the filter is applied twice, part of it gets repeated.

PR #933, opened 2017-06-09
#932 Removes double calls to `aioseop_description` and removes breaki…

amostajo:
Issue #932 

…ng transformation.

Removes double calls to `aioseop_description`, refactores its call and
removes breaking calls to `htmlspecialchars` and `wp_strip_all_tags`
since these are already handled by the filter and are preventing
decoding to take place.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -2441,6 +2441,8 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


wpsmort:
@amostajo I tested this and it works for PHP v5.4 and later but for anything earlier than v5.4 the meta description, og:description and twitter:description are removed completely from the page source.

wpsmort:
@michaeltorbert I have tested this PR and confirmed that it does now fix the issue with the broken filter.  I tested this with PHP versions 5.2.4 through to v7.1.4 and all worked fine.

Issue #963, opened 2017-07-15
Button is broken on email subscription form


wpsmort:
The button is missing from the form in 2.3.13.2 and 2.3.14.

<img width="622" alt="screen shot 2017-07-15 at 9 24 44 am" src="https://user-images.githubusercontent.com/17548525/28239665-b0a94740-693f-11e7-9769-2a4a3c4d2765.png">

wpsmort:
This originally broke in v2.3.12.1

wpsmort:
@michaeltorbert It looks like it was his commit where the Subscribe button broke https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/a4a058933a46cb75e4fa2131c8de2d719e307339

wpsmort:
I believe this is the offending code:

```
<input type="submit" value="<?php __( 'Subscribe', 'all-in-one-seo-pack' )?>" name="subscribe" id="mc-embedded-subscribe" class="btn">
```

Removing the PHP code from the input value fixes this.

michaeltorbert:
FYI in 2.3.12.5 it is a bit broken as well.
<img width="635" alt="screen shot 2017-07-15 at 11 31 16 am" src="https://user-images.githubusercontent.com/17364998/28240496-230fb77c-6951-11e7-9817-f19608faa432.png">

michaeltorbert:
Also, the "Email Address" placeholder is broken.

wpsmort:
This is tested and is good. Both the placeholder text and the button are present and there are no console errors.

PR #967, opened 2017-07-15
subscribe button and placeholder need to echo #963

michaeltorbert:
issue #963


wpsmort:
This is tested and is good.  Both the placeholder text and the button are present and there are no console errors.

Issue #979, opened 2017-07-19
Hyphens in doc title get converted to apostrophes


wpsmort:
Reported here - https://wordpress.org/support/topic/error-in-title-the-became/

When you use a hyphen in a post title it is displayed as an apostrophe in the doc title.  See screenshot:

<img width="1024" alt="screen shot 2017-07-19 at 10 59 19 am" src="https://user-images.githubusercontent.com/17548525/28373993-765be54a-6c71-11e7-9cfa-52e31c00ae8e.png">

wpsmort:
@amostajo I tested this and found that in the doc title, og:title and twitter:title the hyphen is encoded as &#039; but in the meta description, og:description and twitter:description it is output as a hyphen -

michaeltorbert:
The problem is that &#8211 is a dash, not an apostrophe, which we're checking for in aioseop_class.php.

wpsmort:
This is tested and is good.  I tested all of the following special characters:

Exclamation! &quot;Quotes&quot; Hash# Dollar$ Percent% And &amp; Apostrophe&#039; (Parenthesis) Asterisk* Plus+ Comma, Hyphen &#8211; Period. Slash/ Colon: Semicolon; Less&lt; Greater&gt; Equal = Question? At @ [Brackets] Backslash\ Caret^ Underscore_ Grave` {Braces} Pipe | Tilde~

PR #980, opened 2017-07-19
#979 Hot fix for wrong apostrophe charcode.

amostajo:


wpsmort:
This is tested and is good.

Issue #1002, opened 2017-07-31
Add filter for attributes in Google Analytics


michaeltorbert:
Requested here: https://wordpress.org/support/topic/plugin-not-capturing-ecommerce/#post-9343731
This specifically is for cloudflare support.

michaeltorbert:
https://semperplugins.com/documentation/add-attribute-to-google-analytics/

michaeltorbert:
@wpsmort In testing, make sure that a) I didn't break anything and that b) the filter hook works.

wpsmort:
@michaeltorbert This is tested and is good.  I had to make some changes to the documentation for the filter hook in order for that to work as it was missing a semicolon and the return $attributes.

PR #1010, opened 2017-08-01
added aioseop_ga_attributes filter hook for #1002

michaeltorbert:
#1002


wpsmort:
This is tested and is good.

Issue #1000, opened 2017-07-31
Remove filter hook for Google Asynchronous tracking


wpsmort:
Remove the filter hook documented here - https://semperplugins.com/documentation/use-classic-asynchronous-google-analytics-for-all-in-one-seo-pack/

See also issue #999

wpsmort:
This is tested and is good.

Issue #999, opened 2017-07-31
Remove old code related to Google Analytics classic asynchronous tracking


wpsmort:
Google has now migrated all properties to Universal Analytics tracking code.  We should therefore remove all code related to the old classic asynchronous tracking.

wpsmort:
This is tested and is good.

PR #1019, opened 2017-08-02
Remove old ga code

michaeltorbert:
#999 
#1000


wpsmort:
@michaeltorbert I have tested this PR and confirmed it is good.

Issue #1052, opened 2017-08-06
category or tag is changed to noindex if the last post in category is noindex


michaeltorbert:
_I tested my site with All in One Seo and plugin and I discovered that if the last post in category or the last post with specific tag has a “no index” atribute that all category or tag has “no index”. If you publish new post without “no index” all category will by without “no index” again. I think it’s a bug._

Reported here:

https://wordpress.org/support/topic/plugin-generate-noindex-for-new-tags-and-category/#post-9385543

We need to verify the issue, then see when it started.

arnaudbroes:
Checking.

arnaudbroes:
Confirmed. Once a post belonging to a certain a category is marked as "no index", the category page for that specific category will also output "no index" in the source code.

Traced this back to 2.2 so it's been there for a while.

wpsmort:
@michaeltorbert This bug also includes the NOFOLLOW and Disable on this post settings.  In other words, if you select either of three checkboxes (NOINDEX, NOFOLLOW or Disable) on the last post in a category or tag then that setting will get applied to the category or tag archive.

michaeltorbert:
This is also the case for tags and customer taxonomies. When testing, we need to make sure we don't break it for Pro, which handles things a bit differently.

wpsmort:
@michaeltorbert I tested this going from live free to this branch and confirmed that the Noindex, Nofollow, Custom Canonical URL and Disable on this page/post all change to the correct settings.   I then copied the changes to the dev version of Pro and switched between current free, dev free, current Pro and dev Pro.  I confirmed that the changes in the source code were all correct.

PR #1058, opened 2017-08-08
dont use noindex or nofollow settings for latest post in archive #1052

michaeltorbert:
Issue #1052


wpsmort:

Issue #1053, opened 2017-08-07
Social Settings Tab not activated by default for Posts/Pages


arnaudbroes:
The Social Settings tab isn't appearing by default for Posts & Pages when the Social Meta module is activated for the first time.

michaeltorbert:
@arnaudbroes Can you determine when this behavior started?

arnaudbroes:
Traced this back to the commits of July 11th, 2017. Couldn't check the the exact commit because those builds break the site. 

It's any of these (chronological order):

https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/56db40c0a932226d9b83276496d7a5ff54b12e25
https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/04184bb79d04998b8304ed12954df71a291b5e78
https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/eaf66d8556d2eaa9c7b398f5c1bd0bc0ff6a6488
https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/c50b65d684a48a4aa111f2e8a832faa13ade9f3b
https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/762f9e390ae4041c7fc69a5c8611e4caf53e2b45

michaeltorbert:
possibly this? https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/eaf66d8556d2eaa9c7b398f5c1bd0bc0ff6a6488

wpsmort:
@michaeltorbert PR #1060 is tested and is good.

PR #1060, opened 2017-08-08
social tab default to on for post and page #1053

michaeltorbert:
Issue #1053


wpsmort:
This is tested and is good.  It works for newly installed sites and for sites where I'm updating from the current live version.  In both instances Posts and Pages are selected by default.

Issue #1022, opened 2017-08-03
Never Shorten Long Descriptions no longer works


wpsmort:
Auto generated meta descriptions are still being shortened even when you enable the Never Shorten Long Descriptions option and the Use Content For Autogenerated Descriptions options together under Advanced Settings.

If you don't enable the Use Content For Autogenerated Descriptions option then auto generated meta descriptions are not shortened.

I tested back through the versions and found that this has never worked.

amostajo:
Ready for testing. [PR1083](https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1083)

wpsmort:
This is tested and is good.  PR #1083 is ready for code review.

amostajo:
@wpsmort changes made for #951 fixes this issue as well. Please re-test this using PR 1031
https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1031

wpsmort:
PR #1031 is tested and is good.

Enabling the Never Shorten Long Descriptions option now works regardless of whether you are using the excerpt or the content as the auto generated meta description.

Issue #951, opened 2017-07-07
Character counter counts ampersands as 4 characters


wpsmort:
Reported here - https://wordpress.org/support/topic/meta-description-gets-cut-short/

I tested with the same description and whilst the description is only 151 characters, we seem to be treating it as longer.  I believe this is because we're counting the & sign as four characters, i.e. &amp

wpsmort:
I tested and found that the & sign is actually treated as 4 characters &amp

wpsmort:
I noticed today whilst testing with AIOSEOP v2.3.14.2 that apostrophes in the Site Title are being counted as 6 characters in autogenerated titles.  For example, if you set your Site Title in WordPress to Steve's Site, the autogenerated title on the Edit Post screen shows 20 characters even though "Steve's Site" is only 12 characters.  This is because it's counting the apostrophe as 6 characters for `&#039;`

amostajo:
I am not a SEO savvy, but I have a question here, do the character count affect SEO in some sort of way or is it just to make sure that the text will display well on search engine results?

Because if it does matter, then it should be counted right?

As much as I know, the count was just to make sure snippets will look clean, the have not much influence when indexing on Google, if this is right then yes, the count should be altered so it does not count all charactore in html entities codes.
https://stackoverflow.com/questions/891977/do-html-entities-in-meta-tags-influence-indexing 

What are your thoughs @wpsmort @michaeltorbert ?

wpsmort:
@amostajo The character count is to ensure a clean snippet in SERPs that does not exceed the number of characters allowed, resulting in an ellipses cutting off the title or description.  If Google counts an ampersand or apostrophe as a single character then we should too.

amostajo:
@wpsmort  So this only happens for auto-generated?

wpsmort:
@amostajo Yes, it only happens with the auto generated titles.  The easiest way to test this is to set your Site Title to abc' and then leave your post title empty.  The character count for the title should show 12 which breaks down as:

3 for the space | space
3 for the abc
6 for the `&#039;`

You can do the same with the ampersand to get a count of 11 (5 for the `&amp;`)

amostajo:
Ready for testing.

wpsmort:
@amostajo I've tested this and found one problem:

If you enable the Use Content For Autogenerated Descriptions and the content includes an ampersand then the meta description is shortened.  It is not shortened if you disable that option.

[...]

PR #1083, opened 2017-08-12
#1022 Enable `Never Shorten Long Descriptions`

amostajo:
Issue #1022

Make existing setting `Never Shorten Long Descriptions` functional.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -2562,6 +2562,9 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 
 	/**
 	 * @since 2.3.15 Brackets not longer replaced from filters.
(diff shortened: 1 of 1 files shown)


wpsmort:
@michaeltorbert This is tested and is ready for code review.

michaeltorbert:
@amostajo With your change, would you say we definitely don't need this? https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L2451

amostajo:
@michaeltorbert Not sure. I might have to do some testing and check when is that line called.

amostajo:
@michaeltorbert #951 has modified the way truncate works, the changes applied there make this PR unnecessary.

Issue #1071, opened 2017-08-09
Incorrect link for Edit SEO in Admin Bar on taxonomies


wpsmort:
When viewing a taxonomy such as a category or tag page, the Edit SEO link in the Admin Bar takes you to the latest post.

The correct behavior is:

Free - There should be no Edit SEO link in the Admin Bar (this is a Pro feature)
Pro - The link should take you to the Edit term page for the category, tag, etc.

arnaudbroes:
@wpsmort
I tested this with two different, fresh local test sites and a few versions of WordPress (ranging from 4.8.1 to 4.0) and All in One SEO Pack (ranging from 2.4.15.3 (PRO) to 2.3.0 (PRO)) and it seems that this behaviour has been there since the introduction of SEO for taxonomies such as Categories and Tags. 

Furthermore, this link has been present even before SEO for Categories and Tags was launched, so this means the code was most likely never modified to go to the relevant category/tag page when clicked and just goes by default to the latest post. This link works correctly for both Pages/Posts.

michaeltorbert:
Previously, it was probably showing up everywhere. It should only show up on a post (of any type) or a taxonomy, and should link to the correct edit page.

wpsmort:
@michaeltorbert PR #1086 is tested and the correct Edit SEO links show for both free and Pro.

PR #1086, opened 2017-08-14
correct link in admin bar for edit seo on taxonomies 

michaeltorbert:
#1071


wpsmort:
This is tested and is good.

Issue #1091, opened 2017-08-18
Add filters to change metabox priority and context


michaeltorbert:
In #1090 we moved the AIOSEOP metaboxes up on the edit post screen. 
We should also add two filters, aioseop_post_metabox_context and aioseop_post_metabox_priority so people can change them as desired. 

reference: https://developer.wordpress.org/reference/functions/add_meta_box/

Note: We've  changed these values to normal and high, they used to be advanced and default.

wpsmort:
PR #1092 is tested and is good.  I tested both filters and they worked fine.

Issue #1090, opened 2017-08-17
Move AIOSEOP metabox up


michaeltorbert:

michaeltorbert:
For testing, we need to 
1) make sure they're moved up for all post types where they should show up (try this with a of selected post types, and with social on or off for that post type
2) it's not broken for taxonomies

wpsmort:
PR #1092 is tested and is good.  I tested with different post types from different plugins and with plugins that add meta boxes as well.  I tested the code with free and Pro.

PR #1092, opened 2017-08-18
move metaboxes up on edit post screen

michaeltorbert:
Issues #1090 and #1091


wpsmort:

Issue #1104, opened 2017-08-23
Default OG:Type is incorrect


wpsmort:
Whilst testing the final release of v2.3.16 I found a bug where the default og:type for posts and pages is incorrect.  

On a brand new install of WordPress, the og:type in the source code of a post or page is displayed as "object" and the drop down on the Edit Post screen is "Activity".  Both are incorrect, they should be Article which is what's shown under Social Meta > Facebook Settings.

This bug first appears in commit https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/153c08b3917fdbed95da7c2a3a0f031d51b9d6d8 which was for Issue #1053

michaeltorbert:
Notes:
Site must be new (wipe the aioseop_options entry from the database).
Activate the Social Media Module, but don't click save within it. It appears as though posts default to article, but that value isn't actually being saved until the button is clicked.
Because of this, at some point, activity gets assigned. (Not sure where.)

wpsmort:
@arnaudbroes Can you test back with versions prior to v2.3 and see if it's always existed since v2.1.

The steps to reproduce are:
1. Create a new install of WordPress and install AIOSEOP
2. Activate the Social Meta module in Feature Manager but don't change or save any options
3. Create a new post and go to the Social Settings tab and check the OG Type and check the og:type in the source code of the new post

arnaudbroes:
@wpsmort I tested this back to 2.1.0 and it seems this issue has been there since the introduction of the Social Meta Module.

contactashish13:
@wpsmort is this related to https://github.com/semperfiwebdesign/aioseop-pro/issues/321 and does https://github.com/semperfiwebdesign/aioseop-pro/pull/380 resolve this?

wpsmort:
@contactashish13 Yes, this is somewhat related in that the correct defaults aren't set when the plugin is installed for the first time.  semperfiwebdesign/aioseop-pro#380 does not resolve this.

wpsmort:
This should get fixed in the v4 architecture.

PR #1111, opened 2017-08-24
default fb object type to article

michaeltorbert:
#1104


wpsmort:
@michaeltorbert I tested this and found a bug.  The defaults are now set correctly but when you try to change the default for a post type using the drop down for Posts Object Type or Pages Object Type this does not change the og:type for existing posts or pages.

To reproduce this:

1. Go to the Social Meta module and change the drop down for Posts Object Type to something other than the default
2. View the source code for any post and you should see the og:type remains as "Article" when it should be the new default set in the Posts Object Type drop down

Issue #1207, opened 2017-09-29
Filter for adding shortcodes with images for sitemap


michaeltorbert:
Since we can't apply the_content filter when looking for images for the sitemap, we should have a filter that plugin devs or site owners can use to have their shortcodes be seen by our image scanning for the sitemap.

contactashish13:
@wpsmort @michaeltorbert  PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1214

contactashish13:
@wpsmort you can use this

`
add_filter( 'aioseop_image_shortcodes', 'aioseop_image_shortcodes', 10, 1 );
function aioseop_image_shortcodes( $shortcodes ) {
$shortcodes = '[xxx] [yyy] zzzz [aaaa]';
return $shortcodes;
}
`

contactashish13:
@wpsmort moving the discussion from the PR here so that all relevant information is in one place.

The snippet here https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1207#issuecomment-374914932 will add to every post. The signature of the filter needs to be changed below and then filtered according to the post. Let me know if this does not work.

`
add_filter( 'aioseop_image_shortcodes', 'aioseop_image_shortcodes', 10, 2 );

function aioseop_image_shortcodes( $shortcodes, $post ) {
...
}
`

wpsmort:
Requested again here - https://semperplugins.com/support/general-all-in-one-seo-pack-pro-forum/images-not-being-indexed/

michaeltorbert:
Redundant pr 

https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1230
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1546

PR #1214, opened 2017-09-30
Filter for adding shortcodes with images for sitemap

contactashish13:
Issue #1207


michaeltorbert:
@contactashish13 This would just accept a single shortcode. Should we check for and allow arrays to loop through as well?

contactashish13:
do_shortocode processes multiple short codes so it shouldn't be a problem.

michaeltorbert:
@contactashish13 Can you resolve these conflicts?

contactashish13:
resolved conflicts and added test case as well

wpsmort:
@contactashish13 Please resolve the merge conflicts.

contactashish13:
@wpsmort done.

contactashish13:
@michaeltorbert I think the base branch of this should also be changed to development.

wpsmort:
@contactashish13 I tested this but was unable to get it to work properly.  

I performed a test using the [gallery] shortcode that is built into WordPress but when I added a gallery to a post, the images in that gallery showed up in the XML Sitemap under every single post, page and custom post type.  

I performed a test using the example shortcode here - https://generatewp.com/shortcodes/?clone=image-shortcode and the images added using this shortcode did not show up in the XML Sitemap.

contactashish13:
@wpsmort https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1214#issuecomment-408976240 is because you are not filtering on the post ID. The signature of the filter is 

`apply_filters( 'aioseop_image_shortcodes', '', $postID )`

and if you do not filter using post ID, the images will get added to all posts.

Also, while using https://generatewp.com/shortcodes/?clone=image-shortcode please be sure you are providing a local image. External images are not allowed by default unless the filter `aioseop_images_allowed_from_hosts` is used. I have tried with an external image and it does not get added but a local image does get added correctly.

wpsmort:
@contactashish13 I tested this again with $post_id using an ID of a post where I was testing the shortcode from generatewp.com and still could not get it to work.  Please provide an example that works that I can test with.

contactashish13:
@wpsmort I hope you have kept in mind my comment https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1214#issuecomment-423889210 about external images. 

Here is something you can modify for use:

```
add_filter( 'aioseop_image_shortcodes', 'aioseop_image_shortcodes', 10, 2 );
function aioseop_image_shortcodes( $shortcodes, $id ) {
    if( $id === 1 ) {
        $shortcodes = '[some_short_code]';
    }
    return $shortcodes;
}
```

**_Also, for similar cases in the future for any issue, please provide the exact code you are using as that could have problems._**

wpsmort:
@contactashish13 I tested this and got it to work with the code here - https://generatewp.com/shortcodes/?clone=image-shortcode and with a shortcode from NextGen Gallery plugin.  

Is there any way to get this to work regardless of the attributes.  For example, for this to work with a NextGen gallery the filter must include the full shortcode with attributes, i.e. [ngg src="galleries" ids="1" display="basic_thumbnail"].  

Can you make it work with just the basic shortcode, i.e. [ngg] so it finds all galleries on a site?

contactashish13:
@wpsmort Can you confirm that just `[ngg]` works in a post? If that works, so should this.

wpsmort:
@contactashish13 The shortcode for adding a NextGen gallery to a post requires at least one attribute which is the gallery ID so [ngg] does not work in a post.  The problem is that if you have 100 posts each with a different NextGen gallery added then you'd need to add all 100 shortcodes to the filter which is not scalable.  I hope this makes sense as to why we need to ignore shortcode attributes.

contactashish13:
@wpsmort please confirm that this is behavior you are seeking:

1. User can add the shortcode `[ngg attribute1=something attribute2=else]` using the post editor.
2. Using the filter the user can **register** that `ngg` is a valid shortcode and it should be processed. All other short codes contained in the content will be ignored.

instead of the existing behavior:

1. User has to use the filter to **inject** the shortcode along with its attributes into the post content.

wpsmort:
@contactashish13 Yes, this is correct.

contactashish13:
@wpsmort @michaeltorbert there are 2 approaches here and we need to discuss which one to use. I think the first one is more relevant and in-line with the behavior of what filters usually do. The second one is complicated and can slow down processing.

### inject shortcode
User uses the filter to inject the shortcode along with its attributes into the post content. So if there are 100 posts each with its own shortcode, the filter will need to support all 100 on the basis of the post_id. Now, when the plugin processes each post it will run the filter and run `do_shortcode` on whatever shortcode the filter provides. It will add the content of the shortcode and this will create the final content that will be processed by the plugin.

### register shortcode 
Assume that a site supports 100 different shortcodes and only one of them creates images and needs to be supported. Now, when the plugin processes each post it has to extract this shortcode, run `do_shortcode` on it and then replace the the shortcode with this content. The final content thus created will now be processed by the plugin.

michaeltorbert:
@contactashish13 

I think comments about the PR specifically (ie discussing it not working, merge conflicts, etc) should be in the PR. 
I think your comment about the overall direction of the issue should be in the issue.

michaeltorbert:
Offhand I'd agree with you that the first approach is the way to go.

contactashish13:
@wpsmort I have assigned this to you for further testing.

wpsmort:
@contactashish13 Please resolve the merge conflicts and then I can test this.

contactashish13:
@wpsmort I've fixed the merge conflicts but the travis build is failing because of the functions that were rolled back in 2.9 specifically the line below

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/2.9/modules/aioseop_sitemap.php#L3097

in 2.9, `aiosp_common::parse_content_for_images` is being used to parse images from the content and that method also runs the shortcodes.

But in the latest version the line has changed to

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/modules/aioseop_sitemap.php#L3410

which calls `$this->parse_content_for_images` which does not parse for shortcodes. So this PR is dependent on the fixes pertinent to parsing of images in the content.

wpsmort:
@contactashish13 I'm not clear on what you're saying here https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1214#issuecomment-450492782. If this PR is dependent on a fix to that code, then doesn't that need to be fixed before this PR can be tested?

contactashish13:
@wpsmort I think @EkoJR might be able to make sense of this. @EkoJR can you please review https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1214#issuecomment-450492782 and comment?

EkoJR:
@contactashish13 I'm not sure what changes did occur. I thought you might have a better idea than I would. I was only reverting the changes in the commit.

I understand it was changed to a static method, but since the changes to the method were also included, it makes tracking the changes difficult. Thinking about it now, the PR changes probably should have been separated; 1 for the method changes, and 1 for moving it to a static method.

contactashish13:
@EkoJR I believe the test cases related to those changes also needed to be rolled back and were not. That is why the build is failing. 

@wpsmort this issue is about including images in the content using a short code and the core plugin code no longer parses for images in the content. Should I remove this test case altogether?

wpsmort:
@contactashish13 If the test case is causing the Travis failures and is no longer needed or used then remove it.

EkoJR:
@contactashish13 If it was Isolated Testing, then that is not something I would expect to happen. Which I've mentioned before for reasons like this. Test Cases should be small and as specific as possible to the operation/function or PR changes, and avoid multiple testing within a test; or shared testing.

contactashish13:
That test was a feature test. One feature is a unit. I'm not sure testing on a per line basis will achieve anything. The fact that this discrepancy was caught in fact highlights why such test cases are important.

EkoJR:
@contactashish13 You said 'you believe the test cases related to those changes also needed to be rolled back and were not'. As I've mentioned before, the main change appeared to be moving a method to aiosp_common as a static method. Could try starting there...

contactashish13:
@wpsmort I am marking the test as skipped for the time being.

contactashish13:
@wpsmort I have reinstated the code that was removed and this PR can now be tested

wpsmort:
I've tested this on an existing site and a new site with NextGen Gallery and with the following code but it does not work, no images appear in the sitemap:

```
add_filter( 'aioseop_image_shortcodes', 'aioseop_image_shortcodes', 10, 2 );

function aioseop_image_shortcodes( $shortcodes, $post_id ) {
    global $post_id;
    if( $post_id === 1 ) {
        $shortcodes = '[ngg]';
   }
    return $shortcodes;
}
```

michaeltorbert:
@wpsmort The reason it's not working is that do_shortcode on your [ngg] doesn't unfurl the gallery for some reason when our code is running (I realize it works when viewing the post itself). 
Instead it has the message "We cannot display this gallery".
I don't know why this is, and it's something worth looking into. @EkoJR any insights?

If you use [ngg_images] it does work.

EkoJR:
@michaeltorbert First thing that comes to mind is something may not have been registered yet (like the custom post type), or it could be encapsulated within a particular instance.

Although, I thought I was able to test with NextGEN Galleries successfully in a different issue...which, I also think we already handle this with. https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/modules/aioseop_sitemap.php#L3642

michaeltorbert:
Redundant pr

#1230
#1546

Issue #1225, opened 2017-10-02
Notice for >1000 URLs


michaeltorbert:
If we detect that a site has more than 1000 URLs, and sitemap indexes aren't turned on or it's turned on but it's set to >1000, recommend they change their settings

PR #1231, opened 2017-10-02
notice for lots of sitemap URLs

michaeltorbert:
#1225


michaeltorbert:
@wpsmort This will create usermeta to store that it's been dismissed, so just delete that if you need to see it again after you've dismissed it. 
Check for any undefined warnings, etc and all combinations of relevant settings.  (sitemap index, paginate URLs, max urls per page as well as the actual number of URLs you have)

Issue #1250, opened 2017-10-03
Google Search Console validation error in 2.4.1.1


michaeltorbert:
Reported by Bimal via email and 

https://wordpress.org/support/topic/google-webmaster-sitemap-error/

michaeltorbert:
Looking at the site, this is the first image:

https://i2.wp.com/moviesweed.com/wp-content/uploads/2017/10/Tiger-Zinda-Hai-2017-Hindi-Movie-Poster.jpg?resize=150%2C150&ssl=1

Do we just need to strip all the extra url vars?

michaeltorbert:
I'm creating a backup and clone of his site for testing.

contactashish13:
@michaeltorbert this type of URL will automatically be excluded with the changes in #1212

michaeltorbert:
@contactashish13 It isn't though, he's running 2.4.1.1 which has PR #1215 in it.

bimal008:
@contactashish13 , @michaeltorbert hi,
do you have an idea yet, why its causing error in google webmaster sitemap section ?

bimal008:
Re-submit http://prntscr.com/gsrcay  > same error yet http://prntscr.com/gsrcpn

michaeltorbert:
@bimal008 As you can see, we're working on it now. There's no need for repeated messages.

wpsmort:
@michaeltorbert I was able to reproduce the error on my site.  When submitting the XML Sitemap to GWT I was able to get the same parsing error.  It was caused by the &ssl=1 part of the image URL. 

Once I could reproduce the problem, I tested PR #1253 and I confirmed that this fixed the problem.  In addition to fixing the parsing error in GWT, I also found that this fixes the blank sitemap problem reported by Bimal.  Again, this was caused by the &ssl=1 part of the image URL.  Removing that fixed the blank sitemap. 

I verified this on a brand new site and as soon as I added a second URL parameter separated by an & then both errors occurred.

PR #1253 is ready for code review.

contactashish13:
@wpsmort I reached the same conclusion at around the same time after a long struggle :) @michaeltorbert I have created a new PR with a sub-branch of #1253 because I'm making the image URL xml-safe in the same function I had added earlier. PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1256

wpsmort:
See PR #1256

bimal008:
See PR #1256
[...]

Issue #1249, opened 2017-10-03
Blank Sitemap_post.xml 


blogxkomo:
you can see on https://xkomo.com/sitemap_post.xml 
if you visit you will see blank page

michaeltorbert:
Follow the instructions here https://semperplugins.com/documentation/this-page-contains-the-following-errors/ and here https://semperplugins.com/documentation/how-to-troubleshoot-issues-with-all-in-one-seo-pack/

michaeltorbert:
Update: for moviesweed, the sitemap is blank: https://moviesweed.com/sitemap_post_1.xml but when I did a clone of their site to our server, it works fine http://wordpress-119458-341044.cloudwaysapps.com/sitemap_post_1.xml

contactashish13:
@michaeltorbert we parse using the domdocument and iterate over the elements found to get the 'src'. for some strange reason that src is not found. Even if I use the hasAttribute function it misinforms that 'src' attribute exists. But it does not throw any error

the problematic line is https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/modules/aioseop_sitemap.php#L2851

wpsmort:
@contactashish13 Part of the problem with the https://xkomo.com/sitemap_post.xml sitemap is that we're picking up Facebook links as images.  This is the code in question that we see as an image:

```
<a href="https://scontent-sin.xx.fbcdn.net/hphotos-xpa1/v/t1.0-9/10492216_766557356701076_6278933958430330593_n.jpg?oh=abd9f6df2dd4dd65c6216bf486ee8f61&amp;oe=55D93B66" style="margin-left: 1em; margin-right: 1em;"><img alt="Cara Membuat Cover FB Responsive" height="236" src="https://scontent-sin.xx.fbcdn.net/hphotos-xpa1/v/t1.0-9/10492216_766557356701076_6278933958430330593_n.jpg?oh=abd9f6df2dd4dd65c6216bf486ee8f61&amp;oe=55D93B66" width="640" /></a>
```

This is just like Issue #1212 where we picked up Amazon links as images.  I've opened issue #1258 for this.

wpsmort:
Note: The issue reported by blogxkomo is related to Issue #1250 where there are images that have URL parameters separated by an &.  This causes a blank sitemap and it causes parsing errors in GWT.  

[...]

PR #1253, opened 2017-10-03
Google Search Console validation error in 2.4.1.1

contactashish13:

PR #1256, opened 2017-10-03
Blank Sitemap_post.xml - make image url xml-safe

contactashish13:
Issue #1249 and #1250


wpsmort:
@michaeltorbert This PR has been tested and is good.  I believe that this replaces PR #1253 which can be closed and that this addresses issues #1249 and 1250 which is the same issue with URL parameters separated by an &.

Issue #1263, opened 2017-10-04
Improve Filter hook to remove images from sitemap


michaeltorbert:
We currently have a filter hook aioseo_include_images_in_sitemap to remove images from the xml sitemap (this is helpful for sites that need yandex, which doesn't accept images in the sitemap).

While it's currently sufficient to make it work, we should improve it so that not even the column shows up, as well as refactor it into its own function along with the option check.

Related to #1247

wpsmort:
@michaeltorbert PR #1267 is tested and is good.

Issue #1262, opened 2017-10-04
Option to remove images from sitemap


michaeltorbert:
Related to #1247

michaeltorbert:
When testing, we need to make sure that the option defaults to on for new installs as well as for people updating

wpsmort:
@michaeltorbert @contactashish13  The new Include Images option does not default to enabled for new installs.

michaeltorbert:
@wpsmort  It should default to on _for new installs_ now.

wpsmort:
@michaeltorbert I have confirmed that the option is now checked by default on a new install, however, images don't appear in the sitemap until you click Update Options.  I checked in the database and none of the sitemap module options are shown in the database until you click Update Options.

wpsmort:
@michaeltorbert PR #1267 is tested and is good.  I've tested with all the different configurations that we discussed.

PR #1267, opened 2017-10-04
Option and filter to remove images from sitemap

contactashish13:
#1262 and #1263


wpsmort:
See comment here https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1262#issuecomment-334200908

Issue #1264, opened 2017-10-04
support external images


michaeltorbert:
We should add a filter (or option) to allow externally hosted images, such as those from a cdn.

contactashish13:
@michaeltorbert one way could be to have an option to whitelist comma-separated CDNs/hosts. Maybe a filter is not required in that case.

michaeltorbert:
@contactashish13 Either way we should have a filter, so that it can be programmatically done easily, say on large multisite installations for instance.
We'd need to discuss whether to add an option. Ideally we cut down on options, not add them so that it's a simpler interface. This *may be important enough to have an option though. @wpsmort Thoughts?

contactashish13:
@michaeltorbert have added the filter. We can extend this in case an option is required later. PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1281

wpsmort:
I have tested PR #1281 and confirmed this works.  I used the filter and tested with external images and with images from the WordPress CDN (Photon in JetPack).  

I think a filter should be sufficient and we do not need an option for this.

contactashish13:
synced with master and resolved the conflicts.

wpsmort:
@contactashish13 There are still conflicts showing in PR #1281:

Conflicting files: tests/modules/sitemap/test-sitemap.php

contactashish13:
@wpsmort resolved these conflicts.

PR #1281, opened 2017-10-06
support external images

contactashish13:
#1264


wpsmort:
This is tested and is good.

michaeltorbert:
@contactashish13 Can you fix these conflicts?

wpsmort:
This is tested and is ready for code review.

michaeltorbert:
@contactashish13 Since we now no longer support >WP 4.4, we may as well use wp_parse_url()

contactashish13:
@michaeltorbert I'm confused. I thought we were going not support anything below 4.5 and that support was going to be a major release. Is that not the case?

arnaudbroes:
@contactashish13 I think Michael meant to say "we no longer support <WP 4.5".

michaeltorbert:
Corrected. I had the symbol backwards. wp_parse_url came out in 4.4, we already require 4.4 as a minimum

contactashish13:
So, just to confirm, is minimum support going to be 4.4 or 4.5? I remember discussing 4.5 with you.

michaeltorbert:
Right now we already require a minimum of 4.4. we had parse_url instead of wp_parse_url because we used to have an older minimum requirement.

You and I discussed moving to 4.5 minimum.

PR #1302, opened 2017-10-14
Easier installation of Pro

contactashish13:
https://github.com/semperfiwebdesign/aioseop-pro/issues/365


wpsmort:
@contactashish13 Please resolve the merge conflicts

contactashish13:
@michaeltorbert I think the base branch of this should also be changed to development.

michaeltorbert:
@contactashish13  Travisci failed on this build. Also, there are a number of places in the PR that overwrite (presumably unintentionally) code that's new since this PR was initially created. I told you that this is my big concern for these old PRs. Instead of rushing through them, please make sure your PRs are only doing what you intended them to do.

-Did you mean to remove || defined( 'AIOSEOP_UNIT_TESTING' ) https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1302/files?utf8=%E2%9C%93&diff=split&w=1#diff-4edf793ab6d90512e93955513f2e3a88L469

-and removed if ( ! is_array( $actions ) ) { return $actions;	} at line 321

-and swapped single quotes for double quotes on line 251

-The anonymous function used for the callback in 208 would create a PHP 5.3 requirement.

-You had a couple places of // @codingStandardsIgnoreStart	and end that get removed with this PR. Is that on purpose?

-The version should not be changed back to 2.4.2

contactashish13:
>
> I know Travis failed. I'm still working on it so that's why haven't yet
> marked it as Needs Testing. These old PRs have many stakeholders so I can
> properly address only the code that I have added. How do you propose I
> address those changes that I have no clue about? The ones you are talking
> about were not added by me, if memory servers me right.


BTW 2 PRs were marked erroneously to me and I have assigned them to Ben as
he was the primary on those.

michaeltorbert:
_How do you propose I
 address those changes that I have no clue about_

I don’t know, we can figure it out together, but a pulling in a PR that wiped out new code obviously needs to be avoided.

all_in_one_seo_pack.php
```diff
@@ -499,3 +499,39 @@ function disable_all_in_one_free() {
 		}
 	}
 }
+
+add_action( 'admin_post_aioseop_upgrade', 'upgrade');
```
michaeltorbert: Let's not add to this file. Ideally we'd start moving things out of it instead and organize better. 
At most, let's not add more than an add_action if needed.

contactashish13: I agree. Have used this to include another file that will contain similarly-themed functions.

michaeltorbert:
@contactashish13 You're adding a new updates file, but we already have one https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/inc/aioseop_updates_class.php

michaeltorbert:
I'm closing this as a) it's 1.5 years out of date and b) we haven't reached a consensus on how exactly to handle it, and c) with multiple overlapping issues there's no clear way forward at the moment..

Issue #1202, opened 2017-09-29
XML Sitemap - Full size image for sitemap


michaeltorbert:
When we include an image in the xml sitemap, make sure it's the full size image.

contactashish13:
@michaeltorbert @arnaudbroes PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1350

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #920](https://github.com/awesomemotive/aioseop-pro/issues/920) via [**ZenHub**](https://www.zenhub.com/)

PR #1350, opened 2017-11-09
Full size image for sitemap

contactashish13:
#1202


wpsmort:
@contactashish13 I tested this and it still pulls the image that was embedded in the post at the size that was selected, not the full size image.  See screenshot which shows the thumbnail, medium and large sizes I embedded:

screen shot 2018-03-22 at 3 23 11 pm

wpsmort:
@contactashish13 I tested this and it works.  However, it adds an empty <image:caption> tag to a lot of the images which was not present before.  I think this may be caused by this commit - https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1350/commits/66c5e51c0245f1d64305d99e209be153c0269730

contactashish13:
@wpsmort if the caption/title is empty, it will output an empty caption/title. This behavior seems to be valid according to https://www.google.com/schemas/sitemap-image/1.1/sitemap-image.xsd

wpsmort:
This is tested and ready for code review.

Issue #940, opened 2017-06-28
Sitemap.xml Additional Pages break with no date entered


michaeltorbert:
reported here: https://wordpress.org/support/topic/sitemap-xml-additional-pages/
_ not manually entered a Last Modified date that the options stored in the db ended up empty.

In the aioseop_sitemap.xml the date conversion around line 2272 when used with an empty date string returns 1970-01-01. We had entered a lot of URLs before I released so I implemented a temporary solution adding a basic check

$mod = (empty($mod) ? date(“Y-m-d”) : $mod);

just above this line

$mod = date( ‘Y-m-d\TH:i:s\Z’, mysql2date( ‘U’, $mod ) );_

contactashish13:
@michaeltorbert I believe all the 4 parameters are mandatory. We should have a date selector as well as validate the values and reject values that do not have all 4 parameters. Thoughts?


![image](https://user-images.githubusercontent.com/12953439/32874371-d6f9cfe4-cab7-11e7-9bd2-60530e1e6f34.png)

michaeltorbert:
@contactashish13 Agreed.

contactashish13:
@michaeltorbert PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1364

I can't seem to get rid of codeclimate's false positives. Is there a code directive that I can use in the future to make it ignore specific lines?

michaeltorbert:
We don't want it to ignore specific lines, instead we should use the config file to fine tune it. https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/.codeclimate.yml

michaeltorbert:
@contactashish13  We don't want it to ignore specific lines, instead we should use the config file to fine tune it. https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/.codeclimate.yml

PR #1364, opened 2017-11-16
Sitemap.xml Additional Pages break with no date entered #940

contactashish13:
#940


wpsmort:
This is tested and is confirmed good.  All fields are now mandatory and the Last Modified field is now a date field.

Issue #1353, opened 2017-11-09
Fatal error: Cannot unset string offsets in all_in_one_seo_pack.php on line 328


michaeltorbert:
Reported here: https://wordpress.org/support/topic/fatal-error-string-offsets-in-all-in-one-seo-pack/

_I’m getting the following errors when I try to access my admin dashboard:

Warning: array_unshift() expects parameter 1 to be array, string given in ~/public_html/wp/wp-content/plugins/akismet/class.akismet-admin.php on line 89

Fatal error: Cannot unset string offsets in ~/public_html/wp/wp-content/plugins/all-in-one-seo-pack/all_in_one_seo_pack.php on line 328

My front-facing site is accessible but not the backend. Any help would be appreciated._

EkoJr:
I believe there is a conflict being caused by a different plugin that decided to use `'plugin_action_links'` filter instead of `'plugin_action_links_' . $plugin_name`, and in turn is replacing all other plugin action links, including Akismet.

I was unable to reproduce this, so I can't determine exactly what AIOSEOP can do, but I don't think a resolve can be found, considering it may be a different root issue.

michaeltorbert:
@EkoJr Most likely Jetpack or Akismet, based on reports.

michaeltorbert:
Also reported here: https://wordpress.org/support/topic/fatal-error-string-offsets-in-all-in-one-seo-pack/
Warning: array_unshift() expects parameter 1 to be array, string given in ~/public_html/wp/wp-content/plugins/akismet/class.akismet-admin.php on line 89

Fatal error: Cannot unset string offsets in ~/public_html/wp/wp-content/plugins/all-in-one-seo-pack/all_in_one_seo_pack.php on line 328

michaeltorbert:
https://github.com/Automattic/jetpack/issues/8164
Probably related.

wpsmort:
I have been able to reproduce this on a development site here - http://wordpress-128878-370533.cloudwaysapps.com/wp-admin/

I get these errors when I switched Jetpack to using my WP.com account, previously Jetpack was in Safe Mode as this is a duplicate of a live customer site:

Warning: array_unshift() expects parameter 1 to be array, string given in /home/128878.cloudwaysapps.com/bdqtvwzhbf/public_html/wp-content/plugins/akismet/class.akismet-admin.php on line 89

[...]

PR #1367, opened 2017-11-16
Make sure plugin action links array is an array

michaeltorbert:
#1353


wpsmort:
@michaeltorbert I have tested this on a development clone and a live site and confirmed it fixes the fatal error.  I've only been able to access one live site to test this.

Issue #1363, opened 2017-11-15
Special Characters Issue


springtouchweb:
In the Social Meta tool of your plugin when entering an ampersand after updating the ampersand is converted to "&amp; " that's how it would display in the social sharing card. 
![special-characters](https://user-images.githubusercontent.com/33703454/32861950-196b49f8-ca1c-11e7-86bc-3b50f94dfb5b.PNG)
![home-title](https://user-images.githubusercontent.com/33703454/32861951-197add0a-ca1c-11e7-858e-e77a4fb74bb9.PNG)

springtouchweb:
It displays as `&amp;` ... "& amp;" not "&". Github autocorrected the entity in the first report, please let me know if you have any questions.

contactashish13:
@michaeltorbert PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1372

wpsmort:
@contactashish13 I tested this and it works in that it corrects the OG meta output in the source code.  Is it possible to correct the title and description displayed in the fields under Social Meta > Home Page Settings as well?  The screenshot below shows that we still display the `&amp;`:

<img width="1352" alt="screen shot 2018-03-19 at 6 10 50 pm" src="https://user-images.githubusercontent.com/17548525/37625076-e1491be0-2ba0-11e8-942e-c3f882976f54.png">

arnaudbroes:
This is now good. Needs code review.

PR #1372, opened 2017-11-19
Special Characters Issue

contactashish13:
#1363


wpsmort:
@contactashish13 I tested this but the Home Title and Home Description fields still show the `&amp;` HTML character code instead of the & symbol.  The only way to correct this after updating to the code in this PR is to enter the Home Title and Home Description again.  This is a problem because it requires users to make a change.  

There are two things that need to happen when you update to this code:

1). The & symbol is displayed in the Home Title and Home Description fields without a user having to do anything
2). The og:title and twitter:title output in the source code is corrected from `&amp;amp;` to `&amp;`

See screenshot:

screen shot 2018-07-30 at 4 30 18 pm

contactashish13:
@wpsmort have fixed this.

arnaudbroes:
This is good now. Needs code review.

michaeltorbert:
@contactashish13 What are your thoughts on using wp_specialchars_decode instead of htmlspecialchars-decode?

wpsmort:
@contactashish13 I tested this and found that there are instances where special characters are displayed as the character instead of the HTML character.  For example, an & is displayed in the meta tag instead of `&amp;`.

The behavior of the current version of AIO is that the HTML character is output.  This PR changes that behavior.  This happens on the homepage as well as posts, pages, etc.  

Can you make sure that special characters are output in the source code as HTML.

wpsmort:
@contactashish13 Here is the definitive list of special characters and how we currently output them in HTML source code:

Apostrophe ' - should be output as &#039;
Double quotes " - should be output as &quot;
Ampersand & - should be output as &amp;

All of the characters above are affected by this PR, i.e. the way they are output in the source code has changed.

These special characters are output as is:

Brackets [ ]
Slash /
Backslash \
Hash #  
Dollar $
Percent %
Parenthesis ( )
Asterisk *
Plus +
Comma ,
Hyphen - 
Colon : 
Semicolon ; 
Equal = 
Question ? 
At @ 
Caret ^ 
Underscore _ 
Grave ` 
Braces { } 
Pipe | 
Tilde ~ 
Exclamation ! 
Period .

These are not affected by this PR.

contactashish13:
@wpsmort have handled all these characters. You may want to add `<` and `>` as well to the list.

wpsmort:
This is tested and is good.  Ready for code review

PR #1380, opened 2017-11-25
AIOSEOP meta box appears when SEO for post types is disabled

contactashish13:


wpsmort:
Testing was completed by Arnaud on November 28th but this PR was never marked for inclusion in a release.  I've rebased this PR and tested to make sure it still fixes the bug.  This is good to go.

PR #1419, opened 2017-12-18
XML Sitemap - Can't exclude custom taxonomies

contactashish13:


js/lib/chosen.jquery.min.js
```diff
@@ -0,0 +1,2 @@
...
...e.removeClass("search-choice-focus"),this.pending_backstroke=null},Chosen.prototype.keydown_checker=function(a){var b,c;switch(b=null!=(c=a.which)?c:a.keyCode,this.search_field_scale(),8!==b&&this.pending_backstroke&&this.clear_backstroke(),b){case 8:this.backstroke_length=this.search_field.val().length;break;case 9:this.results_showing&&!this.is_multiple&&this.result_select(a),this.mouse_on_container=!1;break;case 13:this.results_showing&&a.preventDefault();break;case 32:this.disable_search&&a.preventDefault();break;case 38:a.preventDefault(),this.keyup_arrow();break;case 40:a.preventDefault(),this.keydown_arrow()}},Chosen.prototype.search_field_scale=function(){var b,c,d,e,f,g,h,i,j;if(this.is_multiple){for(d=0,h=0,f="position:absolute; left: -1000px; top: -1000px; display:none;",g=["font-size","font-style","font-weight","font-family","line-height","text-transform","letter-spacing"],i=0,j=g.length;j>i;i++)e=g[i],f+=e+":"+this.search_field.css(e)+";";return b=a("<div />",{style:f}),b.text(this.search_field.val()),a("body").append(b),h=b.width()+25,b.remove(),c=this.container.outerWidth(),h>c-10&&(h=c-10),this.search_field.css({width:h+"px"})}},Chosen}(AbstractChosen)}).call(this);
```
aioseopbot: __[ESLint]__

Error - Unnecessary escape character: \&. (no-useless-escape)

arnaudbroes:
@contactashish13 this doesn't seem to work for me. Let me know if I did something wrong, but I made Pro using the changed files, installed it and made a few categories (Posts) and product categories (WooCommerce).

image

wpsmort:
@arnaudbroes I was able to test this.  The reason it wasn't working for you was because you hadn't checked any boxes under Excluded Taxonomies.

arnaudbroes:
@wpsmort makes sense. Thanks for the clarification.

wpsmort:
@contactashish13 I tested this and when I check the Categories box next to Excluded Taxonomies it excludes all posts as well. This happens for any post type. I also found that it generates an empty sitemap_category.xml and empty sitemap_post.xml if Sitemap Indexes is enabled.

contactashish13:
@wpsmort the behavior of removing all posts is as per the implementation and as per my comment: https://github.com/semperfiwebdesign/aioseop-pro/issues/240#issuecomment-426956065. 

I think your confusion is because all posts are in `Uncategorized` by default and `Uncategorized` is a valid category (as you can see in the drop down). This means that when you exclude the `Category` taxonomy without providing its terms, all posts are excluded. This is different from how a custom taxonomy behaves viz. if a post is not assigned to a taxonomy term, it does not have a default taxonomy term like `Uncategorized`. Therefore, if you choose to exclude a complete custom taxonomy, it will exclude only those posts that have been assigned any term in that custom taxonomy. If a post is not associated with a term, it is **not** excluded.

If you still feel there is an issue, it might be helpful for testing and test cases if you could highlight all different combinations and the behaviors you expect.

I have fixed the blank post type and category sitemap issue.

wpsmort:
@contactashish13 The reason for my confusion is that I'd expect that when I check Categories next to Excluded Taxonomies then I'd expect all categories to be excluded from the sitemap.  I'd definitely not expect this to affect all posts as well.  If I'm confused by this then users could be too.  

Is there any way to make sure that only the taxonomy archives are excluded when selecting the checkboxes next to Excluded Taxonomies and that post types are not affected?

contactashish13:
@wpsmort I'm not sure what you expect is realistic given the basic structure of WP. Each post by default is in Uncategorized and Uncategorized is a category; there is no way around that. Excluding all categories will mean excluding Uncategorized as well which means all posts inside that. 

The only way to not confuse anyone (I don't see a confusion because everyone knows posts are by default part of Uncategorized) is to be explicit with instructions in the UI and let the user uncheck Uncategorized.

wpsmort:
Closing this after discussing with Michael and Arnaud.  The issue needs more discussion first.

Issue #1402, opened 2017-12-10
Auto-generated Description not working if post starts with image


waqarts:
![setting 1](https://user-images.githubusercontent.com/30706949/33809122-62e073b4-de13-11e7-9074-8bcd0764db9a.png)
![setting 2](https://user-images.githubusercontent.com/30706949/33809123-6316c1e4-de13-11e7-9674-0fc9a427f1b0.png)
![source view](https://user-images.githubusercontent.com/30706949/33809124-634b824e-de13-11e7-9f36-276b5d0faab1.png)

Auto-generated Description not working on post which is starting with image example url https://apk4funfull.com/2017/04/solitaire-vr-v1-0-apk/

wpsmort:
Reported again here - https://semperplugins.com/support/troubleshooting-all-in-one-seo-pack-pro/autogenerate-descriptions-not-working-for-some-regular-posts/#p7275

wpsmort:
@michaeltorbert I have verified this is a bug.  Here are the steps to reproduce:

1. On a new site, install AIOSEOP and enable Autogenerate Descriptions
2. Create a post with an image as the first object followed by text.  The code should look like this:

```
<img src="http://stevemortiboy.com/wp-content/uploads/2017/09/mega-yacht-20-300x139.jpg" />Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed eget ex volutpat, vulputate nibh ut, tincidunt erat. Morbi nunc velit, lacinia sed nunc ac, vehicula dignissim nisl.
```

3. Verify that there's no auto generated description in the Description field or the source code.

The auto generated description does work if the image is wrapped in an anchor tag like this:

```
<a href="http://stevemortiboy.com/wp-content/uploads/2017/09/mega-yacht-20-300x139.jpg"><img class="alignnone" src="http://stevemortiboy.com/wp-content/uploads/2017/09/mega-yacht-20-300x139.jpg" alt="" width="300" height="139" /></a>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed eget ex volutpat, vulputate nibh ut, tincidunt erat. Morbi nunc velit, lacinia sed nunc ac, vehicula dignissim nisl.
```

This first appeared in version 2.4.

contactashish13:
@wpsmort PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1420

contactashish13:
[...]

Issue #1103, opened 2017-08-23
Post starting with <img> and &nbsp; causes auto-generate of description to fail


iangeek:
Any posts that begin with an image and subsequently followed by a non-breaking space fail to populate the auto-generated descriptions (e.g screen below)

![screen1](https://user-images.githubusercontent.com/1556882/29625898-ff1936ac-8824-11e7-80d1-ee0e189aa28e.jpg)

Removal of the `&nbsp;` reverts back to the expected behaviour (see below).

![screen2](https://user-images.githubusercontent.com/1556882/29625925-0d3302c2-8825-11e7-9351-633516f13d48.jpg)

This appears to have started a few months back, so wondering if its related in some way to any work done on https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/899 ?

wpsmort:
@arnaudbroes Can you test and confirm this behavior and if you're able to reproduce this then please see if you can trace back to when it started.

arnaudbroes:
@wpsmort This works fine with multiple versions of the plugin up to 2.3.16 and with a few WordPress versions ranging from 4.0 to 4.8.1.
![image](https://user-images.githubusercontent.com/18156697/29662501-d9072c6e-88c7-11e7-9bf8-41e40d997d8d.png)

iangeek:
@arnaudbroes @wpsmort  Stripped right back to vanilla plugin-wise, and the theme is just based on Automattic's Underscores, but the issue still persists. Odd.

OK, thanks for checking anyway, and sorry if it was a timewaster. Will let you know if I'm ever able to track down the cause, just for reference.

michaeltorbert:
@iangeek Please use the default theme (Twenty Seventeen) instead of your custom theme. You also may want to try turning on debug error logging.

iangeek:
@michaeltorbert Sorry, should have clarified... I'd already tried default themes and all basic settings before posting the issue.

Will set WP_DEBUG_LOG going and see if it shows anything more.

michaeltorbert:
@iangeek  Export your AIOSEOP settings and send it to us.

aleks683:
The problem is urgent. It's not about plugins. And not in the subject. Removal of the `&nbsp;` does not help either. Autogenerated OG Descriptions works perfectly. What to do?

wpsmort:
PR #1420 fixes this.

PR #1420, opened 2017-12-18
Auto-generated Description not working if post starts with image

contactashish13:
Issue #1402 and #1103


wpsmort:
@contactashish13 I have confirmed that this fixes the issue reported in #1402 and that it is the same as what was reported in #1103.  The screenshots from the customer in #1103 show that the image is not wrapped in a link, it has nothing to do with the non-breaking space.

Can you sync this with the master so that the version matches the current live version 2.4.3.1 and assign it back to me for final testing.

PR #1421, opened 2017-12-18
Issue when using post title as description

contactashish13:


wpsmort:
@michaeltorbert I tested this and have confirmed that it does fix the problem with using %post_title% as the Description Format.  I tested all the macros supported by the Description Format field and they all work correctly. 

During testing I found two related problems which are long running bugs which we don't have issues open for.  I have opened issues for these (they should not hold up release of this bug fix):

1. The placeholder text does not show the correct title or description based on the format that has been set.  I believe this is a wider issue of how the placeholder text is created for both Title and Description fields.  Issue #1808 opened.

2. The Preview Snippet also does not show the correct description based on the format.  It always shows the Excerpt.  Issue #1809 opened.

wpsmort:
@contactashish13 Please resolve the merge conflict.

wpsmort:
@michaeltorbert This is tested and is good.  Ready for code review.

michaeltorbert:
I'd say this needs more work, but we can at least use this as a quick fix until we refactor the title code.

Issue #61, opened 2016-03-26
Add a 5 star ratings request


wpsmort:
Look at how WooCommerce does this and add something similar. See screenshot.

![screen-capture](https://cloud.githubusercontent.com/assets/17548525/14155199/aec89090-f68e-11e5-9064-e87ed214c9d7.png)

michaeltorbert:
This is good also. <img width="1280" alt="ratings request" src="https://cloud.githubusercontent.com/assets/17364998/20975782/7b24c8ac-bc6e-11e6-8b23-9dfe92122f22.png">

michaeltorbert:
So is this 
<img width="1383" alt="screen shot 2017-03-29 at 10 50 33 am" src="https://cloud.githubusercontent.com/assets/17364998/24460951/c5eb7988-146d-11e7-9e9a-3bc854cf5d1c.png">

michaeltorbert:
<img width="1418" alt="screen shot 2017-06-02 at 9 52 25 am" src="https://cloud.githubusercontent.com/assets/17364998/26728762/5450eef4-4779-11e7-8341-a3293b2e6a42.png">

michaeltorbert:
Specs:

1) Should give 3 options:
a) link to rate 5 stars on .org
b) Not now/remind me later
c) already rated

2) Explanation of why: keep developing free software, etc

3) free version only

4) Should be in its own class/file

*Use code from other plugins as a guide if needed.

EkoJr:
Wanted to provide some screenshots to the update. However, I had a few things I wanted to mention for any feedback needed on this.

* Are there any changes, or a preferred way, to the wording in both the Notice and Footer?
* How many days til the notice pops up?
* Show the days as a whole integer or a float decimal (and how many decimal places?)?
* Currently, in admin notice, if either 3 are selected, notice_dismissal is set to true. If the user decides to rate the plugin, but doesn't, and still does. Should we prevent the dismissal just in case a user trying to re-navigate to review the plugin?
* Does AIOSEOP have any deactivation or uninstall hooks? In order to add timestamps for those.

![review-notice](https://user-images.githubusercontent.com/2794445/28343596-ffca534c-6bd2-11e7-9215-ca8ec3c80e8b.png)

![review-footer](https://user-images.githubusercontent.com/2794445/28343599-044b506a-6bd3-11e7-8668-47cefd3d9f45.png)

michaeltorbert:
My feedback:

[...]

PR #1426, opened 2017-12-23
Add Notice Management class & add Rate Plugin notice

EkoJR:
Add Notice Management class & add Rate Plugin Notice #61

### DIFF
--- a/admin/class-aioseop-notices.php
+++ b/admin/class-aioseop-notices.php
@@ -1,0 +1,679 @@
+<?php
+/**
+ * AIOSEOP Notice API: AIOSEOP Notice Class
+ *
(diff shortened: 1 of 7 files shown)


EkoJR:
@michaeltorbert This PR is going to be stale. The newer branch has test cases, and a few other modifications (added a community standard `DISABLE_NAG_NOTICES`, fixed an enqueue script error, and few others I discovered).

I'm also thinking of making it a cleaner PR with Notices object as a separate PR (in case a different notice takes priority).

michaeltorbert:
@EkoJR Ok let's make sure that once you restart this, it's finished to shipped code.

Issue #752, opened 2017-02-22
Add hook to filter return of get_tax_args function


michaeltorbert:
Reference: https://wordpress.org/support/topic/sitemap-all-taxonomiy-terms/

In aioseop_sitemap.php, we should add a filter to get_tax_args() on line 2598 so users can manipulate the arguments https://developer.wordpress.org/reference/functions/get_terms/#parameters

We could either a) just filter the return $args array, or b) add our own individual filters for each argument or c) both.

contactashish13:
@michaeltorbert PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1430

michaeltorbert:
@contactashish13 Why are we adding  $page, $this->options as additional arguments?

contactashish13:
I tend to add any parameter that might be useful for the hook - in this case I can see options being useful for sure to determine taxonomy behavior on the basis of the settings. Can't think of a pressing case for `$page`.

wpsmort:
@contactashish13 I tested this but it doesn't appear that the get_tax_args function is being called.  Can you test and make sure the function is being called and provide a working example of how to use this filter.

contactashish13:
@wpsmort the functions gets called. You can test it by adding just an `error_log`.

The filter is

`apply_filters( 'aiosp_sitemap_tax_args', $args, $page, $sitemap_options )`

You can use the below snippet for removing the taxonomy exclusions:

```
add_filter( 'aiosp_sitemap_tax_args', 'aiosp_sitemap_tax_args', 10, 1 );
function aiosp_sitemap_tax_args( $args ) {
    unset( $args['exclude'] );
    return $args;
}
```

arnaudbroes:
@contactashish13 I'm currently drafting documentation for this filter. From what I understand, this filter allows you to filter taxonomy terms. 

Could you please let me know what these arguments exactly are and what they're used for -

- $args
- $page (is this the page ID?)
- $sitemap_options

PR #1430, opened 2017-12-26
Add hook to filter return of get_tax_args function

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/752


wpsmort:
I have tested with the example provided by Ashish (https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/752#issuecomment-423066644) and it works.  

Ready for code review.

PR #1433, opened 2017-12-28
SEO News feed still attempts to access it even when disabled via screen options

contactashish13:


wpsmort:
@michaeltorbert This is tested and I have confirmed that when you remove the SEO News widget using Screen Options it no longer checks the feed.

Issue #538, opened 2016-10-04
XML Sitemap - Add a rewrite rule instead of our current method


michaeltorbert:
This is especially important for NGINX installs where the rewrite rules aren't present.  

If we're unable to correctly rewrite the sitemap URLs then we should display a notice that shows the rewrite rules and tells the user where to add them in their server config.

michaeltorbert:
This would also help in the event someone needs to write nginx rewrite rules.

contactashish13:
@michaeltorbert @wpsmort PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1434

wpsmort:
@contactashish13 I tested the PR and found that on some installs I get a 404 error for the sitemap.xsl when viewing the sitemap.  I was not able to reliably reproduce this with dynamic sitemaps.  In some cases the problem went away when I refreshed my permalinks.  

However, I was able to reliably reproduce this when using static sitemaps.  In this case I either get the 404 error or I get the following console errors:

- Specification mandates value for attribute itemscope
- attributes construct error
- Couldn't find end of Start Tag html line 2
- Extra content at the end of the document

I can give you access to my stevemortiboy.com site where you can reproduce this.

michaeltorbert:
I've scrapped the PR. I think even better would be to make better use of the Rewrite API. 
For instance, instead of accessing the rewrite object directly, or filtering the rewrite rules array, we should interact via the https://codex.wordpress.org/Rewrite_API

wpsmort:
See also https://github.com/awesomemotive/aioseop-pro/issues/831

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #912](https://github.com/awesomemotive/aioseop-pro/issues/912) via [**ZenHub**](https://www.zenhub.com/)

PR #1434, opened 2017-12-28
For the xsl, look into adding a rewrite rule instead of our current method

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/538


michaeltorbert:
I'm closing this, as I don't think it's heading in the right direction. (I'm not a big fan of adding more OB, etc)

Issue #1428, opened 2017-12-23
Look at file_get_contents in sitemap class


michaeltorbert:
See if this can be improved in some way.  

Some hosts can potentially have  a problem with seeking https://wordpress.org/support/topic/errors-creating-sitemap/#post-9808145

> The problem is on a line with file_get_contents, and seems to be with the offset parameter. https://secure.php.net/manual/en/function.file-get-contents.php “Seeking (offset) is not supported with remote files. Attempting to seek on non-local files may work with small offsets, but this is unpredictable because it works on the buffered stream.”

michaeltorbert:
User reports error only comes on PHP 7.10

According to the changelog for PHP: "7.1.0 | Support for negative offsets has been added."
This should be investigated.

EkoJr:
I've been able to reproduce the issue, and investigated it a bit, and it seems the main issue is the difference between remote and local. Both files are local, but the compression prefix is causing PHP to recognize the string file path to be a remote file. The .xml file however works fine.

The issue also seems to be the cause of another issue, and happens that the warning function used also throws a PHP Warning since 7.1.0+.

Callstack
```
#	Time	Memory	    Function	                                           Location
1	0.0000	459168          {main}( )                                              ...\admin.php:0
2	0.4280	32825256	do_action( )                                           ...\admin.php:224
3	0.4280	32825632	WP_Hook->do_action( )                                  ...\plugin.php:453
4	0.4280	32825632	WP_Hook->apply_filters( )                              ...\class-wp-hook.php:310
5	0.4280	32826648	All_in_One_SEO_Pack_Sitemap->display_settings_page( )  ...\class-wp-hook.php:286
6	3.9022	42739440	do_action( )                                           ...\aioseop_module_class.php:2591
7	3.9022	42739816	WP_Hook->do_action( )                                  ...\plugin.php:453
8	3.9022	42739816	WP_Hook->apply_filters( )                              ...\class-wp-hook.php:310
9	3.9022	42740808	All_in_One_SEO_Pack_Sitemap->do_sitemap_scan( )        ...\class-wp-hook.php:286
10	3.9022	42740808	All_in_One_SEO_Pack_Sitemap->scan_sitemaps( )          ...\aioseop_sitemap.php:965
11	3.9132	42741320	All_in_One_SEO_Pack_Sitemap->sitemap_warning( )        ...\aioseop_sitemap.php:982
[...]

PR #1435, opened 2017-12-28
Fix PHP Error with file_get_contents() & Sitemap not generating #1428 

EkoJr:
* Issue #1428 - Fix PHP Error and sitemap not generating.
    * Fix Static Compressed Sitemap being deleted automatically because of a false-false return.
        * Fix AIOSEOP Error message "Removed empty file *"
        * NOTE: This is caused by a 7.1.0 change that is intended to prevent remote access for hackers.
        * ToDo/NOTE: Problem_Files within the array were all being deleted because of a false potential problem.
    * Fix PHP 7.1.0 - 7.2.0+ (current) Warning with Compressed Streams to file path.
* Fix Static Sitemap undetected to delete when Dynamic Sitemap is enabled.
    * NOTE: file_get_contents was grabbing the end of the file.
* Fix Error message "Potential conflict with unknown file *" with a valid .xml file.

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -1027,9 +1027,12 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 						}
 					} else {
 						if ( $compressed ) {
-							$fn = 'compress.zlib://' . $fn;
+							$file_resource = gzopen( $fn, 'rb' );
+							$file = gzread( $file_resource, 4096 );
+							gzclose( $file_resource );
+						} else {
+							$file = file_get_contents( $fn, false, null, 0, 4096 );
 						}
-						$file = file_get_contents( $fn, false, null, - 1, 4096 );
 					}
 					if ( ! empty( $file ) ) {
 						$matches = array();
@@ -1048,6 +1051,8 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 						$msg .= '<p>' . sprintf( __( 'Removed empty file %s.', 'all-in-one-seo-pack' ), $f ) . "</p>\n";
 						$problem_files[] = $f;
 
+						// This is causing all problem_files to be deleted automatically; which may be the intent.
+						// TODO Either create a seperate variable for this set of problem_files, or a final loop to clean problem_files before returning.
 						foreach ( $problem_files as $f => $file ) {
 							$files[ $f ] = realpath( $file );
 							$this->delete_file( realpath( $file ) );


modules/aioseop_sitemap.php
```diff
@@ -1027,9 +1027,12 @@ function get_problem_files( $files, &$msg ) {
 						}
 					} else {
 						if ( $compressed ) {
-							$fn = 'compress.zlib://' . $fn;
+							$file_resource = gzopen( $fn, 'rb' );
+							$file = gzread( $file_resource, 4096 );
+							gzclose( $file_resource );
+						} else {
+							$file = file_get_contents( $fn, false, null, 0, 4096 );
```
contactashish13: @EkoJr maybe we should check if fopen wrappers have been enabled and if not, use curl?
 http://php.net/manual/en/filesystem.configuration.php#ini.allow-url-fopen

Also, gzopen etc. functions might not work correctly on instances that run on shared hosts. Maybe we should explore usage of the WP Filesystem API.

@wpsmort do you have an instance running on share hosts conforming to the below description from https://codex.wordpress.org/Filesystem_API?

> Different hosting systems have different limitations in the way that their webservers are configured.
> 
> In particular, many hosting systems have the webserver running as a different user than the owner of the WordPress files. When this is the case, a process writing files from the webserver user will have the resulting files owned by the webserver's user account instead of the actual user's account. This can lead to a security problem in shared hosting situations, where multiple users are sharing the same webserver for different sites.
> 
> The WP_Filesystem is capable of detecting when the users for written files will not match, and switches to a method using FTP or similar instead. Depending on the available PHP libraries, WP_Filesystem supports three different methods of using FTP (via extension, sockets, or over-SSH) and will automatically choose the correct method.
> 
> In such a case, the plugin or theme implementing this code needs to request FTP credentials from the user. Functions have been added to make this easy to do and to standardize the look and feel of the credentials entry form.

EkoJr: @contactashish13 It appeared to be a PHP change that causes an issue with compressed streams, and explained it more in detail on the issue ticket. https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1428#issuecomment-354277191

There is an attempt with WP Filesystem API, but doesn't return the object https://github.com/EkoJr/all-in-one-seo-pack/blob/848745994fac2afce28070c7080f38b5ab4a359a/modules/aioseop_sitemap.php#L1003

wpsmort:
This is tested and is good.  I tested with PHP 7.1 and 7.2.  I do not have a test environment where WordPress is running under a different user to the owner of the filesystem.

PR #1449, opened 2018-01-02
XML Sitemaps module - Remove Filename Prefix field

contactashish13:


wpsmort:

Issue #1395, opened 2017-12-06
Support longer meta descriptions


maximbrykin:
Good day! Update your WP-plugin please! Google already support long descriptions - but All in One SEO Pack is cutting them now 😮

michaeltorbert:
@maximbrykin  First, there's no need to to use exclamation points. Just say what you want to say. I'm editing your submission to remove the exclamation points and to make your title appropriate.

Second, please provide specific information about the exact lengths you're suggesting, along with links to supporting documentation.

tonyzeoli:
I'm not sure, but he could be talking about this:

https://en.ryte.com/magazine/google-is-testing-longer-snippets-more-space-for-titles-and-descriptions

Or, this:

https://seopressor.com/blog/google-increases-title-meta-description-length/

Or, this:

https://www.searchenginejournal.com/google-titles-and-descriptions-2016/163812/

All in May of 2016.

I don't see documentation in my search on this change from Google
themselves, but I do see that Product descriptions can be longer for the
Google Merchant Center -
https://support.google.com/merchants/answer/6098336?hl=en

michaeltorbert:
@tonyzeoli I'm familiar with the idea that titles are measured by pixels, and are longer than they previously were, but before we change anything, we need to know conclusively what they should be.
Everywhere says something different.

tonyzeoli:
Agreed - there needs to be consensus and finality to these types of
decisions. Just because a bunch of people uncover something doesn't mean
it's absolute. Better to be safe, than sorry later.

maximbrykin:
https://searchengineland.com/google-officially-increases-length-snippets-search-results-287596

michaeltorbert:
This has been requested numerous times lately.

tonyzeoli:
Per maximbrykin's post, he may not have read this or ignored it altogether: "Some webmasters and SEOs may consider updating their meta descriptions, but I don’t believe Google would recommend doing so. The snippets are more often dynamically generated based on the user query and content found in both the meta description and the content visible on the page. If Google is going to go with a longer snippet, it likely will pull that content from the page."

wpsmort:
The only official statement that I can find from Google is on Twitter here - https://twitter.com/dannysullivan/status/936708405363843072?ref_src=twsrc%5Etfw&ref_url=https%3A%2F%2Fwww.seroundtable.com%2Fgoogle-extends-snippets-320-characters-24862.html

[...]

PR #1454, opened 2018-01-03
Change Meta Description to be longer #1395 PR #1454

EkoJr:
Change Meta Description to be longer #1395 PR #1454

Manual Descriptions aren't trimmed.


wpsmort:
@EkoJr  I have tested this PR and have the following requests:

1. Can you update your branch so it matches our Master which is at version 2.4.3 please.
2. Can you let me know what the filter is for changing the cutoff.  See the comment here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1395#issuecomment-355035771
3. Can you change the text under the Home Description field to reflect the wording "Most search engines use a maximum of 230-320 chars for the description."  See screenshot.

screen shot 2018-01-04 at 3 57 43 pm

EkoJr:
1. I updated all my branches, and added 2.4.3. Did you want me to rebase onto 2.4.3 as well?
2. The filter was a pre-existing filter `aiosp_dont_truncate_descriptions`. [Located here](https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1454/files#diff-4e533641bc4be27bca19bde1a2b60fe9R2564)
```
$description = apply_filters(
    'aioseop_description',
    $description,
    empty( $aioseop_options['aiosp_dont_truncate_descriptions'] )
);
```
3. I saw another 'if' check, which checks if it is already set, and may be the cause of it...but was unsure what implications it had. [all-in-one-seo-pack/admin/aioseop_module_class.php on Line 2257](https://github.com/EkoJr/all-in-one-seo-pack/blob/f17df674552dc4204130dee721fbdcfa1c2c3b7b/admin/aioseop_module_class.php#L2257)

wpsmort:
@EkoJr:  I see the text that needs to be changed on line 2268 of that file.  Can you make that change and then this should be fine.

EkoJr:
That line was one of the changes I tried to make, but wasn't able to get it to work with the rest of the code. Which resulted in inserting the difference with the variable that the inserts the data.

EkoJr:
It is quite possible something else is setting the size for that setting. Since it seemed to have a check for that.

Issue #491, opened 2016-09-09
Canonical urls on custom post type archive pages don't include the required URL variable


michaeltorbert:
reported here: https://wordpress.org/support/topic/canonical-urls-on-custom-post-type-archive-pages/

This is tested and confirmed.

contactashish13:
@wpsmort PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1472

PR #1472, opened 2018-01-13
Canonical urls on custom post type archive pages don't include the required URL variable

contactashish13:
#491


wpsmort:

PR #1477, opened 2018-01-14
Refactor settings initialization function

contactashish13:


wpsmort:
Testing is complete.  I fond no problems with either a new install or an existing site when activating the plugin or resetting the options for any module or all options.

Issue #437, opened 2016-08-17
possible canonical URL issue


michaeltorbert:
Reported here: https://wordpress.org/support/topic/wrong-canonical-1?replies=1#post-8764788

_
_The WordPress has a bug, when page with non-existed subpage is a copy from main page.
Example: http://morrock.com/casino-games/32423 is the same as http://morrock.com/casino-games/

But the problem is that page is indexed by Google and is a duplicate of main page, because the canonical URL is http://morrock.com/casino-games/32423/.
Some websites fix this bug just replacing http://morrock.com/casino-games/*SUBPAGE*/ to http://morrock.com/casino-games/

404 page exist, so some pages like 222 or 333 always open same page. On all websites that use _wordpress__
_

michaeltorbert:
This can be reproduced by adding numbers at the end of a real page... https://semperplugins.com/all-in-one-seo-pack-pro-version/123/

It's not a "subpage" but WP apparently thinks it's a page with pagination and so in this example, the 123rd page.

michaeltorbert:
This actually may be a new issue. We need to test going back years if it always existed. There's a report that it worked in 2014.

michaeltorbert:
This should also be tested not just for pages, but for posts, terms, etc.

wpsmort:
This worked in v2.0 and broke in v2.0.1.

michaeltorbert:
@wpsmort Is this supposed to be closed?

wpsmort:
@michaeltorbert No this should not be closed. I have changed this to the milestone for all canonical URL related issues.

contactashish13:
PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1481

PR #1481, opened 2018-01-15
possible canonical URL issue

contactashish13:
Issue #437


wpsmort:
@contactashish13 I tested this and found that it works perfectly for all post types, however, it does not work for taxonomies.  It incorrectly removes the page number from a paginated taxonomy archive.   Here are the steps to reproduce this:

1. Create a category with enough posts so that the category archive is paginated (i.e. /category/uncategorized/page/2/)
2. Verify that the current version of Pro displays the page number in the canonical link for pages 2, 3, etc. 
3. Check the canonical URL output on pages 2, 3, etc when using this PR.  The canonical URL should be incorrectly displayed as /category/uncategorized/.

Let me know if you need access to m test site for this.

contactashish13:
@wpsmort have addressed this scenario.

wpsmort:

Issue #529, opened 2016-10-03
Remove the No Pagination for Canonical URLs option


wpsmort:
We should remove this option and replace it with a filter.

We need to respect the default and any user setting.

michaeltorbert:
We need to determine if we want to keep the functionality to disable pagination for canonical urls. Why would someone want this? This would cause the >1 pages to be excluded from Google.

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #768](https://github.com/awesomemotive/aioseop-pro/issues/768) via [**ZenHub**](https://www.zenhub.com/)

Issue #528, opened 2016-10-03
Remove the Enable Custom Canonical URLs option


wpsmort:
We should just remove this and set it as enabled for all users.  

We could have a filter for anyone who wants to disable it but I don't think that's necessary.

michaeltorbert:
To recap, we're just removing the setting and all checks for it, but not the functionality.

Issue #374, opened 2016-06-30
Remove Canonical URLs setting


wpsmort:
Under All in One SEO > General Settings is a checkbox for Canonical URLs.  This option should be removed as a setting.  

We should respect whether this is enabled or disabled and there should be a hook for those who want to disable canonical URLs.

Note: There's some conditional logic here that shows or hides additional options for canonical URLs such as No Pagination, Custom Canonical URLs and Set Protocol.

michaeltorbert:
@wpsmort What do we want to do about the conditionally available settings (other than protocol, which we're removing in #443)?
Should we remove those too and replace them with hooks?

contactashish13:
@wpsmort if the setting is removed, what values should the conditionally available settings take on?

wpsmort:
@contactashish13 We have two issues open for removing those options - #528 and #529.  All three issues should be done together so we can clean up this code.

contactashish13:
PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1485 takes care of all 3 issues.

As a default, I'm assuming Enable Custom Canonical URLs to be true and No Pagination for Canonical URLs to be false

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #765](https://github.com/awesomemotive/aioseop-pro/issues/765) via [**ZenHub**](https://www.zenhub.com/)

PR #1485, opened 2018-01-17
Remove Canonical URLs setting

contactashish13:
Issues #374, #528 and #529.


wpsmort:
@contactashish13 Can you provide examples of how to use the new filters you've created here for these:

- Disable canonical URLs
- Disable custom canonical URLs
- No pagination for canonical URLs

michaeltorbert:
I'm moving this to 3.0 so we can discuss more what we want to do here.
We really need a rewrite of the code, since WP handles this now. We should discuss what role, if any, AIOSEOP should play in creating canonical URLs. 
Offhand I'd say we let the WP canonical URL pass through as WP generates it, unless a user wants to modify it via AIOSEOP's filter hook or maybe a setting per page.

michaeltorbert:
I'm closing this PR since the issue needs more discussing about what role AIOSEOP should play with canonical URLs. This code is way out of date with merge conflicts.

PR #1520, opened 2018-01-24
Fix PHP Warning on Page Editor (deprecated to PR #1522) #1491

EkoJr:
Fix PHP Warning on Page Editor #1491

Issue #1491, opened 2018-01-17
Warning: count(): Parameter must be an array or an object that implements Countable 


wpsmort:
Reported in both free and Pro forums here:

https://semperplugins.com/support/troubleshooting-subscribe2-html/error-on-php-7-2-parameter-must-be-an-array-or-an-object/
https://wordpress.org/support/topic/getting-warning-after-installing-all-in-one-seo-pack/

Warning: count(): Parameter must be an array or an object that implements Countable in /wp-content/plugins/all-in-one-seo-pack-pro/aioseop_class.php on line 2643

Happens on a page or post. See screenshot:

![aioseo-php72-error_preview](https://user-images.githubusercontent.com/17548525/35066102-7ec0e5f8-fb9d-11e7-8958-bd2b3f15687c.png)

EkoJr:
I have the same error to I believe. Here's the Call Stack as well.

```
Call Stack
--

1  | 0.0007 | 439400   | {main}( )                                      | ...\post.php:0
2  | 0.3448 | 24443752 | include( 'C:\xampp\htdocs\aioseop\wp-admin\edit-form-advanced.php' ) | ...\post.php:174
3  | 0.5023 | 25216296 | do_meta_boxes( )                               | ...\edit-form-advanced.php:714
4  | 0.5026 | 25216296 | All_in_One_SEO_Pack->display_tabbed_metabox( ) | ...\template.php:1063
5  | 0.5027 | 25216888 | All_in_One_SEO_Pack->display_metabox( )        | ...\aioseop_class.php:4470
6  | 0.5027 | 25216888 | All_in_One_SEO_Pack->display_options( )        | ...\aioseop_module_class.php:2491
7  | 0.5035 | 25228432 | apply_filters( )                               | ...\aioseop_module_class.php:2347
8  | 0.5035 | 25228728 | WP_Hook->apply_filters( )                      | ...\plugin.php:203
9  | 0.5035 | 25229720 | All_in_One_SEO_Pack->filter_settings( )        | ...\class-wp-hook.php:286
10 | 0.5036 | 25229768 | All_in_One_SEO_Pack->get_page_snippet_info( )  | ...\aioseop_class.php:3348
11 | 0.5178 | 25241008 | All_in_One_SEO_Pack->aiosp_mrt_get_url( )      | ...\aioseop_class.php:1403
```

EkoJr:
[...]

PR #1522, opened 2018-01-24
Fix PHP Warning on Page Editor #1491

EkoJr:
Fix PHP Warning on Page Editor #1491

Created a new PR in favor of, instead of PR #1520.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -2658,7 +2658,11 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			return false;
 		}
 		$link    = '';
-		$haspost = count( $query->posts ) > 0;
+		$haspost = false;
+		if ( ! empty( $query->posts ) ) {
+			$haspost = count( $query->posts ) > 0;
+		}
+
 		if ( get_query_var( 'm' ) ) {
 			$m = preg_replace( '/[^0-9]/', '', get_query_var( 'm' ) );
 			switch ( $this->strlen( $m ) ) {
--- a/tests/test-All_in_One_SEO_Pack-aiosp_mrt_get_url.php
+++ b/tests/test-All_in_One_SEO_Pack-aiosp_mrt_get_url.php
@@ -7,7 +7,7 @@
  * Actual:    It is throwing a PHP Error. count(): `Parameter must be an array or an object that implements Countable`
  * Reproduce: Go to post edit screen.
  *
- * @since 2.4.4
+ * @since 2.4.4.1
  *
  * @ticket 1491 Warning: count(): Parameter must be an array or an object that implements Countable #1491
  * @link https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1355
@@ -16,7 +16,6 @@
  * @group post_permalink
  * @group All_in_One_SEO_Pack
  * @group aiosp_mrt_get_url
- *
  *
  */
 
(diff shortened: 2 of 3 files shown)


wpsmort:
This PR has been tested. I reproduced the warning in PHP 7.2 and confirmed this fixes it.   This is ready for code review.

wpsmort:
The correct PR for this is #1522

contactashish13:
@EkoJr A few comments/suggestions on the test case:

- the class name seems to be clunky and maybe not in the right case?`All_in_One_SEO_Pack_Test_aiosp_mrt_get_url `

- https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1522/commits/4c2dbd113e2c4046abd97980e48bd969c7ceedcd#diff-a7d32a5404240c8dfddfb717726e1183R79 has been defined in the base class. You can use that or modify it accordingly so that the base class becomes more useful and generic.

- The class should preferably inherit the base class. You can modify the base class for the setup/teardown methods as it will be useful for all other tests too. 

- can you review the phpdoc comments as some seem out of place? e.g. https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1522/commits/8a0a9a4c57d998b674aea4ffad5afcff9581428c#diff-41fde5a00d3cd8f77c1c24d21c445c29R120

- can you also follow the same folder structure that we had discussed viz. `tests/<module>/test-<specfic>.php`?

- I don't know if its valid or not, but do you also want to test with the following combinations?
a) pretty permalinks on
b) category urls
c) tag urls
d) custom post type urls

- Can you also sync with master so that the tests fire when you create a PR?

EkoJr:
@contactashish13

1. It's a WordPress Standard to name classes by the method you are testing. Although I forgot to add Camelcase for the method name ( https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes ). I also had to extended to WP_UnitTestCase because I needed a testcase, and wasn't part of the sitemap module/class.
2. At the time, I had mentioned in a conversation I couldn't use it since I needed a setup specific to a test and isolated, but since then you added changes, and extending it would make it more generic and useful in the future.
3. I did try to follow similar naming to keep it relatively consistent. I have been a bit hesitant to add anything to the base class though until I see more commonalities, and possibly diverse implementations. Also, at the time I was just trying to get a setup function that I could get an overall view of to understand.
4. Probably, there wasn't much info from WP, how would you suggest the order/placement?
5. Yea, we had discussed the file structure, and I'm still unsure about the file structure that you were using. I was explaining how I was trying to follow WP's structure ( without violating its own standards in naming files...ironically ) ( https://github.com/WordPress/wordpress-develop/tree/master/tests ). We had agreed to exclude `phpunit` as a folder inside of `tests`, but inside the `phpunit` folder would be the ideal structure, and may explain the extra `tests` folder.
6. I could double check, but I believe the method was used just for posts/pages. Pretty urls might affect it though.
7. Yea, I can Rebase, but it usually ends up creating new commits and I believe GitHub made it easier to rebase on merge for Repo Admins.

One important point that I have found. These tests are intended to be isolated. With a testcase class to extend to (base class), it would be intended to assist with the test class. Shared testing I'm not too sure about though.

WP & PHPUnit also seems to focus on class/method & function testing, but with globals, files, or single instances I did notice others mention it as a common conflict with testing multiple instances. In this case though, I focused on the method that was the problem.

EkoJr:
6. Looking at it now, you are correct. There are other instances that could be tested for that method. I believe this was one of the first PRs I was using to get my feet wet with PHPUnit, and focused on the failed test case.

EkoJr:
Closing in favor of #1588

Issue #1526, opened 2018-01-26
Improve phpunit's bootstrap for environment setup & config


EkoJR:
Add configs to allow manual, php environment, or default configs.

Also the ability to load other plugins separately for certain tests, as well as themes (uses different concept).

PR #1458, opened 2018-01-04
XML Sitemap - Add sitemap support for AMP pages

contactashish13:
#37

PR #1527, opened 2018-01-26
Improved environment configuration with plugin's bootstrap #1526

EkoJr:
Improved environment configuration with plugin's bootstrap #1526

@contactashish13 do you approve these?

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -2,37 +2,123 @@
 /**
  * PHPUnit bootstrap file
  *
+ * @since 2.4.4.1
+ *
+ * @link https://github.com/JDGrimes/wpppb
+ *
  * @package all-in-one-seo-pack
  */
+/**
+ * Bootstrap Configs
+ */
+// full path, no trailing slash
+$wp_develop_dir   = 'C:/path-to-location/wordpress-develop';
+$mu_plugin_dir    = 'C:/path-to-location/mu-plugins';
+$plugin_dir       = 'C:/path-to-location/plugins';
+$wp_default_theme = 'default';//twentyseventeen
 
-$_tests_dir = getenv( 'WP_TESTS_DIR' );
-if ( ! $_tests_dir ) {
-	$_tests_dir = '/tmp/wordpress-tests-lib';
+// Disable xdebug backtrace. Why? Does it cause issues?
+// @link https://xdebug.org/docs/all_functions
+if ( function_exists( 'xdebug_disable' ) ) {
+	xdebug_disable();
 }
 
-/************* contactashish13 test environment *************/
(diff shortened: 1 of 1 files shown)


contactashish13:
@EkoJr as long as the test cases already checked in work without a change, it should be okay. If that's not the case, can you please change the rest of the dependencies too?

Also I'm not sure why a separate configuration variable is required for plugins and themes. Plugins should be siblings of the AIOSEOP plugin and themes should siblings of the plugins directory, like a normal WP install.

EkoJr:
> as long as the test cases already checked in work without a change, it should be okay.

No severe changes, but I did notice you made an addition when I rebased, but I kept the addition with the aioseop tests directory.

> Also I'm not sure why a separate configuration variable is required for plugins and themes. Plugins should be siblings of the AIOSEOP plugin and themes should siblings of the plugins directory, like a normal WP install.

The plugin is loaded in the `bootstrap.php`, but there is also need to add other repos ( like AIOSEOP Pro ). Unless something changed to allow nesting git repos. Some cases may also require investigating various Repo Versions to reproduce an issue, but it's mostly other plugins I've had to investigate in general where I've found this useful

However, it is still the same concept, and only requires setting the WP Dev directory like before. Everything else defaults to native settings, but I wanted to open it up to working on multiple repos in the same project; which seemed very useful for Free & Pro.

contactashish13:
@EkoJr check https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1458 where I've tested for dependency on another plugin.

EkoJr:
@contactashish13 That is because you added this Constant. Which isn't even in the master yet.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1458/files#diff-10eafd8bf5fbdde867b86fe38d30b9bbR20

contactashish13:
@EkoJr yeah, many things are not in the master and will be "discovered" as we move from test case to test case. In this case, I've added this constant for 2 PRs that both needed to be tested with external plugins. If this suffices for the purpose you have in mind, can we use this instead?

EkoJr:
@contactashish13 Oh ok. I'll pull some of the branches into mine and test. Especially with #1458 and see what exactly happened, because I tried keeping most of the changes mutual.

contactashish13:
@EkoJr sure. Let me know if things don't work and you want me to change the test cases. There are only about 10 PRs so should not be a big problem.

EkoJr:
@contactashish13 Looks like you marked it as skip. https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1458/files#diff-13d1ed975aa3625dfed14448492c176fR131 

There's also a rebased version. https://github.com/EkoJr/all-in-one-seo-pack/commits/a45c637e964d3f1807c82f411d5c1ba2d62b5c70

I also didn't see any other conflicts.

Issue #1345, opened 2017-11-08
Subscribers can see Notice regarding XML sitemap


scottbuscemi:
![screen shot 2017-11-07 at 10 47 20 pm](https://user-images.githubusercontent.com/4336848/32535172-dc548c5a-c40d-11e7-8a37-02ff85000e5d.png)

Subscribers who access the wp-admin dashboard can see the above notice regarding the XML Sitemap. This should only appear for those with admin privileges.

EkoJr:
I believe this is exclusive to the Pro Version.

scottbuscemi:
What’s the reasoning behind hiding an error message to end users only on
Pro? That doesn’t make sense...

EkoJr:
Well, it is possible there is an added feature that requires a message if it fits a set of conditions.

scottbuscemi:
This is happening on the free version for me.

EkoJr:
Hmm, when working the free version I didn't see code for it, but the pro version had it.

It could of been a recent add that I'm missing, but it could also be a database viable. Either way, I wanted to comment on it. Thank you for replying back on it.

Admin notices recently got some work done, and I'll take another look to see where this got added in the free version.

michaeltorbert:
@EkoJr This is definitely in the free version. It was added in #1225 and can be found at https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/5f978e35b467db9ca35b9d08d97fd5f79afaadcd/modules/aioseop_sitemap.php#L369

@scottbuscemi Thank you for reporting this, we'll get it fixed.

PR #1540, opened 2018-02-01
Fix Sitemap Notice from showing to non-admins #1345

EkoJr:
Fix Sitemap Notice from showing to non-admins #1345 PR #1540
This uses the `aiosp_manage_seo` capability.


wpsmort:
I have tested and confirmed that this works.  The message only shows for users with aiosp_manage_seo capability.

Issue #1548, opened 2018-02-07
XML Sitemap - Mandatory field causing settings not to save


wpsmort:
There was a bug with the code in PR #1364 that didn't show up when I was testing.  

The bug is related to making the fields under Additional Pages mandatory.  Now when you try to save any setting in the XML Sitemap module you get a popup notification to fill out the Page URL field under Additional Pages.  

This needs to be fixed urgently as the code in PR #1364 has already been merged with our development branch in readiness for the next release.  The warning message about mandatory fields under Additional Pages should only show when you click the Add URL button and not when you click the Update Sitemap button.

See screenshot below:

<img width="1388" alt="screen shot 2018-02-07 at 1 59 08 pm" src="https://user-images.githubusercontent.com/17548525/35945403-216847ec-0c2e-11e8-89e0-05d2fbe2c54b.png">

contactashish13:
@wpsmort have fixed this and created a PR against dev: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1549

PR #1549, opened 2018-02-08
XML Sitemap - Mandatory field causing settings not to save

contactashish13:
#1548


wpsmort:
This is tested and is good.  This is ready for code review.

Issue #226, opened 2016-04-20
Bug in Custom Post Type Settings - Empty title format fields


wpsmort:
There are two bugs in All in One SEO > General Settings > Custom Post Type Settings as follows:

1). The checkboxes for custom post types such as Products do not appear until you click Update Options. See screenshot below:

![screen shot 2016-04-20 at 4 14 41 pm](https://cloud.githubusercontent.com/assets/17548525/14688945/0a0bdb76-0713-11e6-9aba-86964f51226d.png)

2). The Title Format fields for custom post types such as Products are empty. This happens when you uncheck all settings and disable SEO for Custom Post Types then click away to a different screen, go back and recheck everything.

michaeltorbert:
@wpsmort Was this fixed?

wpsmort:
@michaeltorbert I have tested this with the latest version from the Dev branch and can confirm that the second bug still exists.

wpsmort:
This is related to issue #68

contactashish13:
@wpsmort I was unable to replicate the second bug. Can you please mention the steps I need to follow?

wpsmort:
@contactashish13 Here are the steps to reproduce:

1. Start with a new install of WordPress, AIOSEOP and WooCommerce.  Make sure you have a product in WooCommerce.
2. Go to All in One SEO > Custom Post Type Settings and uncheck all post types and then disable SEO for Custom Post Types
3. Click away to any other screen in WP Admin
4. Go back to All in One SEO > Custom Post Type Settings, enable SEO for Custom Post Types and check the post types for Products.  Enable the Advanced Options and the check the Custom Titles box
5. Click Update Options
6. Now scroll back down to Custom Post Type Settings and you'll see that the Product Title Format field is blank.  It should contain the text %post_title% | %blog_title%

See screenshot.

<img width="670" alt="screen shot 2018-01-15 at 1 30 29 pm" src="https://user-images.githubusercontent.com/17548525/34956727-5baca422-f9f8-11e7-9f7f-994f1099c2c2.png">

contactashish13:
@wpsmort While trying to fix this, this seemed to fix itself. It could be a problem with my (and consequently your) cache. In the below steps, 4a is what I was trying to accomplish and it already seems to be present. Can you please clear your cache and confirm this?

1. Start with a new install of WordPress, AIOSEOP and WooCommerce. Make sure you have a product in WooCommerce.
[...]

PR #1553, opened 2018-02-10
Bug in Custom Post Type Settings - Empty title format fields

contactashish13:
#226


wpsmort:
This is tested and is good.  It fixes the bug with empty title format fields.  

PR #1588, opened 2018-02-23
Fix PHP Warning on Page Editor #1491

EkoJr:
Fix PHP Warning on Page Editor #1491

Updated PR #1522

PR #1595, opened 2018-03-03
Fix PHP Warning on Page Editor #1491

EkoJr:
Fix PHP Warning on Page Editor #1491

Old PRs #1522 #1588


wpsmort:
This is tested and is good.

PR #1589, opened 2018-02-25
Fix PHP Notice Trying to get property of non-object #1545

EkoJr:
Fix PHP Notice Trying to get property of non-object #1545

This primarily affects Pro users, and a PR for it has been opened as well. 
https://github.com/semperfiwebdesign/aioseop-pro/pull/395

Issue #1545, opened 2018-02-05
PHP Notices: Trying to get property of non-object


wpsmort:
Reported here - https://semperplugins.com/support/general-all-in-one-seo-pack-pro-forum/erorr/

PHP Notice: Trying to get property of non-object in /wp-content/plugins/all-in-one-seo-pack-pro/admin/aioseop_module_class.php on line 2864

PHP Notice: Trying to get property of non-object in /wp-content/plugins/all-in-one-seo-pack-pro/modules/aioseop_opengraph.php on line 1137

PHP Notice: Trying to get property of non-object in /wp-content/plugins/all-in-one-seo-pack-pro/modules/aioseop_opengraph.php on line 1146

michaeltorbert:
FYI

`$term_id = $term_id ? $term_id : get_queried_object()->term_id;`
`$term_id = $term_id ? $term_id : get_queried_object()->term_id;`
`$description = get_queried_object()->description;`

EkoJr:
I added some changes that I believe would resolve the issue, but I was unable to reproduce the error.

I ran different tests with different PHP versions, but neither cause the issue to trigger. I suspect either the issue relates to a nested function `get_terms()` where some recent changes have taken place...or...I suspect a 3rd party conflict possibly in `get_terms()` or `get_terms_by()`, or the object itself being return.

It's difficult to say what exactly is taking place, but added a commit anyways that may better handle the return.

michaeltorbert:
@EkoJr I assume you'd need <5.4 to reproduce, such as 5.2.4/5.3

arnaudbroes:
Perhaps related: https://wordpress.org/support/topic/php-notice-trying-to-get-property-of-non-object-13/

michaeltorbert:
@arnaudbroes Nope.

wpsmort:
@contactashish13 Can you take a look at this and see if you can figure out how to reproduce the PHP Notices reported here as neither @EkoJR or I can reproduce them.

contactashish13:
@wpsmort I'm unable to reproduce these notices as well.

michaeltorbert:
@contactashish13 @EkoJR @wpsmort 

What do we think, do we want to just close it (until we gain access to a user's site) since nobody can reproduce it, or do we like the code and figure it probably fixes it?
Personally, I don't like shipping code that we don't know if it actually accomplishes anything.

contactashish13:
[...]

PR #1596, opened 2018-03-03
Fix PHP Notice Trying to get property of non-object #1545

EkoJR:
Fix PHP Notice Trying to get property of non-object #1545

Old PR #1589


EkoJR:
Resolved merge conflicts #1842

PR #1519, opened 2018-01-24
Add Testcase for Descriptions #1518

EkoJr:
Add Testcase for Descriptions #1518

Issue #1518, opened 2018-01-24
Add Unit Testing for Post Data AIOSEOP_Descriptions


EkoJR:
With the new addition of unit testing, add a testcase for (aioseop module) descriptions output.

EkoJr:
Related to recent past issues where this could have be applied.

#1395 Support longer meta descriptions
#1468 broken aioseop_description filter

PR #1597, opened 2018-03-04
Add Testcase for Descriptions #1518

EkoJR:
Add Testcase for Descriptions #1518

Old PR #1519

### DIFF
--- a/tests/base/class-aioseop-test-base.php
+++ b/tests/base/class-aioseop-test-base.php
@@ -167,6 +167,52 @@ class AIOSEOP_Test_Base extends WP_UnitTestCase {
 		);
 	}
 
+	/**
+	 * (Barebone) Setup Posts and return IDs
+	 *
+	 * @param int   $amount
+	 * @param array $args
+	 * @return array {
+	 *     @type int $post_ID ID of the post inserted.
+	 * }
+	 */
+	protected function setup_posts_return_IDs( $amount, $args = array() ) {
+		if ( 1 > $amount ) {
+			return array();
+		}
+		$default_args = array(
+			'post_type'    => 'post',
+			'post_title'   => 'Post Title',
+			'post_content' => 'Post Content - Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
+		);
+
+		$args = wp_parse_args( $args, $default_args );
+
+		$ids = $this->factory->post->create_many( $amount, $args );
+
+		$posts = get_posts(
+			array(
+				'post_type'   => $args['post_type'],
+				'fields'      => 'OBJECT',
+				'numberposts' => -1,
+			)
+		);
+
+		$this->assertEquals( $amount, count( $posts ) );
+
+		foreach ( $posts as $v1_post ) {
+
+			foreach ( $args as $k2_args_index => $v2_args ) {
+				$this->assertNotNull( $v1_post->$k2_args_index );
+				if ( isset( $v1_post->$k2_args_index ) ) {
+					$this->assertSame( $v2_args, $v1_post->$k2_args_index );
+				}
+			}
+		}
+
+		return $ids;
+	}
+
 	/*
 	 * An empty test is required otherwise tests won't run.
 	 */
--- a/tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
+++ b/tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
@@ -32,7 +32,7 @@ require_once AIOSEOP_UNIT_TESTING_DIR . '/base/class-aioseop-test-base.php';
  *
  * @package WP_UnitTestCase
  */
-class All_in_One_SEO_Pack_Test_GetMainDescription extends AIOSEOP_Test_Base	{
+class Tests_Modules_General_GetMainDescription extends AIOSEOP_Test_Base	{
 	/**
 	 * (Test) Config Variables
 	 */
@@ -213,7 +213,7 @@ class All_in_One_SEO_Pack_Test_GetMainDescription extends AIOSEOP_Test_Base	{
 	 *
 	 * @return array
 	 */
-	public function dataProvider_get_main_description() {
+	public function data_get_main_description() {
 		return array(
 			// 1st Iteration.
 			array(
--- a/tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
+++ b/tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
@@ -1,0 +1,263 @@
+<?php
+/**
+ * Testing All_in_One_SEO_Pack_Sitemap::get_main_description();
+ *
+ * Function:  Get's the AIOSEOP, post meta data, description from the post object (including taxonomies, attachments, etc)
+ *            Which is also used for the General module ( aioseop_class ), OpenGraph module, and Page Snippet.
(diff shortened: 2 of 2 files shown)


tests/base/class-aioseop-test-base.php
```diff
@@ -167,6 +167,52 @@ class AIOSEOP_Test_Base extends WP_UnitTestCase {
...
+	/**
+	 * (Barebone) Setup Posts and return IDs
+	 *
+	 * @param int   $amount
+	 * @param array $args
+	 * @return array {
+	 *     @type int $post_ID ID of the post inserted.
+	 * }
+	 */
+	protected function setup_posts_return_IDs( $amount, $args = array() ) {
```
contactashish13: @EkoJR I don't understand why this method needs to be there to check if the factory has created the post type/title/content correctly. This method is just a wrapper around the create_many method and trying to verify that that factory works so I'm confused about the need for this.

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
```diff
@@ -0,0 +1,253 @@
...
+	/**
+	 * Data Provider for (test_) get_main_description
+	 *
+	 * Provides data for the testing function using PHPUnit's annotation.
+	 *
+	 * @since 2.4.4.1
+	 *
+	 * @return array
+	 */
+	public function dataProvider_get_main_description() {
```
contactashish13: @EkoJR is mixing two cases (camel and snake) advisable?

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
```diff
@@ -0,0 +1,253 @@
...
+require_once AIOSEOP_UNIT_TESTING_DIR . '/base/class-aioseop-test-base.php';
+
+/**
+ * Contains the test case scenario.
+ *
+ * Extending allows adding WP's Testing Unit; which extends to PHPUnit
+ *
+ * @package WP_UnitTestCase
+ */
+class All_in_One_SEO_Pack_Test_GetMainDescription extends AIOSEOP_Test_Base	{
```
contactashish13: @EkoJR The class name is too clunky. Do we necessarily have to use "All_in_One_SEO_Pack" when we are obviously in the plugin folder?

EkoJR: It is a Class name (All_in_One_SEO_Pack) testing a Method/Function (get_main_description). I was trying to make it descriptive to point to a specific Class->Method, and the test method is testing the return value's length.

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
EkoJR: I debated myself on this as well. In this case, since it was PHPUnit annotations, I had added it to mirror it to try and as an indicator to make it easier to find.

contactashish13: I don't think its a good option. Can you please remove it?

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
contactashish13: @EkoJR the method's name might change and this would render this naming convention useless. I think it needs to be simpler so that it indicates the module. Maybe use the classes already present and if they don't suffice, create a new one.

tests/base/class-aioseop-test-base.php
contactashish13: @EkoJR which problem are you referring to in that commit?

EkoJR: It seems GitHub changed the URL. It's this exact URL `https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1588/commits/08d3fbae1c33a9f84e6b0119eed1ec0338dd8509#r171503449`

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
EkoJR: In this case, I was also following WP Standards.

> Test class names should reflect the filepath, with underscores replacing directory separators and TitleCase replacing camelCase. Thus, the class Tests_Comment_GetCommentClass(), which contains tests for the get_comment_class() function, is located in tests/comment/getCommentClass.php.

https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes

tests/base/class-aioseop-test-base.php
contactashish13: @EkoJR I think you can just add another element (elements, in case you decided to add `ids_with` and `ids_without`) to the returned array https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/tests/base/class-aioseop-test-base.php#L164 and add the post ids there.

tests/modules/general/test-All_in_One_SEO_Pack_GetMainDescription.php
contactashish13: @michaeltorbert do you want to weigh in? We need to be able to follow a convention and if we follow WP conventions as they have been defined, we will end up having one class testing one particular function instead of having one class testing a variety of functions that interact.

tests/base/class-aioseop-test-base.php
EkoJR: @contactashish13 I’m still unable to. I need to be able to control the content in order to test long content with descriptions, as well as varied content.
Isolated testing is also a concern. Switching functions will mix in unnecessary variables and if something changes with setup_posts (primarily with descriptions) then it could break the test function; which encapsulation would help to prevent.

EkoJR:
@contactashish13 I was just following the WP standards that were mentioned in one of the outdated comments.

michaeltorbert:
I agree with Ashish. Let’s keep the class names reasonable.

michaeltorbert:
@contactashish13 Once you've code reviewed this, if it's good, can you mark it as ready to merge and assign to me?

contactashish13:
@EkoJR can you please reassign this to me after ensuring that travis passes?

EkoJR:
@contactashish13 There's one last error I'm unable to figure out. Do you think you can make sense of it?

contactashish13:
@EkoJR looks like the error is https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/2.9/aioseop_class.php#L1000 and can be removed by a simple array_key_exists check. Can you please try that?

EkoJR:
@contactashish13 How is that even possible when it's not even in the branch?

https://github.com/EkoJR/all-in-one-seo-pack/blob/i-1518-r5/aioseop_class.php#L1000

contactashish13:
@EkoJR because this is running in the 2.8 branch which has that line: https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/2.8/aioseop_class.php#L1000. Maybe you need to sync your code to the base branch viz. 2.8

EkoJR:
@contactashish13 I understand it is using 2.8 branch, but why? It should be using the PR branch to test with, not 2.8. Which has me curious if PR branches are being tested.

contactashish13:
Those questions are best addressed to Google :) I can only tell you what I see

EkoJR:
Closed to rebase https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2130

Issue #1581, opened 2018-02-20
Title counter color off


michaeltorbert:
Reported here: https://wordpress.org/support/topic/something-is-wrong-with-counting-the-signs-in-the-title/

Previously, the color of the title counter in edit post would change from gray to yellow at x characters, and then to red at y characters.
It now starts at yellow. 

This appears to have broken in 2.4.4.

arnaudbroes:
@michaeltorbert this broke here https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/f17df674552dc4204130dee721fbdcfa1c2c3b7b#diff-4e533641bc4be27bca19bde1a2b60fe9R23 and I've already written and tested a tiny fix for it.

arnaudbroes:
This should be fixed now with my latest commit and PR.

PR #1613, opened 2018-03-16
Fixed character counter for title #1581

arnaudbroes:
#1581

### DIFF
--- a/js/modules/aioseop_module.js
+++ b/js/modules/aioseop_module.js
@@ -75,17 +75,24 @@ function countChars( field, cntfield ) {
 		cntfield.style.color = "#fff";
 		cntfield.style.backgroundColor = "#f00";
 	} else {
-		if (field.name == 'aiosp_title') {
+		if ('aiosp_title' === field.name) {
 			if ( cntfield.value > ( field_size - 6 ) ) {
 			cntfield.style.color = "#515151";
 			cntfield.style.backgroundColor = "#ff0";
 			}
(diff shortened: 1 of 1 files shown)


js/modules/aioseop_module.js
```diff
@@ -75,7 +75,12 @@ function countChars( field, cntfield ) {
 		cntfield.style.color = "#fff";
 		cntfield.style.backgroundColor = "#f00";
 	} else {
-		if ( cntfield.value > ( field_size - 91 ) ) {
+		if (field.name == 'aiosp_title') {
```
contactashish13: @arnaudbroes from a coding standard point of view, the below would be preferable

1. yoda condition 
2. === instead of ==

js/modules/aioseop_module.js
```diff
@@ -75,7 +75,12 @@ function countChars( field, cntfield ) {
...
+			if ( cntfield.value > ( field_size - 6 ) ) {
+			cntfield.style.color = "#515151";
+			cntfield.style.backgroundColor = "#ff0";
+			}
+		} else if ( cntfield.value > ( field_size - 91 ) ) {
```
contactashish13: @arnaudbroes earlier these 2 else conditions would be called for the title as well but with your change, they will never be called for the title. is that expected behavior?

arnaudbroes:
@contactashish13 good remarks, I've revised the code with another commit and I'll create a new PR since it actually has to be pulled to 2.5.1 and not development.

PR #1674, opened 2018-05-02
Fix 1$ to %1$s

shoheitanaka:
Fix  1$ to %1$s and  2$ to %2$s


contactashish13:
ref https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1888

wpsmort:
I tested this PR and it does not fix the problem.  Here is what is output by this PR:

screen shot 2018-09-10 at 9 30 21 am

Issue #1709, opened 2018-05-24
Fatal error:  Uncaught Error: [] operator in aioseop_class.php:4823


wpsmort:
Fatal error when testing v2.7 of Pro:

Fatal error:  Uncaught Error: [] operator not supported for strings in aioseop_class.php:4823

This is caused by the empty pair of brackets and is triggered when you disable Custom Post Type Settings.

michaeltorbert:
Check if this commit broke it: https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/9135c839faca710ee9913f2a0429efb844e5c5d7

wpsmort:
In testing, removing the [] from line 4823 clears the fatal error and still outputs the SEO meta box on custom post types and taxonomies.

PR #1710, opened 2018-05-25
pass array for #1709

michaeltorbert:
issue #1709 

This should fix it, but in testing, we need to make sure it doesn't break the original issue with the meta box again.


michaeltorbert:
This should fix it, but in testing, we need to make sure it doesn't break the original issue with the meta box again.

wpsmort:
@michaeltorbert This is tested and is good to go.  It does not break the original issue and does not cause any new issues.

Issue #1730, opened 2018-06-07
Ignore character counter field when tabbing


wpsmort:
On the Edit Post screen when you tab from the Title field to the Description field you tab first to the character count field.  This can easily be fixed by adding tabindex="-1" to the character count field.

PR #1737, opened 2018-06-12
Fixed tabindex for character counter box

wpsmort:
#1730


wpsmort:
@arnaudbroes please review the code and then assign to @michaeltorbert for testing

Issue #1647, opened 2018-04-07
Virtual Robots.txt module


wpsmort:
This is the master issue for the new virtual Robots.txt module.  The following tasks need to be completed:

- [x] Robots.txt should filter built-in WordPress virtual robots.txt (old issue #214)
- [x] Remove all code related to handling a physical robots.txt file
- [x] Remove Robots.txt tab from File Manager module (old issue #213)
- [x] Remove the Optimize your Robots.txt File functionality from current module
- [x] Remove Save Robots.txt File button and Delete Robots.txt File button from current module
- [x] Store all robots.txt rules in database
- [x] Sanitize entries before writing to the database
- [x] Sanitize paths to add trailing slash when none is present
- [x] Check for presence of a physical robots.txt file and display a warning with prompts to Import and delete existing file, delete existing file, do nothing (old issue #596)
- [x] Import existing physical robots.txt file when user visits the Robots.txt module
- [x] Sanitize the import of existing physical robots.txt file during import 
- [x] Support WordPress multisite with sub domain installs (old issue #1143)
- [x] Support WordPress multisite with domain mapping (old issue #1124)
- [x] Network Admins should be able to set default rules (old issue #1600)
- [x] Check for duplicates or conflicts for rules added by Site Admin to rules added by Network Admin
- [x] Allow wildcards in rules and check for duplicates or conflicts

wpsmort:
Google has a developer doc for robots.txt that is a must read regarding order of precedence - https://developers.google.com/search/reference/robots_txt

In testing I found that when there is a Disallow directive and an Allow directive in the same group, the Allow directive always overrides the Disallow regardless of its position within the group (for example, the Allow can be before or after the Disallow, it will always match). This is not mentioned in the developer doc.

**Example of trailing slash precedence in robots.txt:**

User-agent: Googlebot-Image
Disallow: /wp-content/image-test  -  This Network Admin rule is intended to take precedence but it's missing the trailing slash

User-agent: Googlebot-Image
Allow: /wp-content/image-test/  -  This Site Admin rule takes precedence for the directory because it has the trailing slash

**Desired outcome above is that we enforce strict paths so that the Network Admin rule has the correct trailing slash and takes precedence over the Site Admin rule.**

------
**Example of rule matching:***

User-agent: Googlebot-Image
Disallow: /wp-content/image-test/  -  This is Network Admin rule is intended to take precedence 

User-agent: Googlebot-Image
[...]

PR #1742, opened 2018-06-19
remove extra menus 

michaeltorbert:
#1647


wpsmort:
@michaeltorbert This is tested on single site and multisite and I've confirmed it works fine.

Issue #1729, opened 2018-06-07
XML Sitemap - UI problem with date picker


wpsmort:
The icon for the date picker is obscured on some screen sizes.

<img width="267" alt="screen shot 2018-06-07 at 12 49 13 pm" src="https://user-images.githubusercontent.com/17548525/41115693-491e97d8-6a56-11e8-9bf4-ec6311c3269d.png">

wpsmort:
I have improved the layout of the Additional Pages section for XML Sitemap and Video XML Sitemap so  that instead of using a tabular layout, the fields are in a form layout.  This provides more space for the Last Modified date field.  Here is a screenshot of the new layout:

<img width="868" alt="screen shot 2018-06-20 at 5 37 18 pm" src="https://user-images.githubusercontent.com/17548525/41686254-6329d34e-74b1-11e8-88ce-a003f6e1f72c.png">

PR #1746, opened 2018-06-20
Improved layout for Additional Pages section

wpsmort:
#1729

### DIFF
--- a/css/modules/aioseop_module.css
+++ b/css/modules/aioseop_module.css
@@ -810,38 +810,79 @@ div.aioseop_feature#aioseop_coming_soon2 .aioseop_featured_image {
 	/*	max-width: 900px; */
 }
 
-#aiosp_sitemap_addl_pages,
-#aiosp_video_sitemap_addl_pages {
-	clear: left;
-	margin-left: 20px;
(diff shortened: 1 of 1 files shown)


wpsmort:
@arnaudbroes When testing please make sure to test all browsers and test with RTL languages and test both the XML Sitemap and Video XML Sitemap (you'll need to copy the CSS file into Pro to test the Video XML Sitemap).

arnaudbroes:
There are a few issues:

1. Non-Latin characters are shown as UTF-8 hexadecimal values.
image

2. Styling in Microsoft Edge.
image

3. Text does not wrap on smaller screens in Google Chrome.
image

I'm not a fan of using tables because they don't like smaller screen sizes, although I can't think of an alternative, so we might want to set the 'overflow-x' attribute to 'auto' in the CSS code or use CSS3 Flexbox.

wpsmort:
I've opened a new issue for the non-Latin characters.  I believe it's the non-Latin characters that cause the wrapping issue and therefore the MS Edge issue.  I tested with long URLs that use Latin characters and the problem with wrapping and MS Edge overlap does not happen.

Issue #1762, opened 2018-07-10
Help icons not working


wpsmort:
When clicking on a help icon in the 2.7 branch nothing happens and I get these JS errors in the browser console:

Uncaught TypeError: $ is not a function - aioseop_module.js?ver=2.6:851
    at aiospinitAll (aioseop_module.js?ver=2.6:851)
    at HTMLDocument.<anonymous> (aioseop_module.js?ver=2.6:642)
    at i (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,utils,moxiejs,plupload&ver=4.9.6:2)
    at Object.fireWith [as resolveWith] (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,utils,moxiejs,plupload&ver=4.9.6:2)
    at Function.ready (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,utils,moxiejs,plupload&ver=4.9.6:2)
    at HTMLDocument.K (load-scripts.php?c=0&load[]=jquery-core,jquery-migrate,utils,moxiejs,plupload&ver=4.9.6:2)

Uncaught ReferenceError: toggleVisibility is not defined - admin.php?page=aiosp_opengraph:464 
    at HTMLAnchorElement.onclick (admin.php?page=aiosp_opengraph:464)

wpsmort:
@michaeltorbert I found that removing the $ from inside the parenthesis after the aiospinitAll and aiospinitCounting functions fixes the first JS error.  So the code that works looks like this:

`function aiospinitAll(){`
`function aiospinitCounting(){`

However, I cannot work out what is causing the second JS error - toggleVisibility is not defined

arnaudbroes:
This actually does not only break the Help buttons, but all JavaScript on the page due to a scope error.

PR #1767, opened 2018-07-10
Fix JavaScript errors #1762

arnaudbroes:
#1762 
@wpsmort can you test this and then assign it to Michael for code review?


wpsmort:
This is tested and is good.

PR #1822, opened 2018-08-07
Use `rewrite_rules_array` as a filter

adamsilverstein:
Hi! I noticed this when reviewing the plugin functionality.

`rewrite_rules_array` is a filter, not an action and even though this works currently it is technically incorrect and might break in the future: https://developer.wordpress.org/reference/hooks/rewrite_rules_array/


michaeltorbert:
Good catch. FYI it works because add_action is an alias of add_filter. It's been this way for a decade or so and isn't likely to change.
https://developer.wordpress.org/reference/functions/add_action/#source

Having said that, it definitely makes more sense to change it to a filter in the code.

adamsilverstein:
>  add_action is an alias of add_filter.

Ah, right! That makes sense, thanks for pointing that out @michaeltorbert

adamsilverstein:
👍 Happy to contribute!

PR #1828, opened 2018-08-10
Fix/optimize sitemap generation

soulseekah:
Some sitemap generation optimizations:

1. Preheat and cache calls to `get_post_thumbnail_id` when generating sitemap.
2. Defer `All_in_One_SEO_Pack_Sitemap::do_sitemaps` until shutdown and only call it once per request. (i.e. no need to generate the sitemaps 10 times in a row for 10 bulk-deleted posts).

Discussed and developed live (in Russian) https://youtu.be/FHSS7G2_GSI

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -400,7 +400,15 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 				return;
 			}
 
-			$this->do_sitemaps();
+			if ( defined( 'AIOSEOP_UNIT_TESTING' ) ) {
+				$this->do_sitemaps();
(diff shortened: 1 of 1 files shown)


michaeltorbert:
#1832

wpsmort:
I tested this on a site with 5000 posts and 2500 pages.  It reduced the memory used from 107 MB to 73 MB and the page load time from 17 seconds to 10 seconds. 

@contactashish13 Can you perform a code review of this please.

majestica:
I have a blog for 20,000 records, with a site map enabled and your plugin, any update / deletion of the post takes about 90 seconds, it's a total nightmare, with this fix it now takes about 45 seconds, it's much better, but even in this case very long, you did not think to optimize it? why every time there is a reorganization of the entire site map and all the posts we are sorting out, why can not you store somewhere a cache and simply add a new one?

Issue #1761, opened 2018-07-10
WPML string translation hooks


srdjan-jcc:
Translations for Home Page Title and Home Page Description fields are not applied.
AIOSEOP is calling get_options() too early and translations are not applied.

When home page is loaded, get_options() hook is only once called from file/line
all-in-one-seo-pack-pro/all_in_one_seo_pack.php :126

If aioseop_get_options() is called on 'init' hook - translations are applied.

**To Reproduce**
1. Edit Home Page Title and Home Page Description fields
2. Translate Home Page Title and Home Page Description fields using WPML String Translation
3. Visit home page in other language
4. Translations are not applied

**Expected behavior**
Translations are applied to Home Page Title and Home Page Description fields.

michaeltorbert:
Thanks for the report. Feel free to submit a PR if you like.

srdjan-jcc:
@michaeltorbert I have no exact suggestions for changing initialization.
Here is snippet that fixes issue:
https://gist.github.com/srdjan-jcc/e4352428f1c7cc4271b7aef52aeeb0af

Tested and looks it has no (bad) impact on other code.
Would be good to perform your tests too.
 
Do you think it can be added like that somewhere in AIOSEOP?

michaeltorbert:
`Do you think it can be added like that somewhere in AIOSEOP? `

@srdjan-jcc Sure, but as we have hundreds of issues, we have to prioritize, so submitting a PR is the quickest way to get it included.

srdjan-jcc:
@michaeltorbert Created PR here https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1876

Please make sure fix is also applied to PRO version, thank you!

michaeltorbert:
When testing, make sure to first verify the problem, and that the linked PR fixes it.

arnaudbroes:
@wpsmort this has been tested. Works fine. @michaeltorbert needs code review.

PR #1876, opened 2018-09-03
Fixed option filtering

srdjan-jcc:
WPML stopped filtering AIOSEOP options.
get_options() is called too early (soon as plugin is loaded) and translations are not applied.
Fixed by triggering another call on 'init' hook.


michaeltorbert:
#1761

arnaudbroes:
@wpsmort this has been tested. Works fine. @michaeltorbert needs code review.

PR #1887, opened 2018-09-07
Video XML Sitemap - Filename is deleted on update

contactashish13:
https://github.com/semperfiwebdesign/aioseop-pro/issues/424

### DIFF
--- a/all_in_one_seo_pack.php
+++ b/all_in_one_seo_pack.php
@@ -4,7 +4,7 @@
 Plugin Name: All In One SEO Pack
 Plugin URI: https://semperplugins.com/all-in-one-seo-pack-pro-version/
 Description: Out-of-the-box SEO for your WordPress blog. Features like XML Sitemaps, SEO for custom post types, SEO for blogs or business sites, SEO for ecommerce sites, and much more. More than 30 million downloads since 2007.
-Version: 2.7.2
+Version: 2.7.3
 Author: Michael Torbert
 Author URI: https://semperplugins.com/all-in-one-seo-pack-pro-version/
 Text Domain: all-in-one-seo-pack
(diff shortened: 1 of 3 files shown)


wpsmort:
@contactashish13 I tested this and found one problem.  The sitemap name isn't being stored in the database.  This means that when you export settings using the Importer / Exporter module, the value for the video sitemap filename is empty:

aiosp_video_sitemap_filename = ''

Can you correct this so that the filename is stored in the database just as it was prior to the introduction of this bug.

contactashish13:
@wpsmort I suppose that is because you have reset options and once you do that, the filename vanishes from the database because it is not part of the `options` of the plugin. Can you please confirm that the sitemap name is exported as empty even when the database contains the value?

wpsmort:
@contactashish13 It's not because I reset the options, it's because I clicked Update Options in the Video XML Sitemap module in v2.8.3 of Pro.  When you do that it wipes the filename from the database which is the bug.  We need to restore the filename to the database.

contactashish13:
@wpsmort According to the code, fields that do not exist are not touched by Update Options. Additionally, I have tested this and if the field name `aiosp_video_sitemap_filename` exists in the database, merely clicking Update Options does not wipe this value out. This has something to do with Reset Options. Can you please double check?

wpsmort:
@contactashish13 The bug happens every time I click Update Sitemap in the Video XML Sitemap module.  This causes the value in the database to be removed so it looks like this - s:28:"aiosp_video_sitemap_filename";s:0:"";

To fix this bug we must set that value back to s:28:"aiosp_video_sitemap_filename";s:0:"video-sitemap";

To answer your question, this has nothing to do with the Reset button, it happens when clicking the Update button.

contactashish13:
@michaeltorbert the code and my testing reveals that any field that is not explicitly provided in the `options` is not touched when Update Options/Sitemap is clicked. This means that even though we removed the field from the `options`, clicking Update ignores that field and updates all other fields thereby retaining its value. @wpsmort's testing results do not behave in a similar manner. I can add a hidden field so that the filename option is part of the `options` and gets updated along with everything else but I feel that is an overkill considering how the code is supposed to behave. Can you please let me know which option you prefer?

This bug, by the way, was introduced in the free version so the Sitemap module is also affected by this. This can be tested on that module as well.

wpsmort:
@michaeltorbert This PR fixes the immediate high priority bug where the Video XML Sitemap has the same URL as the normal XML Sitemap.  I recommend we open a separate issue for setting the default Video XML Sitemap filename in the database which can be worked on for the next minor or major release.

Issue #1888, opened 2018-09-08
Description in Post Title Format is messed up


contactashish13:
![image](https://user-images.githubusercontent.com/12953439/45252405-8df07580-b34d-11e8-9397-e68861c7906f.png)

contactashish13:
@wpsmort you can disregard and close https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1674. I have corrected the bug in https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1898

PS - @michaeltorbert I think testing ought to come before code review as the latter will not make sense if things do not work as expected.

michaeltorbert:
@contactashish13 So that @shoheitanaka gets some of the credit, I edited his PR with your changes.

PR #1898, opened 2018-09-11
Fix 1$ to %1$s

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1888


wpsmort:
@michaeltorbert This is tested and is good to go.

Issue #1899, opened 2018-09-11
Uploaded to this post dropdown (images) disappeared


netinial:
After Auto Update (2.8)
Uploaded to this post (images) disappeared

Pages or Posts
Click Add Media
From Media Library
Under 
All Media Items
There should be Uploaded to this post.

I confirm that when I disabled AIOSEO from plugins it is appearing again.
False Screen
![adsiz](https://user-images.githubusercontent.com/1303482/45349155-36564200-b5b9-11e8-98ee-f3a892fffc85.jpg)

True Screen
![adsiz2](https://user-images.githubusercontent.com/1303482/45349218-60a7ff80-b5b9-11e8-913a-43f6b1149dac.jpg)

michaeltorbert:
@netinial We're unable to reproduce this, can you send us your wp-admin information so we can log in and resolve the issue? https://semperplugins.com/contact/support-login-information/

netinial:
I send via
https://semperplugins.com/contact/support-login-information/

netinial:
If there is no way. Ok.

netinial:
Dear Michael, This is almost 3 days and we still can not use our website to update pictures. (It shows all pictures instead of post's)

Since disabling is not good option.

Would you please let us know are there any update?

michaeltorbert:
@netinial When there are updates, you'll know on this page. Asking if there are any updates will only serve to slow development.

michaeltorbert:
When testing, we need to determine conclusively whether we introduced this in 2.8.

contactashish13:
That line of code (https://github.com/contactashish13/all-in-one-seo-pack/commit/968b1891e87e53e5f3cf4f2cc397d3ad94f08234#diff-604ad90165044b2b600ad4f4148dfacaL1771) seems to be there for the last 2 years so I doubt it was introduced in 2.8.

michaeltorbert:
I've confirmed that the bug does not exist in 2.7.3. @arnaudbroes Can you figure out which commit caused it?
Steps to reproduce:

Install AIOSEOP.
Install the active3 theme.
Create new post, add media, change tab to "media library," then "Uploaded to this post" is missing.

[active3.zip](https://github.com/semperfiwebdesign/all-in-one-seo-pack/files/2385990/active3.zip)

netinial:
Michael Is this private temporary clone ?
Please delete our theme from this place.

wpsmort:
[...]

PR #1905, opened 2018-09-14
Uploaded to this post dropdown (images) disappeared #1899

contactashish13:
#1899 (Edited to add issue -Michael Torbert)

### DIFF
--- a/admin/aioseop_module_class.php
+++ b/admin/aioseop_module_class.php
@@ -1768,7 +1768,12 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Module' ) ) {
 		}
 
 		function admin_enqueue_scripts() {
-			wp_enqueue_media(); // WP 3.5+ Media upload.
+			global $post;
+			$args = array();
+			if ( $post ) {
+				$args = array( 'post' => $post->ID );
+			}
+			wp_enqueue_media( $args );
 		}
 
 		/**


michaeltorbert:
@contactashish13 @EkoJR What's the status of this? Ready for testing and code review?

contactashish13:
@michaeltorbert sorry, this must have fallen off the radar somehow. I have assigned the issue to @wpsmort for testing.

EkoJR:
I didn't notice Ashish's commit either...at the time, I had just finished a related PR and was looking for an issue to tackle, then I noticed issue #1899 was related.

The problem was with `wp_enqueue_media()` in `\All_in_One_SEO_Pack_Module::admin_enqueue_scripts`, but the proper enqueue is already handled in  `\All_in_One_SEO_Pack_Module::enqueue_scripts`. So the problem then became an issue with the instance being used; which is usually determined whether the hook was added or not. Which was then traced back to when a hook is added, however the "style" enqueue was still needed for the menu icon fix.

So, technically as an example, if you remove wp_media_enqueue() in `\All_in_One_SEO_Pack_Module::admin_enqueue_scripts` it would work properly since the next hook `enqueue_scripts` has the proper handling for it, and is for the intended screen instances as well.

Ashish's PR does solve the issue, but it just patches an issue I introduced, and might be better to reduce than add to it (also avoids creating duplicates). ...but PR #1935 is set for 2.9 as well.

michaeltorbert:
Closing since #1935 also fixes the issue.

Issue #535, opened 2016-10-03
XML Sitemap module - Remove the Link From Virtual Robots.txt option


wpsmort:
We should remove this option.  

~~We should respect the users setting.~~

It should be enabled by default and we should have a filter to disable this.

michaeltorbert:
> Is there any reason we want this to be optional? If we remove it, we just need to make it clear it happens... or do we?

> I'm thinking something like:

> [green light] Added to virtual robots.txt
[green light] Some other thing we do
etc

michaeltorbert:
Just adding a reminder that there should be some indication that this is happening.

contactashish13:
PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1909

arnaudbroes:
So here's what we need to happen:

1. Sitemap URL is added automatically to the virtual robots.txt file
2. XML Sitemap menu indicates whether or not the link is present in the virtual robots.txt file
3. Filter hook (can be a simple boolean) needs to be added so that users can disable this setting if they want to

contactashish13:
Filter to use:

```
add_filter( 'aiosp_sitemap_link_to_robots', '__return_false' );
add_filter( 'aiosp_video_sitemap_link_to_robots', '__return_false' );
```

michaeltorbert:
Playing devil's advocate, why exactly do we want to respect the old setting? This is something that helps, and can not hurt, SEO. If someone for some crazy reason doesn't want the link to their sitemap in the robots.txt, do we really want to code burden to allow the previous setting, for a setting that doesn't even exist anymore?

wpsmort:
I don't see any reason to respect the old setting for this option.  There's no downside to including the sitemap URL in the robots.txt and therefore there's no damage if someone had this option disabled and we enable it by default for them.  

If someone had this option disabled then they would see the red X indicating that it's disabled which isn't helpful and is likely to create a support burden from users wondering why it's there.  

I'm therefore in favor of enabling this by default and not respecting the previous setting.

PR #1909, opened 2018-09-17
Remove the Link From Virtual Robots.txt option #535

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/535


arnaudbroes:
@contactashish13 I added the requirements for this to #535. Currently, the XML Sitemap menu is correctly indicating whether or not the sitemap link is added to the virtual robots.txt file. However, we also need to automatically the sitemap link to the virtual robots.txt file by default. Then there should be filter hook (no longer a setting) for users who want to disable this.

wpsmort:
@contactashish13 I tested this and when I first updated my sitemap the indicator showed a green checkmark and the link to my sitemap appeared in the default WordPress robots.txt.  However, when I changed settings for my sitemap by unchecking Enable Sitemap Indexes and checking Create Compressed Sitemap I got a red X and the link to my sitemap no longer appeared in the default WordPress robots.txt.  I couldn't figure out why the link suddenly disappeared and the red X was not helpful.  I also tested on a brand new site and couldn't get a green checkmark.

I think we need to output something in the debug log or in our own log file (if Log Important Events) is enabled) to say why the link could not be added to the virtual robots.txt.

Also, please provide an example of how to use the new filter.

contactashish13:
@wpsmort have fixed this and I have added the filter details in the issue.

contactashish13:
@wpsmort except for the first 2 lines (which I've fixed), this seems to be an preexisting bug. Please confirm

wpsmort:
This is tested and is good.  The PHP notices and warning were gone and the filters both worked.

Ready for code review.

michaeltorbert:
On a brand new installation, when activating the sitemap module, I get the following:

`( ! ) Warning: array_key_exists() expects parameter 2 to be array, null given in /app/public/wp-content/plugins/all-in-one-seo-pack-cb979456707e7a0eb1b0a7c29694490b8bc7a84c/inc/aioseop_functions.php on line 1110
--`

This is likely from there being no module information saved to the db yet.

Issue #413, opened 2016-07-22
Remove our current Schema.org markup


wpsmort:
We currently add two lines of Schema.org markup to the HTML tag as follows:

```
<html lang="en-US"
    itemscope 
    itemtype="http://schema.org/WebSite" 
    prefix="og: http://ogp.me/ns#"  class="no-js">
```

These two lines need to be removed:
    itemscope 
    itemtype="http://schema.org/WebSite" 

This will leave the HTLM tag looking like this:

`<html lang="en-US" prefix="og: http://ogp.me/ns#" class="no-js">`

We will be replacing this with the new JSON+LD markup, refer to issue #72 and #73  for this.

wpsmort:
This is now causing issues with the AMP markup validator, reported here - https://wordpress.org/support/topic/problem-with-amp-plugin/

wpsmort:
When we remove this, we need to remove the code added under issue #575 to detect the AMP plugin.

contactashish13:
PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1911 (as this is related to #407)

michaeltorbert:
@wpsmort Does this need to be done at the same time as  #72 and #73 or is it fine to push this out prior to those?

PR #1911, opened 2018-09-17
Remove our current Schema.org markup #413

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/413


arnaudbroes:
This looks good and doesn't cause issues with the AMP plugin.

Issue #1918, opened 2018-09-17
Change Help Icon to Icon-Font


EkoJR:
![help icons](https://user-images.githubusercontent.com/2794445/45653356-8edf8080-ba8c-11e8-93af-77d6912524bb.png)

Related to #289

EkoJR:
Should use WP's Dashicon

https://developer.wordpress.org/resource/dashicons/#editor-help

wpsmort:
Related to #289

EkoJR:
Closing this since, as mentioned, we'll be using the built in dashicon.

michaeltorbert:
Removing the 3.0 milestone.

PR #1927, opened 2018-09-17
Change Help Icon to Icon-Font

EkoJR:
Issue #1918

Requires PR #1919


arnaudbroes:
@EkoJR this works fine but it doesn't look that great imo. The icon sticks right against the text. I'd increase the right margin a bit.

image

Issue #728, opened 2017-01-23
admin_print_scripts should be admin_enqueue_scripts


michaeltorbert:
For many if not all of our scripts and styles. This is in aioseop_module_class.php, and maybe some other files as well.
Related to #726

EkoJr:
I noticed while messing with #1394, the scripts were using the old concept for enqueueing scripts, and have ran into this multiple times.

## WP Hooks

* Frontend - [wp_enqueue_scripts](https://codex.wordpress.org/Plugin_API/Action_Reference/wp_enqueue_scripts)
* Backend - [admin_enqueue_scripts](https://codex.wordpress.org/Plugin_API/Action_Reference/admin_enqueue_scripts)
* ~~Login - [login_enqueue_scripts](https://codex.wordpress.org/Plugin_API/Action_Reference/login_enqueue_scripts)~~

PR #1928, opened 2018-09-18
Change to admin_enqueue_scripts

EkoJR:
Issue #728


arnaudbroes:
@EkoJR this breaks the Media Library.

image

michaeltorbert:
When testing, make sure to test the media uploader as well.

EkoJR:
@arnaudbroes The menu icon was also missing. I thought I had a fix and I noticed it had also defaulted to the cog icon.

I had to back-pedal to master to figure out what was occuring with the timing, since there was an incorrect enqueue, and found there was an old fix causing an early hook (when there should be no hook added).

I may end up closing this and rebasing the branch for a new PR in order to try and keep it clean with some of the reverts that occured.

Issue #1943, opened 2018-09-25
nofollow doesn't work when no SEO for a post type


EkoJR:
**Describe the bug**
NoFollow doesn't work when SEO isn't enabled on Custom Post Type (CPT). Similar to #1797

**To Reproduce**
1. Go to AIOSEOP > General Setting.
2. Disable SEO for target CPT.
3. Enable NoFollow for target CPT.
4. Go to post from CPT, and NoFollow is missing.

**Expected behavior**
NoFollow should be added to CPT post/page.

**Version Info (please complete the following information):**
 - WordPress version: 4.9.8
 - All in One SEO Pack version: 2.8
 - PHP version: 7.2

PR #1944, opened 2018-09-25
Fix nofollow not applied when SEO isn't active on post type

EkoJR:
Issue #1943


arnaudbroes:
This has been tested. Needs code review.

Issue #1957, opened 2018-10-08
noindex/nofollow for no SEO on post type only works together


michaeltorbert:
For #1797 and #1943 we made it so noindex/nofollow works when SEO for that post type is disabled, however this only works if you're using both noindex and nofollow. If you're trying to use just one it doesn't work.

To make the logic easier, it may be best to pull this out into its own function(s). Eventually it should be in its own file, rather than junking up that big class file.

EkoJR:
It seems the git merge didn't go as smooth as would of hoped; looks like it duplicated code instead of line changes.

Issue #1797, opened 2018-07-19
noindex doesn't work when no SEO for a post type


michaeltorbert:
While working on https://semperplugins.com/support/configuring-all-in-one-seo-pack-pro/when-to-noindex-media/, I realized that if a post type has "no seo for this post type" selected, we  can't noindex it... it completely takes away the AIOSEOP section

related to #1796

EkoJR:
@michaeltorbert It appears that nofollow is also affected in the same way. Should I include it here, or make a separate issue?

michaeltorbert:
@EkoJR You can make a separate issue, however, depending on what we do to fix it, the fix for this issue may apply to both.

Issue #68, opened 2016-03-29
Custom Post Type Settings - Remove options from Custom Post Type Settings section 


wpsmort:
Completely remove the following options from the Custom Post Type Settings meta box:

SEO for Custom Post Type Settings
Enable Advanced Options
Custom Titles

These should be enabled by default for all.  

Completely remove the Posts and Pages checkboxes from the SEO on only these post types option.    Again, this should be enabled by default for all.

No filters are required here.

nishant981:
we have done with above task.
![4](https://cloud.githubusercontent.com/assets/13958896/14354921/c3e05728-fcfc-11e5-89ed-889e463f9321.png)

michaeltorbert:
@nishant981 Submit the code as a pull request to the nishant branch.

wpsmort:
Updates to this issue:

1). Let's get rid of the Custom Post Type Settings section completely.
2). Let's check for only custom post types that are public (i.e. 'public' => true).
3). Let's add a filter so that users can add other custom post types if they want.

wpsmort:
See also #226

wpsmort:
When we get rid of this we need to ensure it does not affect the NOINDEX for taxonomies under Noindex Settings.  Currently this field only appears when you select Enable Advanced Options.

PR #1962, opened 2018-10-09
Fix noindex & nofollow both required to operate

EkoJR:
Issue #1957

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -3549,20 +3549,13 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 				if ( ! empty( $aioseop_options['aiosp_cpostnoindex'] ) && is_array( $aioseop_options['aiosp_cpostnoindex'] ) ) {
 					$cp_noindex_active = in_array( $post_type, $aioseop_options['aiosp_cpostnoindex'] );
 				}
-
-				if ( is_singular() && ! in_array( $post_type, $wp_post_types ) && ! is_front_page() && ! $cp_noindex_active ) {
-					return false;
-				}
-				if ( is_post_type_archive() && ! is_post_type_archive( $wp_post_types ) ) {
-					return false;
-				}
-			
 				// Make sure nofollow works even of SEO is turned off for that post type.
 				$cp_nofollow_active = false;
 				if ( ! empty( $aioseop_options['aiosp_cpostnofollow'] ) && is_array( $aioseop_options['aiosp_cpostnofollow'] ) ) {
 					$cp_nofollow_active = in_array( $post_type, $aioseop_options['aiosp_cpostnofollow'] );
 				}
-				if ( is_singular() && ! in_array( $post_type, $wp_post_types ) && ! is_front_page() && ! $cp_nofollow_active ) {
+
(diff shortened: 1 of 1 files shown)


wpsmort:
@EkoJR I tested this and although it does add back the noindex and nofollow meta tags, it also incorrectly adds back the SEO and Open Graph meta tags as well.  I tested this with Posts and with a  Custom Post Type.  

Can you make sure that only the noindex and nofollow meta tags are displayed and all other SEO and OG meta is removed.

EkoJR:
@wpsmort Programmatically, it doesn't seem it will be an easy change. The head function relies on a check to carry out the rest of the operation, which also relies on "SEO on only these post types". However, noindex and nofollow was still needed even if "SEO on only these post types" is unchecked. https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1797#issue-342578925

I think I'll have to change the concept in `\All_in_One_SEO_Pack::wp_head` to still print some data before the boolean check; which I was trying to avoid big changes. I think the only option is to migrate the changes to `\All_in_One_SEO_Pack::wp_head` as well as the rest of the Meta Robot code.

wpsmort:
@EkoJR I tested that and it works fine when you uncheck the box for the post type under Custom Post Type Settings >  SEO on only these post types.  However, if you then select Disabled in the SEO for Custom Post Types option then the SEO meta is added back.

To reproduce, follow these steps:

1. Go to Custom Post Type Settings >  SEO on only these post types and uncheck Posts.
2. View the source code of any post and there should be no SEO meta, only the noindex and nofollow tags should be present.
3. Go to Custom Post Type Settings >  SEO for Custom Post Types and select Disabled.
4. View the source code of your post and the SEO meta should be back.

EkoJR:
@wpsmort It looks like that is a seperate bug. 

The PHP code that exits the wp_head, checks `if ( empty( $aioseop_options['aiosp_enablecpost'] ) )`, but doesn't return a false boolean. It seems that is it still required to have the 'SEO on only these post types' setting.

This would be the code in question in `\All_in_One_SEO_Pack::is_page_included`
```
if ( empty( $aioseop_options['aiosp_enablecpost'] ) ) {
	$wp_post_types = get_post_types( array( '_builtin' => true ) ); // Don't display meta if SEO isn't enabled on custom post types -- pdb.
	if ( is_singular() && ! in_array( $post_type, $wp_post_types ) && ! is_front_page() ) {
		return false;
	}
}
```

EkoJR:
It also seems SEO is added to builtin post types by default when disabled.

EkoJR:
I also tested this with the last release (v2.8), and the effect is also a pre-existing.

wpsmort:
@EkoJR Thanks for pointing that out.  In this case I don't think we want to fix that behavior because we intend to remove that option entirely - See #68.

I tested again with both Posts and a CPT.  Here are the results:

Testing with Posts:
- This works fine.  The noindex/nofollow meta tag is output correctly when you uncheck this post type in SEO on only these post types.

Testing with a CPT:
- This does not work.  The noindex/nofollow meta tag is output when you uncheck the post type in SEO on only these post types.

Can you fix this for custom post types please.

wpsmort:
@EkoJR Thanks for pointing that out.  

This only affects the built-in post types not CPTs.  Given that we plan to remove that option (issue #68) and it's a preexisting bug, we should not fix it under this PR.  We should clean up that code when we remove the option.

I tested with built-in post types and with CPTs and this works correctly.   

This is now ready for code review.

Issue #1691, opened 2018-05-15
conflict with starfishwp - running their shortcode


michaeltorbert:
Reported by Tevya via email:

> https://starfishwp.com. We've found a conflict with AOI SEO. If you have "Run Shortcodes in Description" checked, under All in One SEO > Social Meta > Advanced Settings, AiO tries to process the shortcode somehow, causing it get blocked and our little "funnel" doesn't get rendered for the shortcode. You can see here: http://wordpress-125652-359449.cloudwaysapps.com/sample-page-1/ what it should look like because AiO SEO is disabled on this page. Then here: http://wordpress-125652-359449.cloudwaysapps.com/sample-page/ is what it looks like with AiO SEO enabled and the above mentioned option turned on.

arnaudbroes:
Confirmed. Has been so since the release of the 'Run shortcodes in auto-generated descriptions' feature. AIOSEOP runs the shortcode in the meta description (and truncates it).

michaeltorbert:
Reopening this, as moving the shortcode down far enough in the content that it's not in the description doesn't help.

michaeltorbert:
That's correct. Arnaud didn't explain it completely. At first glance, it appeared as though when it was running in meta description, that interfered in some way (javascript conflict or whatever) with the one in the content, such that the content one never rendered in the browser. This would mean everything is working on our end as intended.
However, I moved the shortcode down far enough in the content that it wouldn't show up in the meta description, and it was still broken. This suggests there's a problem on our end.

michaeltorbert:
Note for whoever works on this issue:
at first glance in the code, it appears we could probably just move our do_shortcode on the content down a bit to only run on the description at the end.
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L2583

tevyaw:
Okay, great! Thanks.

tevyaw:
Any updates on this?

michaeltorbert:
@tevyaw No, you'll see updates in the issue here. No need to ask.

EkoJR:
@michaeltorbert That seems to be all that was needed. 

To summarize it: Currently it relies on the description to be empty in order to generate one, which is also nested with do_shortcodes. If the aioseop description was typed manually, then it just returns that without do_shortcodes.

wpsmort:
[...]

PR #1967, opened 2018-10-11
Fix not running shortcodes in post_meta _aioseop_description

EkoJR:
issue #1691


wpsmort:
I have emailed Tevya to request a copy of his StarfishWP plugin pro version so I can test this.

wpsmort:
@EkoJR I tested this and it works great.  However, we need to apply the same fix to the Run Shortcodes in Description option in Social Meta > Advanced Settings.  Enabling that option causes the same problem with the Starfish Reviews plugin. 

I'll send you the details of my test site on Slack.

EkoJR:
Invalid and breaks a concept by adding shortcodes to manual aioseop descriptions

Issue #1995, opened 2018-10-22
Autogenerated descriptions with shortcodes affected by 2.9 release


wpsmort:
Reported by Iskan by email.

Iskan uses the Enfold theme which as a visual page builder called Avia.  After updating from Pro v2.9 to v2.10 his autogenerated SEO and OG descriptions changed to include the shortcodes from Avia.  

Here's the SEO description autogenerated by Pro v2.9:

`Removed at customer request`

Here's the SEO description autogenerated by Pro 2.10:

`<meta name="description"  content="[av_one_full first min_height=&#039;&#039; vertical_alignment=&#039;&#039; space=&#039;&#039; custom_margin=&#039;&#039; margin=&#039;0px&#039; padding=&#039;0px&#039; border=&#039;&#039; border_color=&#039;&#039; radius=&#039;0px&#039; background" />`

If I enable the Run Shortcodes in Autogenerated Descriptions option then two things break:

1. It breaks the output on the page.  One of the content blocks containing a JavaScript embed doesn't show on the front end.  
2. It outputs a different autogenerated SEO and OG description which still includes shortcodes.

Related to #1990 and #1677.

# Updated

### To Reproduce
1. Go to AIOSEOP General Settings, and enable...
    1. Autogenerate Descriptions.
    2. Use Content For Autogenerated Descriptions.
    3. Run Shortcodes In Autogenerated Descriptions.
2. Go to AIOSEOP Social Meta, and enable...
    1. Run Shortcodes In Description.
    2. Use Content For Autogenerated OG Descriptions.
3. Activate Fusion Plugins & Enfold Theme.
4. Go to "Add New" Post/Page.
5. When using the page builder, DO NOT click "Use Fusion Builder" (plugin). This step will use the Theme's code; which is where the code in question exists.
6. Be sure to add "Layout Elements" to put the Content & Media Elements into; Layout Elements is what creates `avia_codeblock_placeholder` shortcode.
7. Save/Update post/page.

wpsmort:
I've confirmed it's the code change in PR #1964 that broke this.

wpsmort:
@EkoJR @contactashish13 Here is a synopsis of this problem:

SEO / Open Graph descriptions can be created in three ways:

1).  We auto generate the description from the excerpt.  This is the behavior if Autogenerate Descriptions is enabled.

[...]

PR #1964, opened 2018-10-09
stop stripping brackets from description #1035

michaeltorbert:
issue #1035

Testing notes: 
-make sure no other bracket functionality was broken in any of the manually entered or automatically generated descriptions.

PR #1998, opened 2018-10-23
Fix shortcodes affecting Auto Descriptions

EkoJR:
Issue #1995

## Proposed changes
This semi-restores changes that occurred in #1964. This uses WP function for stripping shortcodes; [strip_shortcodes()](https://developer.wordpress.org/reference/functions/strip_shortcodes/).

### Commit Desc
Strips valid shortcodes and preserves non shortcodes. Includes removing end-shortcodes ( `[/shortcode]` )

This can still allow shortcodes that aren't registered/enabled (plugin isn't active)


wpsmort:
@EkoJR I have tested this on two sites. It worked fine on the site running the Avada theme and Fusion Builder plugin (Issue #2004).  It did not work with the site running Enfold.  I have updated issue #1995 with my testing results.

Issue #2011, opened 2018-10-25
PHP Timeout Issues in v2.9.1


EkoJR:
There are still reports of sites breaking, and is partially related to #1997 which was to fix this (different version though).

`get_image_attributes` was the other function that was being spotted at the time, and is still showing up on xDebug Profile (using QCacheGrind).

EkoJR:
Duplicate of #2008

Issue #2008, opened 2018-10-24
Server Error 503 (Max Time/Memory) in v2.9.1


wpsmort:
This is related to #1997.

Whilst testing PR #2007 I was able to reproduce the fatal error below reliably on a clone of a client site.   The site has nearly 8,000 posts and was using static XML sitemaps.  When I installed AIOSEOP v2.9 the site crashed with a 503 error and these errors:

PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 72 bytes) in /wp-content/plugins/all-in-one-seo-pack-1997/modules/aioseop_sitemap.php on line 3089

WordPress database error Lost connection to MySQL server during query for query SELECT * FROM hon_posts WHERE ID = 27621 LIMIT 1 made by do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, AIOSEOP_Updates->version_updates, do_action('after_doing_aioseop_updates'), WP_Hook->do_action, WP_Hook->apply_filters, call_user_func_array, All_in_One_SEO_Pack_Sitemap->do_sitemaps, All_in_One_SEO_Pack_Sitemap->do_simple_sitemap, All_in_One_SEO_Pack_Sitemap->get_simple_sitemap, All_in_One_SEO_Pack_Sitemap->get_all_post_priority_data, All_in_One_SEO_Pack_Sitemap->get_prio_from_posts, All_in_One_SEO_Pack_Sitemap->get_images_from_post, wp_get_attachment_image_url, wp_get_attachment_image_src, image_downsize, wp_attachment_is_image, wp_attachment_is, get_post, WP_Post::get_instance

[...]

PR #1907, opened 2018-09-15
XML Sitemap - add title and caption for images #1753

contactashish13:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1753

PR #2012, opened 2018-10-25
Revert Sitemap add title & caption for images Issue-1753 PR-1907

EkoJR:
Issue #2008 ~~#2011 (duplicate)~~

This reverts PR #1907

With my environment, I was getting the following time/memory; using xDebug Profiler Enabled & QCacheGrind.

| Branch        | Time Cost          | Memory Cost |
| ------------- |:-------------:| -----:|
| Master      | 116,053,444 | 939,026,568 |
| i-2011 (This PR)      | 70,269,334      |   684,063,000 |

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -3027,8 +3027,6 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 					if ( $image ) {
 						$images['image:image'] = array(
 							'image:loc'     => $image,
-							'image:caption' => wp_get_attachment_caption( $thumbnail_id ),
-							'image:title'   => get_the_title( $thumbnail_id ),
 						);
 					}
 				}
@@ -3071,8 +3069,6 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 				if ( $attributes ) {
 					$images[] = array(
 						'image:loc'     => $this->clean_url( $attributes[0] ),
-						'image:caption' => wp_get_attachment_caption( $post->ID ),
-						'image:title'   => get_the_title( $post->ID ),
 					);
 				}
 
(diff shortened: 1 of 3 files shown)


wpsmort:
@EkoJR I tested this PR on the site where I can reliably reproduce the memory issue with static XML sitemaps and I got this error when trying to generate the sitemap by clicking the Update button:

PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 72 bytes) in /wp-content/plugins/all-in-one-seo-pack-2011/modules/aioseop_sitemap.php on line 3085

However, please note that I did the same test with AIOSEOP v2.8 and got the same result but with this error:

PHP Fatal error:  Maximum execution time of 30 seconds exceeded in /wp-includes/taxonomy.php on line 0

I think we have to consider either getting rid of static sitemaps or enabling sitemap indexes by default, as per my comments here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2008#issue-373697660

EkoJR:
@wpsmort with Maximum Execution Time, it's going to be quite difficult to nail it down any further unless optimization becomes a key focus, especially while scanning. With the bigger CPU/Memory heaps gone, it's probably going to become more random with what error stack trace is returned. Right now, that error just points to a simple `if` statement. https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/996ae09c41aa66a14ac9ab744d780d9046a9f683/modules/aioseop_sitemap.php#L3085

30 seconds is a bit low, but when running a site it is considered ideal to keep load times low.

michaeltorbert:
@wpsmort @EkoJR  Does this issue get resolved if we remove static sitemaps and and enable indexes?

EkoJR:
@michaeltorbert I believe the issue still can occur when creating the static sitemap; which could prevent the sitemap from being created. However, if an admin can get past the memory heap, then it would avoid continuous crashes, but then there's concern when content is added/updated.

EkoJR:
Closing this in favor of #2051

Issue #2015, opened 2018-10-25
Move JS for Preview Snippet to a JS file


EkoJR:
Currently the script is stored in a PHP Array. This should be in a JS file.

PR #2016, opened 2018-10-26
Change Preview Snippet JS to a JS file

EkoJR:
Issue #2015


wpsmort:
@EkoJR I tested this and found that he character counter doesn't work for the Home Title and Home Description fields in Home Page Settings.  It work fine on the Edit Post screen.

wpsmort:
@EkoJR I get this notice when I click Update Options on the General Settings screen:

Undefined variable: title_format in all-in-one-seo-pack/aioseop_class.php on line 1450

wpsmort:

aioseop_class.php
```diff
@@ -3335,6 +3347,34 @@ public function admin_enqueue_scripts( $hook_suffix ) {
 		add_filter( "{$this->prefix}display_settings", array( $this, 'filter_settings' ), 10, 3 );
 		add_filter( "{$this->prefix}display_options", array( $this, 'filter_options' ), 10, 2 );
 
+		// This ensures different JS files are enqueued only at the intended screens. Preventing unnecessary processes.
+		$extra_title_len = 0;
+		switch ( $hook_suffix ) {
+			// Screens `post.php` & `./aioseop_class.php` share the same `count-char.js`.
+			case 'post.php' :
```
contactashish13: maybe add post-new.php so that it works for new posts as well

aioseop_class.php
```diff
@@ -3335,6 +3347,34 @@ public function admin_enqueue_scripts( $hook_suffix ) {
...
+				$info         = $this->get_page_snippet_info();
+				$title        = $info['title'];
+				$title_format = $this->get_title_format( array( 'name' => 'aiosp_snippet' ) );
+
+				if ( ! empty( $title_format ) ) {
+					$replace_title   = '<span id="aiosp_snippet_title">' . esc_attr( wp_strip_all_tags( html_entity_decode( $title ) ) ) . '</span>';
+					$extra_title_len = strlen( $this->html_entity_decode( str_replace( $replace_title, '', $title_format ) ) );
+				}
+			case 'toplevel_page_all-in-one-seo-pack/aioseop_class' :
```
contactashish13: when we use grunt in the future, it will complain as there is no comment indicating a fall through from one case to another so add a comment `// fall through` so that it is also obvious to future reviewrs

js/count-chars.js
```diff
@@ -0,0 +1,88 @@
+/**
```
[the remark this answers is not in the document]
EkoJR: I'm still concerned about adding this, and would much rather create a different Issue/PR since it essentially affects the end result. Plus, tried it before and had to remove it. https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1767/files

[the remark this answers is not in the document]
EkoJR: ...but that did make me notice I forgot to rename aioseop_post_edit. Since it's not just for post edit now.

aioseop_class.php
EkoJR: Ah, thank you, good catch!

aioseop_class.php
EkoJR: Ok cool. I wasn't sure how to comment on it other than above the 2 cases.

aioseop_class.php
```diff
@@ -3350,8 +3350,9 @@ public function admin_enqueue_scripts( $hook_suffix ) {
 		// This ensures different JS files are enqueued only at the intended screens. Preventing unnecessary processes.
 		$extra_title_len = 0;
 		switch ( $hook_suffix ) {
-			// Screens `post.php` & `./aioseop_class.php` share the same `count-char.js`.
+			// Screens `post.php`, `post-new.php`, & `../aioseop_class.php` share the same `count-char.js`.
 			case 'post.php' :
+			case 'post-new.php' :
```
contactashish13: add `// fall through` here as well

js/count-chars.js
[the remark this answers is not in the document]
contactashish13: @EkoJR not sure how it affects the end result. Am I missing something here?

aioseop_class.php
EkoJR: I don't think it does on this one, are you absolutely sure? https://eslint.org/docs/rules/no-fallthrough

js/count-chars.js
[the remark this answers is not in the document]
EkoJR: Yes, a different Issue/PR after this one gets merged. We've already been through this before where we had to go back and strip out code that was added into a PR in the same manner (link above was when it was removed). I would much rather have a seperate PR merge/squash into the main branches. Makes repo more organized and easier to handle if it does go wrong again.

aioseop_class.php
contactashish13: @EkoJR grunt might not dislike it, I'm not sure, but I find it helpful to state a fall through

js/count-chars.js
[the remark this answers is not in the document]
contactashish13: @michaeltorbert do you want to create a separate issue for this? Maybe something that is relevant to all JS files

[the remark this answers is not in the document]
michaeltorbert: @contactashish13 @EkoJR Sure a separate issue is fine. You can go ahead and do that.

Issue #2014, opened 2018-10-25
XML Sitemap doesn't load after updating to v2.9


wpsmort:
Reported by Miguel Emmara by contact form on 10/23.

Miguel reported that after updating to v2.9 and v2.9.1 that individual sitemaps in his indexed sitemap wouldn't load.  I as able to verify this on his site and that it works fine with v2.8.

I noticed that he has over 61GB of images in his uploads directory and when I exclude images from the XML Sitemap it works fine.

wpsmort:
Possibly related to #2008 and #2011

HOTDOG2705:
Hello, I have the same problem, with image selected no sitemap! With 2.8 worked well.. I'll might close the other thread. 
Thanks!

wpsmort:
@HOTDOG2705 Can you send us an Administrator login for your site and FTP login for your server using the secure form below so we can troubleshoot this:

https://semperplugins.com/contact/support-login-information/

wpsmort:
Reported again here - https://semperplugins.com/support/general-all-in-one-seo-pack-pro-forum/sitemap-xml/

I got a login for this site and confirmed that it has 2200 posts with 8800 images.  Dynamic Sitemaps was enabled but Sitemap Indexes was not.  Even after enabling Sitemap Indexes and Pagination, trying to view a sitemap resulted in a 504 error.  However, excluding images resolved the problem.  Rolling back to the last version resolved the problem.

wpsmort:
@EkoJR @contactashish13 I've set up separate clones of a site where this issue can be reproduced and sent you login information via Slack.

EkoJR:
@wpsmort I found this is the same as #2008 and PR #2012 does fix it for this issue.

wpsmort:
@contactashish13 I have tested PR #2022 on both cloned sites where we could reproduce this problem.  It works fine on both sites, resolving the problem with the XML Sitemap timing out and with the 503 errors reported in #2008.  However, I do see that it still takes a long time to generate the sitemap. 

Here are the timings and memory usage from both sites:

**Site 1 - Dynamic sitemap with indexes and pagination (21205 URLs and 32,204 images)**
**AIO v2.8**  -  25.56 MB memory used generating the dynamic sitemap in 1.059 seconds, 37.00 MB total memory used.
**AIO PR2022**  -  68.51 MB memory used generating the dynamic sitemap in 0.392 seconds, 80.04 MB total memory used.
[...]

PR #2022, opened 2018-10-27
cache query to determine image attributes

contactashish13:
#2014


wpsmort:
@EkoJR Does your work with PR #2051 replace this or supplement this PR?

EkoJR:
@wpsmort This PR was created separately to accomplish the same goal (I think it was posted a few days before mine), but yea, it does modify the same area of code.

However, at one point I did take this code into consideration, and tried to include the concept as well.

wpsmort:
I'm closing this as @EkoJR is working on a PR #2093 that replaces this.

Issue #2037, opened 2018-11-02
Change/Remove get_facebook_debug() function to enqueue JS instead


EkoJR:
Currently, in `All_in_One_SEO_Pack_Opengraph::get_facebook_debug`, JavaScript is being printed via function using `ob_start()`. The JS should be in a JS File to be enqueued, and move code to remove class method since the only justification seemed to be with JS.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/bfef6379ed054a52ff858967fff5e8ead6218eb9/modules/aioseop_opengraph.php#L1728

michaeltorbert:
@EkoJR Does this fix a recently introduced issue? If not we should have this in 3.0.

wpsmort:
@michaeltorbert This is related to https://github.com/semperfiwebdesign/aioseop-pro/issues/450.  The move of this code is required in order to get the patch for that issue to work.

PR #2038, opened 2018-11-02
Change/Remove method get_facebook_debug() to enqueue JS instead

EkoJR:
Issue #2037

## Proposed changes

Previously located in `All_in_One_SEO_Pack_Opengraph::get_facebook_debug()`. The JS code was changed be in a JS File instead, and remaining code was moved to where it was being called; since the only justification was with JS.

For testing: This seemed to only related to the admin Post Edit screen > AIOSEOP Social Settings > Facebook Debug.


wpsmort:

PR #2044, opened 2018-11-06
Properly escape image upload variables

EkoJR:
Related to https://github.com/semperfiwebdesign/aioseop-pro/issues/450

The text `'><script>alert(1)</script>'` in this case is the expected bug being created in Custom Twitter Image with the test hook/function. The script is prevented from executing, but renders as text instead; as far as I know, that is expected, and the source of the problem should be identified and removed. I don't think it can be exploited any further though.

i-450-social-settings


wpsmort:
@EkoJR I tested this but there is still script being output in the Custom Twitter Image field.  See screenshot:

screen shot 2018-11-06 at 5 57 09 pm

Issue #2029, opened 2018-10-28
Disappearing SEO meta section in 2.9.x


michaeltorbert:
Reported by _many_ people.

Steps to reproduce:
1) Install 2.8. 
2) Check posts and pages under SEO for Custom Post Types (Content Types in 2.9)
3) Disable "SEO for Custom Posts Types" radio.
4) Verify that the SEO meta section appears in wp-admin for edit posts and on the post/page on the front-end.
5) Update to 2.9.
6) SEO meta section disappears.
7) Enabling SEO for Content Types makes them reappear.

We should investigate what in 2.9 caused this behavior to change. I suspect it has something to do with the addition of is_seo_enabled_for_cpt() but someone should confirm this.

First, we need to determine conclusively what change in 2.9 caused this behavior change.

Second, we need to decide what to do. My .02 is that we simply remove "SEO fo Content Types" and just have the check boxes. https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2028
This would require going through all uses of the aiosp_enablecpost option and adjusting as needed.

contactashish13:
I can confirm its because of that method. If I return true, the meta is shown.

If the radio button `disabled` is checked, does that not mean that no meta should show and therefore the behavior is correct?

To continue with previous behavior, we have 2 options 
- remove the radio buttons unless they serve any other purpose besides toggling the view in the settings screen
- make https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L3868 an `OR` conditions instead of an `AND` viz.

```
return 'on' === $aioseop_options['aiosp_enablecpost'] || in_array( get_post_type(), $aioseop_options['aiosp_cpostactive'], true );
```
instead of
```
return 'on' === $aioseop_options['aiosp_enablecpost'] && in_array( get_post_type(), $aioseop_options['aiosp_cpostactive'], true );
```

wpsmort:
Reported again here - https://wordpress.org/support/topic/problems-with-home-page-settings/

After getting a login for the users site, I found that SEO for Custom Post Types was disabled.  Enabling it rectified the problem.

contactashish13:
@wpsmort can you please comment on my observations?

wpsmort:
@contactashish13 To answer your first question: 

[...]

Issue #2028, opened 2018-10-28
Remove option "SEO for Content Types"


michaeltorbert:
I'm not sure why we need this.  @wpsmort?

![seo for content types](https://user-images.githubusercontent.com/2794445/47691531-b4919680-dbaf-11e8-8098-f1bffbd4d274.png)

arnaudbroes:
I'd like to see it removed as well to be honest.

EkoJR:
What about plugins that use Custom Post Types to store their data, and forcing SEO on those is needless processing and irrelevant data? Allowing others to disable/enable Post Types would prevent that.

For each Custom Post Type adds additional processing, and for some big sites it would have to deal with all the post types.

michaeltorbert:
@EkoJR It's not that we'd take away the check boxes to enable/disable SEO meta per post type, only the radio for all of them. It's an unneeded extra layer.

EkoJR:
In that case, I would agree then. That setting has always felt redundant.

wpsmort:
I'd really like to get rid of this option.  I'd also like to get rid of the Enable Advanced Options radio and the Custom Titles checkbox.  I see no reason for any of them to be there.  

We have a trim options issue open for this here - #68

arnaudbroes:
@wpsmort yes, yes and yes!!

PR #2050, opened 2018-11-09
Remove option "SEO for Content Types"

contactashish13:
#2028 and #2029


wpsmort:
@contactashish13 I tested this and found two problems as follows:

1.  The checkboxes for custom post types are not being output in these settings:
 - Display Settings > Show Columns Labels 
 - Google Settings > Display Google Authorship
 - Noindex Settings > Default to NOINDEX
 - Noindex Settings > Default to NOFOLLOW

See screenshot below:

screen shot 2018-11-12 at 2 27 24 pm

2. The title format for custom post types are coming from the Post Type Format field and not the Title Format field under Content Type Settings.  

To reproduce this, enter the format %post_title% - POST | %blog_title% in the Post Title Format field and then enter the format %post_title% - MEDIA | %blog_title% in the Media Title Format field.  View the source code of any media attachment page and you'll see that the title tag format is coming from the Post Type Format field and not the Media Title Format field.

wpsmort:
@contactashish13 I have tested your latest commit and both bugs I reported above are still present.  Issue 1 does not exist in the master branch, it only exists in this PR.  I've set up a test site and sent you the login on Slack so you can see what I'm seeing.

contactashish13:
@wpsmort once I reset all the settings I was able to reproduce both issues at my end and have fixed them in the latest commit. I have also applied my changes to the cloudways instance and tested successfully.

wpsmort:
This is tested and is good. I've tested every combination of options and tested updating from v2.8 and v2.9.1 to make sure nothing breaks. This is ready for code review.

Issue #2055, opened 2018-11-14
No SEO output on homepage when there are no posts


wpsmort:
Reported by David Tiong (by email) and Shaul Hader (https://wordpress.org/support/topic/after-the-update-the-plug-in-does-not-work/)

The bug occurs in the v2.9 release and happens when Settings > Reading > Your homepage displays is set to Your latest posts.  If there are no posts on the site then AIOSEOP doesn't output anything in teh source code of the homepage.  This bug was introduce in v2.9.

To reproduce this follow these steps:

1. Create a new site with no posts
2. Go to Settings > Reading and set Your homepage displays to Your latest posts
3. Go to All in One SEO Pack > Home Page Settings and add a title and description
4. View the source code of the homepage and you'll see that nothing is being output by AIOSEOP
5. Add a post and then view the source code of the homepage and you'll see that now AIOSEOP is outputting the title and description

wpsmort:
@contactashish13 This bug was introduced in v2.9.  Please let me know ASAP if you need me to demonstrate how to reproduce this.

contactashish13:
@wpsmort I have fixed this. PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2058

Issue #1616, opened 2018-03-20
PHP Notice: Trying to get property of non-object in aioseop_class.php on line 1616


michaeltorbert:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L1616

Reported here: https://wordpress.org/support/topic/php-notice-trying-to-get-property-of-non-object-13/

It's an easy fix, but I'm curious under what circumstances we'd get that warning. The function shouldn't be getting called when there's no global $post.

Spreeuw:
Getting the same issue with 2.5 on my front page (none of the other pages throw this notice)

davettt:
I see it in the debug logs on sites where WordPress settings are 'Settings > Reading ->Front page displays' is set to "Your latest posts". I think it happens around line 1629 of aioseop_class.php when $post->ID is called. 

This occurs in the get_aioseop_title function as the is_front_page() conditional returns true but $post is null, for themes that use the default WordPress reading settings to display their home pages.

wpsmort:
This comment is copied from duplicate issue #678:

In aioseop_class.php on line 1624, we have

global $post;
$post_id = $post->ID;

Reported here, https://wordpress.org/support/topic/trying-to-get-property-of-non-object-35, there could be conditions in certain edge cases where there's no $post
We could consider checking if $post has anything, and if not, maybe bail out of the function gracefully?

For dev and testing, we'd need a theme that displays the posts without there being global $post.

djboris88:
This is how I fixed it, since I believe it happens only if the static homepage is being loaded.

These are lines 1622-1626 in [`All_in_One_SEO_Pack::get_aioseop_title`](https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/1eb745292046d926fecf7ecf65663807cde4cce3/aioseop_class.php#L1622-L1626), latest version:
```php
global $post;
$frontpage_id = get_option( 'page_on_front' );

if ( $post === null ) {
	$post_id = $frontpage_id;
} else {
	$post_id = $post->ID;
}

if ( is_post_type_archive() && is_post_type_archive( 'product' ) && $post_id = wc_get_page_id( 'shop' ) && $post = get_post( $post_id ) ) {
	// $frontpage_id = get_option( 'page_on_front' ); - removed
```

So if the global `$post` is null, I suppose it is the static frontpage, since on every other page there will be something loaded as the global `$post`.

[...]

PR #2058, opened 2018-11-15
No SEO output on homepage when there are no posts

contactashish13:
#2055


wpsmort:
This is tested and is good.  I've tested every combination of options and tested updating from v2.8 and v2.9.1 to make sure nothing breaks.  This is ready for code review.

wpsmort:
This is tested and is good.  I tested every combination of settings for the homepage both in WordPress and in AIOSEOP.  Whilst testing I did get the PHP notice detailed here #1616. 

Issue #2059, opened 2018-11-15
Clean AIOSEOP_Sitemap with WPCS


EkoJR:
Clean up sitemap using WPCS ( WordPress Code Sniffer ).

PR #2060, opened 2018-11-15
Cleanup Sitemap with WPCS

EkoJR:
Issue #2059

No changes in operations; TODOs were added instead. This was to inspect the sitemap code, and find any potential issues through code sniffing.

TODOs added...
* Fix undefined variables.
* Add esc_* functions.
* Change `parse_url()` to `wp_parse_url()`.
* Add 'true' in 3rd arg with `in_array()`.
* Add nonce verification.
* Add/Change `$_POST` to `filter_input()`.
* Change `urlencode()` to `rawurlencode()`.


wpsmort:
@EkoJR I tested this and found one problem:

The Excluded Pages option doesn't work when you enter a post ID.  It does work if you enter a slug.  For example, if I enter post ID 2 then this post is still included in the sitemap.  But if I enter the slug hello-world then the post is excluded.

EkoJR:
@wpsmort It looks like changing the hardcoded numberic check to a strict comparison was the problem. Should be good now.

wpsmort:

EkoJR:
Closed to rebase #2084

Issue #2065, opened 2018-11-19
XML Sitemap - Characters aren't being encoded for image:caption


wpsmort:
Reported here - https://semperplugins.com/support/troubleshooting-all-in-one-seo-pack-pro/google-webmaster-parsing-error-in-xml-sitemaps/#p8416

When you have a character such as an ampersand in the image caption, it doesn't get encoded correctly and therefore Google Search Console returns a parsing error.  

To reproduce, add a caption to an image that has an ampersand in the caption, i.e. "my caption has an & ampersand".  Then look at the source code of the XML sitemap and you'll see that the ampersand is displayed as a character instead of the HTML character code.

wpsmort:
@contactashish13 Can you ensure that the same encoding is performed on the <image:title> and <image:caption> tags so that HTML character codes are output in the source code of the XML Sitemap and Video XML Sitemap.  See the section on Non-alphanumeric and non-latin characters here  - https://support.google.com/webmasters/answer/183668?hl=en

PR #2070, opened 2018-11-20
XML Sitemap - Characters aren't being encoded for image:caption

contactashish13:
#2065


arnaudbroes:
This has been tested. Ready for code review.

arnaudbroes:
@contactashish13 unfortunately it looks like your fix introduced another bug. If you enable Sitemap Indexes in the XML Sitemap module and then go to the sitemap index, you'll see in the source code that each sitemap has `<changefreq>` and `<priority>` tags, which should not be the case.

So in order to reproduce this behaviour:

1. Install this PR
2. Go to the XML Sitemap module and enable Sitemap Indexes
3. Go to /sitemap.xml and view the source code

contactashish13:
@arnaudbroes I've addressed this.

Issue #2057, opened 2018-11-14
RSS Sitemap should be optional and disabled by default


wpsmort:
Whilst troubleshooting issues with the XML sitemap (issues #2008 and #2014), we looked into whether the addition of the RSS sitemap in v2.9 contributed to the issues being reported by users.

As the RSS sitemap is not a requirement of any search engines, we should make this feature an option in the XML Sitemap settings and it should be disabled by default.

michaeltorbert:
Are the performance problems caused by the RSS sitemap functionality only on large sites? 
If so, I'd recommend that the documentation (on semperplugins.com and in AIOSEOP) warn that this option can cause performance issues on large sites.

wpsmort:
@EkoJR is going to look into this and see if the RSS sitemap is causing or contributing to the performance problems.

contactashish13:
@wpsmort I had also suggested on call that we can actually count how many elements are going to be included in the sitemap and on the basis of that we can recommend whether the user should enable indexes or not. Let me know if you think that needs to be done as well.

contactashish13:
PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2074

contactashish13:
@wpsmort shouldn't the module be renamed to Sitemaps rather than XML Sitemaps as this can generate RSS as well?

arnaudbroes:
@contactashish13 feel free to open an issue to discuss it.

PR #2074, opened 2018-11-21
RSS Sitemap should be optional and disabled by default

contactashish13:
#2057


arnaudbroes:
@contactashish13 the option doesn't seem to influence the generation of the RSS sitemap; only the link that's displayed when the option is enabled. When you disable the option, the sitemap is still dynamically being generated, but you need to directly navigate to it.

@wpsmort currently, Ashish added "Create RSS sitemap as well" as tooltip documentation. Do you want to fleshen that out a bit?

contactashish13:
Please give me an an example with the settings you have chosen. Screenshot would help too

contactashish13:
@arnaudbroes I tried on a new site as well and cannot replicate the problem. From the problem statement, it appears you already had the plugin installed (or there were some settings already present) because of which the rewrite rules already existed. That is why instead of getting a 404 you are seeing the sitemap when you navigate to the RSS sitemap URL. 

You can try on this public instance where I have installed the plugin but not activated it. You can activate it, enable sitemap, go to the RSS sitemap and you will get a 404. 
```
URL: http://splendid-salamander.w6.poopy.life/wp-admin/
Login: demo
Pass: 11Qw9rzqZJ4F
```

arnaudbroes:
@contactashish13 correct, but when the option is disabled, the RSS sitemap continues to be updated when posts are added/modified. Is that supposed to happen?

contactashish13:
It's a dynamic sitemap so if the URL is accessible, it will fetch posts. 

A fresh install should be a fresh install without prior data. I'm not sure I understand how you tested. Can you please clarify?

arnaudbroes:
@contactashish13 that's the point, the URL shouldn't be accessible once the option is disabled.
Follow these steps to reproduce the issue:

1. Create a fresh test site
2. Install the plugin and enable the XML Sitemap module
3. Go to domainname.com/sitemap.rss. You'll get a 404 (which is supposed to happen).
4. Enable RSS sitemap in the XML Sitemap module and go to /sitemap.rss. It will work.

Now here's when things get interesting:

5. Go back to the XML Sitemap module and disable the RSS sitemap option. Go to /sitemap.rss and you'll see that the RSS sitemap is still being generated. If you create a new post/page, it will be added to the RSS sitemap. This isn't supposed to happen. If the option is disabled, then the RSS sitemap should be disabled as well.

contactashish13:
@arnaudbroes that happens because the rewrite rules have been added to the database and they get refreshed after every 12 hours. This is not a bug.

contactashish13:
@arnaudbroes I cannot reproduce this behavior on any instance. Did you try it on the instance I had shared? If you did not, please create and share a similar public instance where you are able to reproduce the behavior. I'm not sure about the cloudflare instances - they might have some weird varnish cache at work. 

Please also note that there is no cron for clearing the rules. You will need to login.

contactashish13:
@wpsmort I have been able to reproduce and fix this.

PR #2093, opened 2018-12-06
Add sitemap optimization

EkoJR:
Issue #2008

## URL to Post IDs Filter
Delete & Disable Transient Data; for testing peak performance.
```
function aioseop_remove_set_transient_url_postids() {
	remove_action( 'shutdown', 'aiosp_common::set_transient_url_postids' );
	if ( is_multisite() ) {
		delete_site_transient( 'multisite_aioseop_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
	} else {
		delete_transient( 'aioseop_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
	}
}
add_action( 'shutdown', 'aioseop_remove_set_transient_url_postids', 9 );
```

## Query Step Filter
Used to control which query steps are used.

```
/**
 * Skip heavy query step.
 *
 * @param $query_step
 * @return int
 */
function aioseop_post_images_step_1( $query_step ) {
	switch ( $query_step ) {
		case 2:
			$query_step++;
			break;
	}

	return $query_step;
}
add_filter( 'aioseop_post_images_query_step', 'aioseop_post_images_step_1' );

/**
 * Fire target query steps on every load.
 * Note: It's best not to use all 3, and value 2 is a heavy query step.
 *
 * @param $query_step
 * @return int
 */
function aioseop_post_images_step_2( $query_step ) {
	return array( 0, 1, 2 );
}
//add_filter( 'aioseop_post_images_query_step', 'aioseop_post_images_step_2' );

/**
 * On initial step ( 0 ), set the query step to target step. Default set the step as finished.
 *
 * @param $query_step
 * @return int
 */
function aioseop_post_images_step_3( $query_step ) {
	switch ( $query_step ) {
		case 0:
			$query_step = 2;
			break;
		default:
			$query_step = 3;
			break;
	}

	return $query_step;
}
//add_filter( 'aioseop_post_images_query_step', 'aioseop_post_images_step_3' );

/**
 * On initial step ( 0 ), set the query step to do all query steps. Default set the step as finished/skip.
 *
 * @param $query_step
 * @return int
 */
function aioseop_post_images_step_4( $query_step ) {
	switch ( $query_step ) {
		case 0:
			$query_step = array( 0, 1, 2 );
			break;
		default:
			$query_step = 3;
			break;
	}

	return $query_step;
}
//add_filter( 'aioseop_post_images_query_step', 'aioseop_post_images_step_4' );
```

Used to delete transient data
```
function aioseop_remove_set_transient_post_parents_images() {
	global $wpdb;
	remove_action( 'shutdown', array( 'All_in_One_SEO_Pack_Sitemap', 'set_transient_post_parents_images' ) );

	if ( is_multisite() ) {
		$wpdb->get_results( "DELETE FROM `wp_options` WHERE `option_name` LIKE ('multisite_aioseop_posts_images_%')" );
		$wpdb->get_results( "DELETE FROM `wp_options` WHERE `option_name` LIKE ('multisite_aioseop_posts_images_query_step_%')" );
	} else {
		$wpdb->get_results( "DELETE FROM `wp_options` WHERE `option_name` LIKE ('aioseop_posts_images_%')" );
		$wpdb->get_results( "DELETE FROM `wp_options` WHERE `option_name` LIKE ('aioseop_posts_images_query_step_%')" );
	}
}
add_action( 'shutdown', 'aioseop_remove_set_transient_post_parents_images', 9 );
```

### DIFF
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -15,6 +15,13 @@ class aiosp_common {
 // @codingStandardsIgnoreEnd
 
 	/**
+	 * @var null|array
+	 *
+	 * @since 2.9.2
+	 */
+	public static $attachment_url_postids = null;
+
+	/**
 	 * aiosp_common constructor.
 	 *
 	 */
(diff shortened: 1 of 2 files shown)


EkoJR:
Closed to rebased to updated code #2086

wpsmort:
This is tested and does improve performance over v2.9 and v2.9.1.  This is ready for code review.

inc/aiosp_common.php
```diff
@@ -182,4 +189,255 @@ public static function make_xml_safe( $tag, $value ) {
...
+	 * Sets the transient data at the last hook instead at every call.
+	 *
+	 * @see set_transient()
+	 * @link https://developer.wordpress.org/reference/functions/set_transient/
+	 *
+	 * @since 2.9.2
+	 */
+	public static function set_transient_url_postids() {
+		if ( is_multisite() ) {
+			set_site_transient( 'multisite_aioseop_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
```
contactashish13: better to use DAY_IN_SECONDS

modules/aioseop_sitemap.php
```diff
@@ -3376,91 +3396,312 @@ private function get_images_from_post( $post ) {
...
+					// GET BUILT-IN ATTACHMENT OBJECTS.
+					// Query_Step 0 may include Post Thumbnails, but there are a few objects that could be miss.
+					$args = array(
+						'post_type'   => 'attachment',
+						'numberposts' => 1000, //-1, // It's unlikely there will be over 1000 image on a single page.
+						'post_status' => null,
+						'post_parent' => $post->ID,
+					);
+
+					$attachments = get_posts( $args );
```
contactashish13: use WP_Query here as get_posts is inefficient. set cache params to false as well

modules/aioseop_sitemap.php
```diff
@@ -3376,91 +3396,312 @@ private function get_images_from_post( $post ) {
...
-			global $wpdb;
-			$attributes = array();
-
-			$attachment = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE guid='%s';", $url ) );
-			if ( $attachment && is_array( $attachment ) && is_numeric( $attachment[0] ) ) {
-				$attributes = array(
-					'image:caption' => wp_get_attachment_caption( $attachment[0] ),
-					'image:title'   => get_the_title( $attachment[0] ),
-				);
+		public function query_post_images_step( $query_step, $post, $current_images = array() ) {
```
contactashish13: this method can be moved to aiosp_common

contactashish13:
@EkoJR can you fix the travis failure? 

@michaeltorbert @wpsmort on a slight tangent, I feel PRs should only be tested once travis succeeds.

Also, I'm not sure what I'm doing wrong but it does not seem like its working for me. Here are my steps:

1) created 100 pages with image in the content
2) added a simple debug here https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/7406736833424458e6997a35a18dced63d89bbe8/modules/aioseop_sitemap.php#L3473 to print the value of `query_step` 
3) view the sitemap without indexes enabled


My conclusions:

1) first time I load the sitemap, I get `0` printed 100 times in the log. 
2) if my posts contain only images in the content (no featured image in any post), this fails to report any image
3) if I add a featured image to one of the posts, it still does not report any image in the sitemap but the value of `query_step` starts with 1. If I refresh the sitemap again, it shows the featured imaged in the XML (this, I suspect, is because the value of `query_step`  becomes 2, which is what the logs show)
4) Every refresh of the sitemap increases the value of `query_step` but no matter how many times I refresh, the images in the content are reported as 0 in the XML.
5) Incidentally, `query_step`  is always an integer and never an array as expected here https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/7406736833424458e6997a35a18dced63d89bbe8/modules/aioseop_sitemap.php#L3476 because it is always set as an integer in the transient no matter how many times I refresh.

michaeltorbert:
> I feel PRs should only be tested once travis succeeds.
@contactashish13 Agreed. @wpsmort In addition to only testing if there are no merge conflicts, can you make sure the tests all pass as well? 
@EkoJR Let's remember to make sure travis passes before assigning a PR for testing.

EkoJR:
I understood it needed to pass, and I thought I added a commented to this regarding Travis (saying it needed to pass before merging).

In this case, I was...
 
* Trying to provide a prototype for @wpsmort to see if it does pass testing.
* I was ultimately looking for feedback on the concept itself. Possibly with some different ideas.

I actually have a hard time favoring this PR due to the **last 2 commits** which introduces a query step. I believe it will be too complex in the way it operates; it currently takes 3 instances to produce full results.

I was able to reduce URL to ID, but there are still images in the Content that need to be parsed, and converted from URL to ID; which is where most of the memory load occurs at. WP does try to track images added to posts, but doesn't always work for a variety of reasons. So, still end up having to parse the content for images.

contactashish13:
@EkoJR gotcha :)

So are you saying it does indeed need 3 refreshes to parse everything? I was worried I was testing incorrectly.

EkoJR:
@contactashish13 That is correct, and I added in the description hooks to remove the transient data. As well as info to create a hook to execute all query_steps in 1 instance; which is where the `is_array()` comes into play.

EkoJR:
Just a little added info, each query step works as follows...

1. Refer to Attachments set by WP; which doesn't always work for the following reasons
   1. Images can only have 1 post parent (in general, there can only be 1 post parent).
   2. Plugins didn't seem to always assign it with WP as well.
2. Uses Fallbacks without needing URL to ID conversion.
3. Parses content, but it is mostly URL to ID that causes the most load.

wpsmort:
@EkoJR I've assigned this back to you as the Travis checks failed.  

Please be sure to continue the discussion with @contactashish13 about your proposed changes here. Whilst I've tested this on the two clone sites and verified that it does improve performance for those specific sites, I want us to be sure that the approach you're introducing here is the right approach for all sites including large sites and sites with lots of images.  We must be certain we're fixing the problems introduced v2.9.x and not introducing any new ones.

EkoJR:
Closing in favor of #2136

Issue #2097, opened 2018-12-06
Non-static method WP_Feed_Cache::create() should not be called statically


wpsmort:
Reported here - https://wordpress.org/support/topic/error-messages-since-version-2-10/

Reported right after release of v2.10
Non-static method WP_Feed_Cache::create() should not be called statically /wp-includes/SimplePie/Registry.php:215

Issue #2096, opened 2018-12-06
Undefined index: aiosp_cpostactive aioseop_class.php:953


wpsmort:
Reported here - https://wordpress.org/support/topic/error-messages-since-version-2-10/

Reported right after release of v2.10
Undefined index: aiosp_cpostactive /wp-content/plugins/all-in-one-seo-pack/aioseop_class.php:953

EkoJR:
I wasn't able to reproduce this, but I created a PR that may fix it.

martijnrusschen:
Still seeing this issue. Example: https://studio-hoogeveen.nl/.

wpsmort:
@martijnrusschen Can you confirm you are running version 2.10.1 of All in One SEO Pack which has the patch for this.  If you are, then please can you let us know the exact steps to reproduce this and can you send us an Administrator login for your site using the secure form here - https://semperplugins.com/contact/support-login-information/

martijnrusschen:
Yep see below:
![screenshot 2018-12-17 at 5 38 02 pm](https://user-images.githubusercontent.com/1412392/50101216-85e48380-0222-11e9-85c9-677b3b6dc0e2.png)

I'd like to avoid sharing Admin credentials, it seems this started to happen after the Wordpress 5 upgrade. There was an upgrade for the plugin available at that time, which I executed as well, but it didn't resolve the errors I'm seeing.

arnaudbroes:
@martijnrusschen om de waarschuwing te verbergen (aangezien het vermoedelijk om een website van een klant gaat) kan je trouwens in de tussentijd in het wp-config.php-bestand van de website debugging uitschakelen.

PR #2099, opened 2018-12-06
Fix undefined index in aioseop_class.php

EkoJR:
Issue #2096


wpsmort:
This is tested.  I'm also not able to reproduce the PHP notice reported in the issue.  However, I've tested all variations of the Content Type Settings and Display Settings and they all work.  I've tested on PHP 5.2 to 7.2.

Ready for code review.

Ready for code review.

michaeltorbert:
I'm not comfortable pushing a possible fix that we can't test. I've reached out to the user to verify the fix works.

michaeltorbert:
The user has tested and confirmed the fix.

michaeltorbert:
The user reported this fixed #2097 also.

Issue #1650, opened 2018-04-10
Bug with video sitemap when added to XML Sitemap indexes 


wpsmort:
Reported here - https://semperplugins.com/support/configuring-all-in-one-seo-pack-pro/sitemap-errors-1/

When a site is running the XML Sitemap module with sitemap indexes enabled and they activate the Video Sitemap module but don't save anything then the video sitemap is added to the main XML sitemap as:

http://domain.com/.xml

It should show up as:

http://domain.com/video-sitemap.xml

To reproduce this bug follow these steps:

1. Install AIOSEOP Pro on a new site
2. Activate the XML Sitemap module and Video XML Sitemap module
3. Go to the Sitemap module and enable Sitemap Indexes and click Update Options
4. View the XML Sitemap and you should see the last sitemap entry is http://domain.com/.xml.gz

arnaudbroes:
@wpsmort  I checked this and this bug was introduced with 2.5 when we decided to include the Video XML Sitemap URL in the regular sitemap.xml.

contactashish13:
@arnaudbroes I think this was introduced much later in 2.7.1. as part of the fix for https://github.com/semperfiwebdesign/aioseop-pro/issues/412. Can you please confirm?

@wpsmort I believe this issue is related to https://github.com/semperfiwebdesign/aioseop-pro/issues/424 and is fixed by #1887.

arnaudbroes:
@contactashish13 I reconfirmed that this issue was introduced in v2.5. However, you're correct that the issue is no longer present.

wpsmort:
Reopening this as it is not fixed and is still present.

contactashish13:
@wpsmort PRs: https://github.com/semperfiwebdesign/aioseop-pro/pull/459 and https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2101

PR #2101, opened 2018-12-07
Bug with video sitemap when added to XML Sitemap indexes

contactashish13:
#1650

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -826,7 +826,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 		 *
 		 * @return string
 		 */
-		private function get_filename() {
+		protected function get_filename() {
 			$filename = 'sitemap';
 			if ( ! empty( $this->options[ "{$this->prefix}filename" ] ) ) {
 				$filename = $this->options[ "{$this->prefix}filename" ];


wpsmort:
@contactashish13 I tested this by following the steps to reproduce (https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1650#issue-313108957) and it did not resolve the problem.  The Video XML Sitemap still appeared in the XML Sitemap index as http://domain.com/.xml.gz.  I tested this on an existing site and got the same result.

In addition, I got this PHP Notice:

PHP Notice:  Undefined index: aiosp_video_sitemap_options in all-in-one-seo-pack-pro/pro/video_sitemap.php on line 201

contactashish13:
@wpsmort I am unable to reproduce this problem on a freshly installed WP 5.0 site. My PR appears to fix the problem. Can you share details of a site where I can test this?

wpsmort:
I tested this again on a new site and it worked.  I then tested it on an existing site and it also worked fine.  I wasn't able to reproduce the PHP notice in my comment here https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2101#issuecomment-449430596.  I believe this was due to a mistake I'd made when merging the code from the Pro PR here https://github.com/semperfiwebdesign/aioseop-pro/pull/459.

Issue #2104, opened 2018-12-09
Remove exabot from Bad Bot Blocker


michaeltorbert:
Reported here: https://wordpress.org/support/topic/alexa-com-is-blocked-by-bad-bot-blocker/

We block the bot, exabot, in our Bad Bot Blocker module. Unfortunately this also blocks Alexabot from alexa.com.
Code reference: https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/8a245b26873757ca5ed52c24323e99fbb6564370/admin/aioseop_module_class.php#L542

Issue #1769, opened 2018-07-11
Bad Bot Blocker module blocks SEMRush's site audit


michaeltorbert:
Reported here: https://twitter.com/Paul47Insights/status/1016816541264826368

To fix this, we'll probably need to run a site audit and log exactly what bot name/string is getting blocked.

We may want to see if we can get a test account for semrush.

arnaudbroes:
I was able to reproduce this. Their bot is part of our blocklist.
![image](https://user-images.githubusercontent.com/18156697/46638491-73ced080-cb61-11e8-8d26-26556d6c3ae3.png)

PR #2107, opened 2018-12-09
Remove Exabot from bad bots blacklist

arnaudbroes:
#2104

### DIFF
--- a/inc/aioseop_updates_class.php
+++ b/inc/aioseop_updates_class.php
@@ -122,7 +122,7 @@ class AIOSEOP_Updates {
 		}
 		
 		if (
-			( ! AIOSEOPPRO && version_compare( $old_version, '2.12', '<' ) ) ||
+			( ! AIOSEOPPRO && version_compare( $old_version, '3.0', '<' ) ) ||
 			( AIOSEOPPRO && version_compare( $old_version, '3.0', '<' ) )
 		) {
 			$this->bad_bots_remove_exabot_201902();
--- a/inc/aioseop_updates_class.php
+++ b/inc/aioseop_updates_class.php
@@ -122,8 +122,7 @@ class AIOSEOP_Updates {
 		}
 		
 		if (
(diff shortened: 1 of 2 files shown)


michaeltorbert:
@arnaudbroes 

We need to remove it from the database as well. See https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1769 for an example.

arnaudbroes:
@wpsmort please go ahead and test this using these steps:

1. Install aioseop from the WordPress.org plugin directory
2. Enable the Bad Bot Blocker and click the "Save Options" button
3. Enable "Use Custom Blocklists" and confirm that "Exabot" appears in the list
4. Delete aioseop and install the dev version from this PR
5. Confirm that "Exabot" no longer appears in the list

wpsmort:
@arnaudbroes I followed the steps to test and confirmed that this is not working.  Exabot is still in the list and in the database.

arnaudbroes:
@wpsmort please try with this version - https://we.tl/t-Y9eaIWmewC. It's exactly the same as the PR version, with the exception that it runs `$this->bad_bots_remove_exabot_201902();` in the constructor of the update class. Only an update of the plugin will trigger this function. I'm not sure how we could reproduce this on a test site otherwise.

arnaudbroes:
Steve and I tested this both again and confirm this works.

inc/aioseop_updates_class.php
```diff
@@ -120,7 +120,13 @@ function do_version_updates( $old_version ) {
 		) {
 			$this->bad_bots_remove_semrush_201810();
 		}
-
+		
+		if (
+			( ! AIOSEOPPRO && version_compare( $old_version, '3.0', '<' ) ) ||
```
michaeltorbert: Since we're merging the version numbers now we don't need to distinguish between free and pro when it comes to version numbers.

wpsmort:
This is tested and is good to go.

PR #2132, opened 2019-01-09
Use Placeholder to make a string translatable.

mayukojpn:
## Proposed changes

The string ` to upgrade to Pro Version and get:` hasn't finished sentence and couldn't be translated to Japanese which has different word ordering.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] Unit tests pass with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

I checked on my staging WP site and added a comment for translators.

## Further comments

Now I'm in a process to fix wrong translations and translation guideline errors. :) I hope more Japanese users can use your product.
https://translate.wordpress.org/locale/ja/default/wp-plugins/all-in-one-seo-pack

### DIFF
--- a/admin/display/general-metaboxes.php
+++ b/admin/display/general-metaboxes.php
@@ -42,8 +42,11 @@ class aiosp_metaboxes {
 						<p>
 							<strong>
 								<?php
-								echo aiosp_common::get_upgrade_hyperlink( 'side', __( 'Pro Version', 'all-in-one-seo-pack' ), __( 'CLICK HERE', 'all-in-one-seo-pack' ), '_blank' );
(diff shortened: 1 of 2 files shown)


arnaudbroes:
@mayukojpn we've closed this since these three commits are not related to each other. Can you please create a PR for each commit.

michaeltorbert:
@mayukojpn  Please create three separate Github issues. Then each issue should have its commit(s). Then each issue should have its own separate PR, explaining what was improved.

Issue #1929, opened 2018-09-18
Default should be sitemap indexes


michaeltorbert:
We have a notice "Notice: To avoid problems with your XML Sitemap, we strongly recommend you enable Sitemap Indexes and set the Maximum Posts per Sitemap Page to 1000.
Click here to make these recommended changes"

We should check when this is displaying. On a brand new site it showed up for me. Shouldn't we just default to indexes? Should this even be an option?

wpsmort:
Ashish and I have discussed auto-enabling sitemap indexes and pagination on sites with over x number of URLs.  This was discussed because of the problems reported in #2008 and #2014 where users with large sites reported 500 and 503 errors with their large sites when updating to v2.9.

arnaudbroes:
I'm all for default sitemap indexing. I don't see a reason not to.

michaeltorbert:
Why would we only auto-enable for >x URLs? That seems tricky and buggy, shouldn't we just auto enable it for all new sites and set it to 1000?

michaeltorbert:
We simply need to change the default option from off to on in the settings code.

Issue #541, opened 2016-10-05
XML Sitemap - Remove static sitemap option


wpsmort:

michaeltorbert:
Though we were the _first_ WordPress plugin to have dynamic sitemaps, we're now the **ONLY** plugin still being updated to have static sitemaps. 

The static sitemap functionality is outdated and has a much higher technical burden, especially when dealing with the filesystem. At some point we should remove it entirely. 

We may want to look into caching via transients or something.

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #844](https://github.com/awesomemotive/aioseop-pro/issues/844) via [**ZenHub**](https://www.zenhub.com/)

PR #2136, opened 2019-01-10
Sitemap Optimization (revised)

EkoJR:
Issue #2008

## URL to Post IDs Filter
Delete & Disable Transient Data; for testing peak performance.
```
function aioseop_remove_set_transient_url_postids() {
	remove_action( 'shutdown', 'aiosp_common::set_transient_url_postids' );
	if ( is_multisite() ) {
		delete_site_transient( 'multisite_aioseop_attachment_url_postids' );
	} else {
		delete_transient( 'aioseop_attachment_url_postids' );
	}
}
add_action( 'shutdown', 'aioseop_remove_set_transient_url_postids', 9 );
```

## Image IDs => URLs cache filter
```
function aioseop_remove_set_transient_image_ids_urls() {
	remove_action( 'shutdown', array( 'All_in_One_SEO_Pack_Sitemap', 'set_transient_image_ids_urls' ) );
	if ( is_multisite() ) {
		delete_site_transient( 'aioseop_multisite_image_ids_urls' );
	} else {
		delete_transient( 'aioseop_image_ids_urls' );
	}
}
add_action( 'shutdown', 'aioseop_remove_set_transient_image_ids_urls', 9 );
```

### DIFF
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -400,7 +400,8 @@ class aiosp_common {
 
 				// Set the URL => PostIDs.
 				$uploads_dir = wp_upload_dir();
-				$custom_img_base_url = $uploads_dir['baseurl'] . '/' . str_replace( basename( $meta_value['file'] ), '', $meta_value['file'] );
+				$uploads_dir2 = wp_get_upload_dir();
+				$custom_img_base_url = $uploads_dir['baseurl'] . '/' . str_replace( wp_basename( $meta_value['file'] ), '', $meta_value['file'] );
 				foreach ( $meta_value['sizes'] as $image_size_arr ) {
 					$tmp_arr[ md5( ( $custom_img_base_url . $image_size_arr['file'] ) ) ] = $results_2[ $i ]['post_id'];
 				}
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -230,7 +230,7 @@ class aiosp_common {
 		// If static variable is still empty, load transient data.
 		if ( is_null( self::$attachment_url_postids ) ) {
 			if ( is_multisite() ) {
-				self::$attachment_url_postids = get_site_transient( 'multisite_aioseop_attachment_url_postids' );
+				self::$attachment_url_postids = get_site_transient( 'aioseop_multisite_attachment_url_postids' );
 			} else {
 				self::$attachment_url_postids = get_transient( 'aioseop_attachment_url_postids' );
 			}
@@ -298,7 +298,7 @@ class aiosp_common {
 	 */
 	public static function set_transient_url_postids() {
 		if ( is_multisite() ) {
-			set_site_transient( 'multisite_aioseop_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
+			set_site_transient( 'aioseop_multisite_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
 		} else {
 			set_transient( 'aioseop_attachment_url_postids', self::$attachment_url_postids, 24 * HOUR_IN_SECONDS );
 		}
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -419,6 +419,7 @@ class aiosp_common {
 	 * Error Hand Images
 	 *
 	 * Unused/Conceptual function potentually used in `aiosp_common::attachment_url_to_post_id_query_2()`.
+	 * This is to handle errors where a normal try/catch wouldn't have the exception needed to catch.
 	 *
 	 * @see aiosp_common::attachment_url_to_post_id_query_2()
 	 *
@@ -430,7 +431,7 @@ class aiosp_common {
 	 * @throws ErrorException
 	 */
 	public static function error_handle_images( $errno, $errstr, $errfile, $errline ) {
-		// Possibly handle know issues differently.
+		// Possibly handle known issues differently.
 		// Handles unserialize() warning notice.
 		if ( 8 === $errno || strpos( $errstr , 'unserialize():' ) ) {
 			throw new ErrorException( $errstr, $errno, 0, $errfile, $errline );
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -282,7 +282,7 @@ class aiosp_common {
 			 *
 			 * @see aiosp_common::set_transient_url_postids()
 			 */
-			add_action( 'shutdown', 'aiosp_common::set_transient_url_postids' );
+			add_action( 'shutdown', array( 'aiosp_common', 'set_transient_url_postids' ) );
 		}
 
 		return $id;
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -15,6 +15,13 @@ class aiosp_common {
 // @codingStandardsIgnoreEnd
 
 	/**
+	 * @var null|array
+	 *
+	 * @since 2.9.2
+	 */
+	public static $attachment_url_postids = null;
+
+	/**
 	 * aiosp_common constructor.
 	 *
 	 */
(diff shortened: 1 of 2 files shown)


EkoJR:
When done testing, have @contactashish13 review the code as well.

@contactashish13 this doesn't have the Query Step concept, but were a few good changes included. Main one, using the ID (instead of the URL) as much as possible. The URL to ID conversion was the main culprit, and even ID to URL can be heavy but not much can be done there. 

So, when returning image data, it will try Image IDs first, and then the remaining Image URLs (if image_url_is_valid).

I do think get_images_from_post() could be moved to aiosp_common as a static method, but would very much prefer keeping it as a separate commit/PR. So that way it can reflect the changes better (just in case for future reference) instead of appearing as new code.

The aiosp_common class has the same changes as the other with attachment_url_to_postid()

contactashish13:
@wpsmort as discussed on the call last week, why are we not

1. removing static sitemap generation immediately if that is one of the known symptoms?
2. thinking of auto enabling indexing and pagination if we know we can determine how many URLs will be in the sitemap? 

I've talked about both these points almost 2 months ago but I don't see them as action points in any issue/PR dealing with the problems caused in 2.9 when we know these will mitigate the situation. Working on complex concepts such as the ones @EkoJR is working on is important and not something that can be implemented in the short term considering how sensitive these changes are but I'm not sure why we are not addressing low hanging fruits that will not only ease testing but also reduce code complexity.

wpsmort:
@contactashish13 We have an issue open for removing static sitemaps here #541.  Removing this doesn't address the problem with how we're handling images on sites with thousands of images.

As for detecting sites with x number of URLs and automatically enabling Sitemap Indexes and pagination, I'm all for this if we are in agreement it will work.  Issue #1929 covers this.

contactashish13:
@wpsmort removing static sitemaps does not address the problem but it mitigates it on the sites that use that option. At least that is what I believe we agreed was the case while debugging the cause of errors. If any kind of mitigation helps, shouldn't that be fixed without further delay? I believe https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/541 has already been delayed for 2 years.

wpsmort:
@contactashish13 Feel free to submit a PR for #541 but it **must not** get in the way of resolving the performance issues introduced in v2.9 (#2008 and #2014).  We have users who cannot update because their sites become unresponsive when updating to v2.9 or later.  Their sites work fine on v2.8 with whatever sitemap settings they have (static or dynamic / indexes or not) but break in v2.9 and later.  That is our highest priority and must be fixed in this release.

inc/aiosp_common.php
```diff
@@ -182,4 +189,255 @@ public static function make_xml_safe( $tag, $value ) {
...
+			}
+
+			self::$attachment_url_postids[ $url_md5 ] = $id;
+
+			/**
+			 * Sets the transient data at the last hook instead at every call.
+			 *
+			 * @see aiosp_common::set_transient_url_postids()
+			 */
+			add_action( 'shutdown', 'aiosp_common::set_transient_url_postids' );
```
contactashish13: @EkoJR I suggest using `add_action( 'shutdown', array( 'aiosp_common', 'set_transient_url_postids' ) )` instead

contactashish13:
@EkoJR can you please add more inline comments to the code so that it becomes easier to understand what you intended to do?

inc/aiosp_common.php
EkoJR: I could of swore that had issues with static functions, but I guess I forgot to change it back at some point.

contactashish13:
@wpsmort can you please test this and confirm that it reduces memory? I can then continue the code review.

wpsmort:
I have tested this thoroughly and can confirm that it works and is an improvement over the current release. 

When compared with v2.10.1, it does not reduce the memory used to generate the sitemap_post.xml when sitemap indexes are enabled. On a site with lots of images, the memory used is almost double what is used in v2.10.1.  However, on a site with lots of URLs (posts), the memory is the same as what is used in v2.10.1.

However, this PR does significantly reduce the time taken to generate the sitemap resulting in less 503 timeouts.

When compared with the competition, it performs better on sites with lots of images resulting in a usable sitemap when the competition times out.  

This works best when sitemap indexes is enabled and pagination is set to 1000 URLs per sitemap.

This works fine with Jetpack.

modules/aioseop_sitemap.php
```diff
@@ -3397,72 +3430,119 @@ private function get_images_from_post( $post ) {
...
+			$post_image_ids = array_merge( $post_image_ids, $this->get_gallery_image_ids( $post ) );
 
-			$this->get_gallery_images( $post, $images );
+			$this->get_gallery_images( $post, $post_image_urls );
 
+			// Get image URLs from content.
+			$content  = $post->post_content;
 			$content .= $this->get_content_from_galleries( $content );
-			$this->parse_content_for_images( $content, $images );
+			$this->parse_content_for_images( $content, $post_image_urls );
```
contactashish13: @EkoJR can we move this and related methods to common?

inc/aiosp_common.php
```diff
@@ -182,4 +189,256 @@ public static function make_xml_safe( $tag, $value ) {
...
+			}
+		}
+
+		// Search for URL and get ID.
+		if ( isset( self::$attachment_url_postids[ $url_md5 ] ) ) {
+			// If static is already loaded and has URL, then return the URL's Post ID.
+			$id = intval( self::$attachment_url_postids[ $url_md5 ] );
+		} else {
+			// Check to make sure Image URL is not outside the website.
+			$uploads_dir = wp_upload_dir();
```
contactashish13: @EkoJR better to use wp_get_upload_dir as its lightweight https://developer.wordpress.org/reference/functions/wp_get_upload_dir/

inc/aiosp_common.php
```diff
@@ -182,4 +189,256 @@ public static function make_xml_safe( $tag, $value ) {
...
+
+		$tmp_arr = array();
+		$results_2 = $wpdb->get_results(
+			"SELECT post_id, meta_value FROM {$wpdb->postmeta} WHERE `meta_key` = '_wp_attachment_metadata' AND `meta_value` != '" . serialize( array() ) . "';",
+			ARRAY_A
+		);
+		if ( $results_2 ) {
+			for ( $i = 0; $i < count( $results_2 ); $i++ ) {
+				// TODO Investigate potentual memory leak(s); currently with unserialize.
+				$meta_value = unserialize( $results_2[ $i ]['meta_value'] );
```
contactashish13: maybe https://developer.wordpress.org/reference/functions/maybe_unserialize/ is better?

inc/aiosp_common.php
```diff
@@ -182,4 +189,256 @@ public static function make_xml_safe( $tag, $value ) {
...
+	 * @param $errfile
+	 * @param $errline
+	 * @return bool
+	 * @throws ErrorException
+	 */
+	public static function error_handle_images( $errno, $errstr, $errfile, $errline ) {
+		// Possibly handle known issues differently.
+		// Handles unserialize() warning notice.
+		if ( 8 === $errno || strpos( $errstr , 'unserialize():' ) ) {
+			throw new ErrorException( $errstr, $errno, 0, $errfile, $errline );
```
contactashish13: ErrorException might not work in PHP versions below 5.1.0

modules/aioseop_sitemap.php
EkoJR: I thought I had mentioned this before in one of the previous PRs. For this PR, I would have to disagree. I had already added changes to that method, and would be better in a separate PR to reflect the changes accurately.

In other words.
* This PR will reflect changes to the function(s).
* Next PR would just be a refactor to reflect that is was just moved with no modifications; which makes sifting through the changes easier for future reference.

Conceptually, I will agree with you. Just not on this PR; since the commits are likely to be squashed.

inc/aiosp_common.php
EkoJR: It should be expected as serialized values, but wouldn't hurt to add/change.

However, this is in a disabled method for conceptual use.

inc/aiosp_common.php
EkoJR: I believe that is true. However, this was mainly used for dev purposes.

This is also in a disabled method for conceptual use.

inc/aiosp_common.php
contactashish13: I thought the minimum version was 4.5 but I'm not sure if it makes sense to put this code with a version check.

inc/aiosp_common.php
contactashish13: maybe just change ErrorException into Exception if you are going to retain the method for future use?

inc/aiosp_common.php
```diff
@@ -182,4 +189,256 @@ public static function make_xml_safe( $tag, $value ) {
...
+				*/
+
+				// Images and Videos use different variable structures.
+				if ( false === $meta_value || ! isset( $meta_value['file'] ) && ! isset( $meta_value['sizes'] ) ) {
+					continue;
+				}
+
+				// Set the URL => PostIDs.
+				$uploads_dir = wp_upload_dir();
+				$custom_img_base_url = $uploads_dir['baseurl'] . '/' . str_replace( basename( $meta_value['file'] ), '', $meta_value['file'] );
```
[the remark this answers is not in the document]
EkoJR: Oh gotcha. Didn't see the function at the end. I thought you meant the variable itself.

inc/aiosp_common.php
EkoJR: My thoughts exactly. So, either would need to...

* Increase the WP version requirement from 4.4 to 4.5.
* Keep as wp_upload_dir.

@michaeltorbert Should we increase the required version?

inc/aiosp_common.php
EkoJR: I believe that type of function was required for that type of warning. Because I remember trying Exception and the warning wasn't being caught.

Right now it's all disabled, and would rather leave it as what I know was last functional.

Issue #2142, opened 2019-01-13
apply_post_title_format() sometimes called on page/custom post type


arnaudbroes:
Reported by @harryrobbins #2137. 

> "Sometimes (for reasons which are not clear to me) apply_post_title_format() is called on a page or custom post type. This currently causes the title in the of the page to be printed as the format (e.g. <title>%page_title% | %blog_title%</title>) instead of the parsed template.

arnaudbroes:
Here's some more info from Harry.

> "I've done a bit more digging and I think I've found the issue.
> 
> Your plugin uses the "is_page()" to determine which title function to use.
> e.g on e.g. aioseop.class.php line 1656
> 
> The "is_page" function relies on the WP_Query::is_page property which is
> sometimes set to false even if the underlying post type is a page (due to
> various things such as the "show_on_front" variable).
> 
> This results in a page post-type making it all the way through the various
> If statements in get_aioseop_title and getting to Line 1732 where "
> apply_post_title_format" is called.
> 
> I believe the issue is that is_page() is intended for determining which
> template to use, rather than which post type to query."

arnaudbroes:
@EkoJR can you have a look and see if you're able to reproduce and/or whether Harry's information makes sense to you from a code perspective.

EkoJR:
There's 2 functions that achieve similar results. 

`apply_page_title_format()` & `apply_post_title_format()`

It seems he's trying to replicate the page function on a post. Which would be resolved by using %post_title% instead, but seems to alternate between the two. I don't think I would be able to reproduce this without more info about the website; since I suspect another plugin played a part (either part of its modifications, or post was converted to a page).

I believe it just needs to check if the post is in a hierarchical post type, and then use 1 of the 2 functions listed above.

I also think it would help if the plugin would drop %page_title% in favor of just %post_title%; as well as reduce any other %page_&ast;% in favor of %post_&ast;% shortcodes. Both likely operate the same already, and it would reduce code, bugs like this, and confusion.

arnaudbroes:
@harryrobbins are you able to provide us with steps to reproduce this problem, or alternatively, give us access to your website so we can take a look?

michaeltorbert:
Closing this as nobody has been able to reproduce it.

PR #2154, opened 2019-01-18
Fix apply_post_title_format() called on pages

EkoJR:
Issue #2142

Alternate PR #2137

Unable to reproduce issue. @harryrobbins might be able to confirm if this fixes it though.

> I don't think I would be able to reproduce this without more info about the website; since I suspect another plugin played a part (either part of its modifications, or post was converted to a page).

https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2142#issuecomment-455443891


michaeltorbert:
Closing this as nobody has been able to reproduce it.

Issue #2158, opened 2019-01-18
Undefined index: aiosp_opengraph_options in aioseop_opengraph.php line 1101


wpsmort:
I got this PHP Notice whilst testing the 2.11 branch on a new site:

PHP Notice:  Undefined index: all-in-one-seo-pack/modules/aioseop_opengraph.php on line 1101

It is related to this commit for PR #1873 - https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1873/commits/527c4cf730d4047103b2cfccc9eec3d1478592cf

EkoJR:
It looks like the options array in the backend had some restructuring done on it, but doesn't seem to be getting set/initialized.

@contactashish13 might have a better idea. I'm unsure where to apply changes.

contactashish13:
@wpsmort I can see the potential problem but I cannot reproduce this. Can you outline the exact steps?

wpsmort:
@contactashish13 Here are the steps to reproduce this:

1. Download the v2.11 branch of free and the v2.12 branch of Pro and create the Pro plugin from these branches.
2. Install the plugin on a brand new site.
3. Enable wp_debug in the wp-config.php file
4. Activate the Social Meta module.
5. View the Uncategorized category and you should see the PHP notice.

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2159

PR #2159, opened 2019-01-22
Undefined index: aiosp_opengraph_options in aioseop_opengraph.php line 1101

contactashish13:
#2158


wpsmort:
This is tested and is good.

Issue #1003, opened 2017-07-31
use Grunt


michaeltorbert:
from private PM

"for phplinting and php coding standards...
it uses NodeJS and runs locally. It does quite a few things - running PHPCS, minimizing CSS/JS, optimizing images, running text-domain validations, preparing the POT file and even ensures phpunit test cases are run in case they are configured. PHPStorm makes the integrator responsible for a lot of these things whereas by using grunt, this responsibility can be distributed to the contributors :smile: I find it useful because I don't have to worry about checking in improper code. I mentioned it because I see lots of code that does not conform to coding standards. I can set it up if you want."

amostajo:
My little grain of salt... just so we can choice the right JS tasker, between `grunt` and `gulp`.
https://medium.com/@preslavrachev/gulp-vs-grunt-why-one-why-the-other-f5d3b398edc4
Let's discuss further on slack.

contactashish13:
@michaeltorbert here are all the tasks that need to be performed:

1. we will need to create a new repo which will contain all the grunt plugins that we need to use. This will ensure that both the free and pro plugins can share the same set of tasks. I will create this repo on my account and you can clone it.
2. locally install node, npm and grunt
3. when I run grunt locally, it will change all the files according to the grunt tasks. 

Do you want me to push changes only for task 1 above or also for 3? If for both, do you want me to create one single PR for both or would you prefer 2 different PRs?

michaeltorbert:
@contactashish13 I think it's better if the grunt plugins are in the WP plugin repo.

contactashish13:
@michaeltorbert i would suggest they be in a separate repository because they are going to be shared by the free and pro plugins. Also grunt tasks are not part of the plugin's functionality and this will unnecessarily complicate the grunt scripts. 

Could you also respond to my question about the PRs?

michaeltorbert:
@contactashish13 How would it complicate the grunt scripts? That should change them any.

-Let's hold off on any PRs until we've finished deciding exactly what to do.

contactashish13:
@michaeltorbert is there a particular reason you want to keep grunt as part of the source code of the plugins despite it not adding any functionality to the plugin per se?

michaeltorbert:
What happens when something only applies to one? Or when we stop having Pro be effectively a fork of free?
It would only be in the development repository. Certainly we'd add it to .gitattributes to keep it out of releases.

contactashish13:
[...]

PR #2171, opened 2019-01-30
grunted changes

contactashish13:
#1003


modules/aioseop_importer_exporter.php
```diff
@@ -146,18 +146,18 @@ function debug_post_types() {
 					$this->default_options['import_export_help']['default'] .= "\n</ul>\n";
 				} else {
 					$this->default_options['import_export_help']['default'] .= '<br />'
-																			   . __(
-																				   'There are no other modules currently loaded!',
-																				   'all-in-one-seo-pack'
-																			   );
+																			. __(
```
EkoJR: White space needs to be removed, and could probably just remove the new lines.

modules/aioseop_feature_manager.php
```diff
@@ -27,16 +27,19 @@ function __construct( $mod ) {
 			parent::__construct();
 			$this->module_info = array(
 				'sitemap'           => array(
-					/* translators: the XML Sitemaps module allows users to generate a sitemap in .xml format for their website and
+					/*
```
EkoJR: This looks incorrect if done this way for wrapping a comment on a new line. Could probably use...

```
/*
 * translators: comment.
 */
```

js/count-chars.js
```diff
@@ -12,24 +12,32 @@
...
  */
-jQuery(document).ready( function() {
-	jQuery("#aiosp_title_wrapper").bind("input", function() {
-		jQuery("#aiosp_snippet_title").text(jQuery("#aiosp_title_wrapper input").val().replace(/<(?:.|\n)*?>/gm, ""));
-	});
-	jQuery("#aiosp_description_wrapper").bind("input", function() {
-		jQuery("#aioseop_snippet_description").text(jQuery("#aiosp_description_wrapper textarea").val().replace(/<(?:.|\n)*?>/gm, ""));
-	});
-});
+jQuery( document ).ready(
```
EkoJR: This also doesn't need to be broken down. Everyone generally expects `jQuery( document ).ready( function() {`

I also don't think it is ideal to indent so much; which is a result from the change in coding style. Just ends up looking like a lot of nested code. 

...and it looks like this occurs for every line with jQuery(element).do_something( function() {

js/count-chars.js
```diff
@@ -12,24 +12,32 @@
 
 var aiosp_title_extra = parseInt( aioseop_count_chars.aiosp_title_extra );
 
-jQuery( document ).ready( function() {
-	aioseopInitCounting();
-});
+jQuery( document ).ready(
```
EkoJR: This doesn't need to be broken down. Everyone generally expects `jQuery( document ).ready( function() {`

aioseop_class.php
```diff
@@ -126,7 +126,7 @@ function __construct() {
 				__( '%blog_description% - Your blog description', 'all-in-one-seo-pack' ) . '</li><li>' .
 				__( '%post_title% - The original title of the post', 'all-in-one-seo-pack' ) . '</li><li>' .
 				__( '%category_title% - The (main) category of the post', 'all-in-one-seo-pack' ) . '</li><li>' .
-				__( '%category% - Alias for %category_title%', 'all-in-one-seo-pack' ) . '</li><li>' .
+				__( '%1$category% - Alias for %2$category_title%', 'all-in-one-seo-pack' ) . '</li><li>' .
```
EkoJR: This appears to be a Change. Did you add this by accident?

admin/display/dashboard_widget.php
```diff
@@ -23,10 +23,12 @@ function __construct() {
 		 */
 		function add_dashboard_widget() {
 			if ( current_user_can( 'install_plugins' ) && false !== $this->show_widget() ) {
-				wp_add_dashboard_widget( 'semperplugins-rss-feed', __( 'SEO News', 'all-in-one-seo-pack' ), array(
-					$this,
-					'display_rss_dashboard_widget',
-				) );
+				wp_add_dashboard_widget(
```
EkoJR: It appears params like this were missed.

```
wp_add_dashboard_widget(
	'semperplugins-rss-feed',
	__( 'SEO News', 'all-in-one-seo-pack' ),
	array(
		$this,
		'display_rss_dashboard_widget',
	)
);
```

modules/aioseop_importer_exporter.php
contactashish13: thats what it did.

modules/aioseop_feature_manager.php
contactashish13: grunt tried its best to make sense of a wrongly formatted comment but things such as this that are totally wrong may need manual intervention. not a breaking change though as it is just a comment

admin/display/dashboard_widget.php
contactashish13: i'm not sure what you are referring to here

aioseop_class.php
contactashish13: this is done by grunt and it follows the standards of http://php.net/manual/en/function.sprintf.php - look at "Example #3 Argument swapping". This is in fact the correct way of using any sprintf-based function.

js/count-chars.js
contactashish13: I'm not sure about what everyone expects but I believe this is correct as it is an anonymous function defined within a function and that needs indentation. It is not a breaking change and if it improves readability, why not? 

Also, the only way to exclude such things that deviate from the definition is to add comments on the lines that need to be excluded which I think will add more clutter. I don't think this change is a big deal.

admin/display/dashboard_widget.php
EkoJR: ...it was just the proper way to WPCS, but nevermind.

aioseop_class.php
EkoJR: Where is the `sprintf()`?!

js/count-chars.js
EkoJR: It does not improve readability, if anything it throws it off. Everyone expects `jQuery( document ).ready( function() {`, and it's just redefine a coding style that no one uses...but us...

modules/aioseop_importer_exporter.php
EkoJR: Whoa...that is one hell of an indent. I didn't even see the text til now.

modules/aioseop_feature_manager.php
EkoJR: Ah ok, so this was grunt and not you?

Normally it would be a 1 line comment, but this was a bit different with the length. I checked to see what WPCS considered valid, and it seems to asks for...

```
/*
 * translators: comment.
 */
```

With an empty line preceding it. I tried this change, and the old version, both styles get flagged by WPCS. Should it be corrected now, or on the next WPCS?

Issue #2174, opened 2019-01-30
No support for < WP 4.6


enaimi:
Immediately after upgrading from 2.10.1 to 2.11 server returns error 500  

PHP Fatal error:  Call to undefined function wp_get_attachment_caption() in /var/www/[domain]/wp-content/plugins/all-in-one-seo-pack/modules/aioseop_sitemap.php on line 3502, referer: [domain]/wp-admin/index.php?page=aioseop-about

Notice that - 
I have 2 websites (2 separate WP installs ) under the same domain. one of them is in English which works fine with the new version, and the other one is in Hebrew and the update leads to the error mentioned above

wpsmort:
@enaimi Can you send us an Administrator login for your site and FTP login for your server using the secure form here please - https://semperplugins.com/contact/support-login-information/

enaimi:
Since I'm part of an organization that has a restricted network policy, access from outside our network is highly complicated to arrange... is there anything else I can do instead?

wpsmort:
@enaimi This may be something that can be resolved by enabling Sitemap Indexes and setting the Maximum Posts Per Sitemap Page to something such as 1000 (https://semperplugins.com/documentation/xml-sitemaps-module/#enable-sitemap-indexes).  Can you try that and see if it resolves the 500 error please.

wpsmort:
@enaimi What version of WordPress do you have?

enaimi:
I believe it is already enabled. please see attached screenshots
although the file is empty http://1062fm.co.il/he/sitemap.xml

![2019-01-30_190118](https://user-images.githubusercontent.com/47187008/51998526-86c73100-24c1-11e9-9478-998ec1f6e659.jpg)
![2019-01-30_190012](https://user-images.githubusercontent.com/47187008/51998527-86c73100-24c1-11e9-922c-70f9e8361698.jpg)


WP version is 4.4.4 but like I said... the English version of this website is works with the update. 

I''ve noticed that the change log says:
XML Sitemap – improved special characters encoding
Fix for Japanese language

So... although I really know nothing about this... my guess would be that this might have something to do with Hebrew characters or RTL lang... something like that...?

Thanks

EkoJR:
This was introduced before 2.11 (not sure exactly when yet), and wasn't introduced in WP until 4.6.

There could be some form of backwards compatibility, but looking at the function it's just a post excerpt being wrapped.
[...]

PR #2175, opened 2019-01-30
Add reverse compatibility 4.6 with wp_get_attachment_caption()

EkoJR:
Issue #2174

Call to undefined function wp_get_attachment_caption() in prior versions of WP v4.6


michaeltorbert:
Closing since we're dropping support for <4.6.

Issue #2179, opened 2019-02-01
Use Schema.org Markup should remove all schema markup


wpsmort:
Requested here - https://wordpress.org/support/topic/schema-org-markup-cannot-be-turned-off/

The option Use Schema.org Markup (https://semperplugins.com/documentation/general-settings/#use-schema-markup) should remove ALL of the schema markup that we output.  Schema markup is output from these settings:

- Display Sitelinks Search Box (https://semperplugins.com/documentation/google-settings/#display-sitelinks-search-box)
- Set Preferred Site Name (https://semperplugins.com/documentation/google-settings/#set-preferred-site-name)
- Social Profile Links / Person or Organization / Associated Name (https://semperplugins.com/documentation/social-meta-module/#social-profile-links)

Here's example code that is output by all of these options:

```
<script type="application/ld+json">
{
  "@context": "http://schema.org",
  "@type": "WebSite",		  "name": "Steve&#039;s Test Site",        "potentialAction": {
  "@type": "SearchAction",
  "target": "https://aioseop.local/?s={search_term}",
  "query-input": "required name=search_term"
  },		  "url": "https://aioseop.local/"
}
</script>

<script type="application/ld+json">
{ 
  "@context" : "http://schema.org",
  "@type" : "Organization",
  "name" : "Steve&#039;s Test Site",
  "url" : "https://aioseop.local",
  "sameAs" :["https://www.facebook.com/smortiboy","https://twitter.com/wpsmort","https://plus.google.com/102594674486171076537"]
}
</script>
```

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2181

PR #2181, opened 2019-02-05
Use Schema.org Markup should remove all schema markup

contactashish13:
#2179


wpsmort:
This is tested and is good.

Issue #2157, opened 2019-01-18
E-mail debug information form missing


wpsmort:
The E-mail debug information form is missing from the System Status box in the Performance module.  This happened in v2.10.

<img width="629" alt="screen shot 2019-01-18 at 11 21 58 am" src="https://user-images.githubusercontent.com/17548525/51399163-4952c300-1b13-11e9-837e-641812118e3c.png">

wpsmort:
@arnaudbroes Can you find out which commit resulted in the form not being displayed anymore.

arnaudbroes:
Traced it back to https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/77bbe410dc75765aee0e53947dc964046c05b6e6.

PR #2182, opened 2019-02-05
Add 'input' to allowed tags with 'html' escaping

EkoJR:
Issue #2157

Fix missing Email Debug Information inputs on Performance settings page.

### DIFF
--- a/admin/aioseop_module_class.php
+++ b/admin/aioseop_module_class.php
@@ -2252,7 +2252,27 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Module' ) ) {
 		 * @since ?
 		 * @since 2.12 Add 'input' to allowed tags with 'html'. #2157
 		 *
-		 * @param $args
+		 * @param array $args {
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good.

admin/aioseop_module_class.php
```diff
@@ -2245,10 +2245,14 @@ function do_multi_input( $args ) {
...
 		/**
+		 * Get Option HTML
+		 *
 		 * Outputs a setting item for settings pages and metaboxes.
 		 *
-		 * @param $args
+		 * @since ?
+		 * @since 2.12 Add 'input' to allowed tags with 'html'. #2157
 		 *
+		 * @param $args
```
contactashish13: add the description of the parameter as per WPCS

EkoJR: It also needs the Type added, as well as @type for the description. Which I'd prefer breakpointing multiple instances since it seemed the values were dynamic, and so I held off on it, but yea definitely!

Issue #2177, opened 2019-01-31
Problem in XML sitemap caused by rsquo; character


media377:
Hi,

I have a problem in the All in One SEO plugin generated Sitemap :

Erreur d’analyse XML : entité non définie
Emplacement : http://www.jeremy-taburchi.com/sitemap.xml
Numéro de ligne 37, Colonne 40 :	image:captionDessin de presse de l`&rsquo;`eau sur mars</image:caption>

I tried to update everything, I changed the theme, deactivate (major) plugins, but nothing seems to fix that...

Thanks,

Jerry

media377:
Some characters have disapeared from my previous post sorry (I add blankspace to & r s q u o ; ) : 

Hi,

I have a problem in the All in One SEO plugin generated Sitemap :

Erreur d’analyse XML : entité non définie
Emplacement : http://www.jeremy-taburchi.com/sitemap.xml
Numéro de ligne 37, Colonne 40 :	image:captionDessin de presse de l & r s q u o ; eau sur mars</image:caption>

The "& r s q u o ;’" seems to bug parsing.

I tried to update everything, I changed the theme, deactivate (major) plugins, but nothing seems to fix that...

Thanks,

Jerry

wpsmort:
@media377 You are correct, this is caused by the use of the `&rsquo;` character.  However, I cannot reproduce this if I use the shortcut for that character.  I can reproduce this if I type the `&rsquo;` character out.  

@arnaudbroes Can you see if you can reproduce this on Windows using the Windows shortcut listed here - http://www.rsquo.net/

media377:
Hi, thanks I will test this in the next hours, and come back there as soon after.

arnaudbroes:
@wpsmort I'm not able to reproduce this when I copied the ’ character. The shortcut at rsquo.net doesn't make sense to me.

media377:
Tried to change the character with its escape code but nothing changes.

arnaudbroes:
@media377 do you have a keyboard with a French layout by any chance?

media377:
Yes I have !

Le 2019-02-01 12:44, Arnaud Broes a écrit :

> @media377 [1] do you have a keyboard with a French layout by any chance? 
> 
> --
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].
 

Links:
------
[1] https://github.com/media377
[2]
[...]

PR #2183, opened 2019-02-07
Problem in XML sitemap caused by rsquo; character

contactashish13:
#2177

### DIFF
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -192,15 +192,10 @@ class aiosp_common {
 		} else {
 			// some tags contain sanitized to some extent but they do not encode < and >.
 			if ( ! in_array( $tag, array( 'image:title' ), true ) ) {
-				// use the WP core functions if they exist.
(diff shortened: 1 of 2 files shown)


arnaudbroes:
@contactashish13 FYI, it's best to name your commits after what they actually do or fix instead of the name of the issue because "problem in XML sitemap" doesn't really tell us much.

contactashish13:
In that case I think the name of the issue should change to "problem in sitemap because of rsqo character" because the commit addresses that. If there were multiple commits, subsequent commits do mention what changes from the first one.

arnaudbroes:
@contactashish13 you definitely have a point there. If you think an issue could be described better with a different title and anyone else hasn't changed it already, then please feel free to edit it. :-)

wpsmort:
This is tested and is good.  I was able to reproduce the original bug report and verify that this PR fixes it.

EkoJR:
@contactashish13 If you'd like to improve the PHPUnit test, I found a list of Named Entities. https://gist.github.com/EkoJR/7caa0b4664be2ff1d6f508c8b4b362c9

PR #2187, opened 2019-02-09
Add Grunt

EkoJR:
Issue #1003

~**During merge**, it might be ideal to NOT squash commits into one with this PR. Each commit was organize to each stage of the grunt process, and might be helpful for future reference.~

It occured to me the PR # could be referenced in the commit instead.

Alternate PR #2178


wpsmort:
@EkoJR You've tagged as Needs Code Review and Needs Testing, which is correct here?  Does it need to be assigned to me for testing or are you and Ashish still reviewing the code changes?

EkoJR:
@wpsmort Initially I was thinking about assigning both you and Ashish; since it seemed like testing would be small/quick with this, but last second I thought it would be best to stick to the same routine with the Grunt PR and assign Ashish but forgot to remove the label in the process.

Gruntfile.js
```diff
@@ -90,8 +101,9 @@ module.exports = function(grunt) {
 	grunt.loadNpmTasks( 'grunt-phplint' );
 	grunt.loadNpmTasks( 'grunt-contrib-jshint' );
 	grunt.loadNpmTasks( 'grunt-contrib-uglify' );
+	grunt.loadNpmTasks( 'grunt-eslint' );
 
 	// Default task(s).
-	grunt.registerTask( 'default', ['mkdir', 'phpcbf', 'phpcs', 'phplint', 'jshint', 'uglify'] );
+	grunt.registerTask( 'default', ['mkdir', 'phpcbf', 'phpcs', 'phplint', 'jshint', 'uglify', 'eslint'] );
```
contactashish13: I think eslint should be before uglify so that if eslint fails, uglify does not happen.

contactashish13:
@wpsmort I've marked this as good for testing as whatever changes I have suggested will not change the code.

Gruntfile.js
EkoJR: Got the change added @contactashish13

wpsmort:
@EkoJR I tested this and found some problems with the UI.  On the General Settings screen, many options are hidden by a display:none; even though the relevant setting that controls this is set to display the options.  Here are the instances where this is happening:

**Home Page Settings:**  The fields are not displaying despite Use Static Front Page Instead  being set to Disabled
**Title Settings:**  The fields are not displaying despite Rewrite Titles being set to Enabled
**Content Type Settings:**  The title format fields are not displaying when the checkboxes are checked for custom post types
**Keyword Settings:**  The fields are not displaying despite Use Keywords being set to Enabled

EkoJR:
@wpsmort I found that a strict condition wasn't possible for one of the operations, and it didn't seem like quick fix. So for the time being, I changed it back to a loose condition, and added a comment & `// TODO` of the line ignored.

wpsmort:
@EkoJR I found one last bug.  The character counter for the Home Title and Home Description on the General Settings screen does not increment when you type.  It does show the correct number when you click Update Options.  

The character counter on the Edit Post screen works fine.

EkoJR:
Issue #2199 (Fix enqueue count-char.js when plugin folder name is altered) fixes that count issue

wpsmort:
@EkoJR Can you resolve the merge conflict here please.

EkoJR:
Closing due to merge conflicts

Issue #2189, opened 2019-02-11
Class name output in link rel next and prev with WP5.1


wpsmort:
Whilst testing the current version of AIOSEOP on WordPress v5.1 RC1 (https://wordpress.org/news/2019/02/wordpress-5-1-release-candidate/) I found an issue with the rel next and prev links on paginated content.  When you view the links in the source code they have %20class=post-page-numbers added to the end of the link.  For example:

`<link rel='next' href='http://wp51.local/test-post/2/%20class=post-page-numbers' />`

I tested back through several version of AIOSEOP and got the same result.  I confirmed that this only happens with WP v5.1.

To reproduce this:

1. Create a new site using WP v5.1 RC1 which can be downloaded here https://wordpress.org/news/2019/02/wordpress-5-1-release-candidate/
2. Create a post or page with page breaks
3. Install AIOSEOP and view the source code of any of the paginated pages and you'll see the link looks like this: 

http://domain.com/test-post/2/%20class=post-page-numbers

PR #2191, opened 2019-02-12
Fix href/URL output with <link> elements on paginated posts

EkoJR:
Issue #2189

## Commit Description

WP 5.1 introduced the issue by adding a new CSS class name to prev & next links.

Accessibility: use `aria-current` for the paginated post links output…

https://github.com/WordPress/wordpress-develop/commit/1258d1d9d2c7bef7c18b4e3420219cfa716d0aca

https://core.trac.wordpress.org/ticket/41859

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4480,10 +4480,12 @@ EOF;
 				}
 			}
 			if ( ! empty( $prev ) ) {
-				$prev = $this->substr( $prev, 9, - 2 );
+				$tmp_str = $this->substr( $prev, ( $this->strpos( $prev, 'href="' ) + 6 ) );
+				$prev    = $this->substr( $tmp_str, 0, ( $this->strpos( $tmp_str, '"' ) -  $this->strlen( $tmp_str ) ) );
 			}
 			if ( ! empty( $next ) ) {
-				$next = $this->substr( $next, 9, - 2 );
+				$tmp_str = $this->substr( $next, ( $this->strpos( $next, 'href="' ) + 6 ) );
+				$next    = $this->substr( $tmp_str, 0, ( $this->strpos( $tmp_str, '"' ) -  $this->strlen( $tmp_str ) ) );
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


michaeltorbert:
When testing, make sure this is backwards compatible.

EkoJR:
I hadn't tested for it yet, but it was something I kept in mind.

This PR is currently designed to extract the href attribute. Prior it was just an integer to get the substr.

wpsmort:
This is tested and is good.  I confirmed backwards compatibility is good.

aioseop_class.php
```diff
@@ -4480,10 +4480,12 @@ function get_prev_next_links( $post = null ) {
 				}
 			}
 			if ( ! empty( $prev ) ) {
-				$prev = $this->substr( $prev, 9, - 2 );
+				$tmp_str = $this->substr( $prev, ( $this->strpos( $prev, 'href="' ) + 6 ) );
```
contactashish13: @EkoJR 

I don't think _wp_link_page should be used as it is a private function. https://developer.wordpress.org/reference/functions/_wp_link_page/. 

I also don't think substr is the best way to do this. Maybe a more resilient method would be to use http://php.net/manual/en/class.domdocument.php

EkoJR: @contactashish13 It is not possible to use `wp_link_page()`. The function is for rendering purposes and has no control over the page number.
https://developer.wordpress.org/reference/functions/wp_link_pages/

I did however implement DOMDocument as requested, but between the two it does add more processing which would make it less efficient. Just a note to be mindful of with large processing times.

contactashish13: I agree with @EkoJR about processing time. @wpsmort maybe this can be tested with a large site?

contactashish13: @EkoJR maybe add a TODO to not use a private function that we can revisit later in case it cannot be fixed now?

wpsmort:
I tested this with WP5.1 RC2 on a large site with 4000 paginated posts.  It worked fine and there were no performance problems.

Issue #2192, opened 2019-02-13
Character counts not updating in Home Page Settings


wpsmort:
The character counters for Home Title and Home Description in the Home Page Settings section are not incrementing as you type.  This appears to affect Pro only but needs more testing.

michaeltorbert:
We need to determine if this used to work. If we recently broke it we can include it in 2.12, otherwise we should probably punt to 3.0.

wpsmort:
This bug only happens in Pro, the character counters works fine in free.  This last worked in v2.10.1 of Pro.

@arnaudbroes Can you track down which commit caused this.

EkoJR:
It seems it was the plugin folder name that is triggering the issue. 

To reproduce, you just rename the folder name from `all-in-one-seo-pack` to something else; be sure to re-activate plugin after renaming.

tonyzeoli:
Hey guys,

I was seeing this to, but hadn't reported it yet.


[image: photo]
Tony Zeoli
Founder, Digital Strategy Works LLC
O  (828) 412-0990  <(828) 412-0990> M  (917) 705-4700  <(917)
705-4700> E  az@digitalstrategyworks.com
<az@digitalstrategyworks.com>
W  https://digitalstrategyworks.com
Whatsapp 9177054700
Skype tonyzeoli
<http://www.facebook.com/digitalstrategyworks>
<http://www.twitter.com/dswks>
<http://www.linkedin.com/company/digital-strategy-works>
<http://plus.google.com/digitalstrategyworks>
<http://slideshare.com/tonyzeoli> <http://about.me/tonyzeoli>
<http://angel.co/tonyzeoli>
<https://calendly.com/tonyzeoli> Schedule a 30-minute Meeting
<https://calendly.com/tonyzeoli>
<https://digitalstrategyworks.com/drupal-to-wordpress-migration-for-bookstr-com/2019/02/05/2575/>
Latest
post: Drupal to WordPress Migration for Bookstr.com
<https://digitalstrategyworks.com/drupal-to-wordpress-migration-for-bookstr-com/2019/02/05/2575/>
<https://facebook.com/digitalstrategyworks>   <https://twitter.com/dswks>
<https://linkedin.com/in/+tonyzeoli>
<https://www.youtube.com/+digitalstrategyworks>


[...]

PR #2199, opened 2019-02-14
Fix enqueue count-char.js when plugin folder name is altered

EkoJR:
Issue #2192


wpsmort:
This is tested and is good.  The character counter works regardless of the plugin directory name.

Issue #555, opened 2016-10-07
Change location of all-in-one-seo-pack.log


wpsmort:
This log file is created when you enable the Log important events option under All in One SEO > General Settings.  The location is the /all-in-one-seo-pack plugin directory.  We should change the location to /wp-content instead so it's in the same location as the WP debug log.

PR #2208, opened 2019-02-16
change location of logfile to content dir #555

michaeltorbert:
#555


wpsmort:
This is tested and is good.  I tested back to PHP 5.2.

Issue #674, opened 2016-12-07
Google Webmaster Tools is now called Google Search Console


dawatorbert:
But it still is called by the old name in our settings. Maybe it should say something like (formerly Google Webmaster Tools) in the contextual help?

arnaudbroes:
All content on the website and plugin has been updated so this issue can be closed once #2213 is merged and released with 2.12.

PR #2213, opened 2019-02-18
Rename Google Webmaster Tools to Google Search Console

arnaudbroes:
#674


wpsmort:
@arnaudbroes One minor change is needed.  Google Search Console Tools should be renamed Google Search Console in the POT file.

michaeltorbert:
I was actually wondering if we should just delete the pot file since translations are handled by .org now.

arnaudbroes:
@wpsmort thanks, I fixed it. @michaeltorbert if you're 100% sure about that, then yes, let's go ahead and delete it along with the i18n folder. I'd suggest merging this first so I don't have to change this PR, then create a new issue with a PR that deletes it.

Issue #1013, opened 2017-08-02
Social Meta module - Remove Blog from Object Type drop downs


wpsmort:
Blog is no longer a valid og:type (http://ogp.me/#types) so we should remove it from the drop downs found under Social Meta > Facebook Settings > Object Type (there are drop downs for each post type). 

https://semperplugins.com/documentation/social-meta-module/#content-object-types

wpsmort:
Blog also needs to be removed from the Facebook Object Type drop down on the Social Settings tab of the Edit Post screen.

wpsmort:
@michaeltorbert I was hoping this was a simple change of just removing Blog from the drop down array in aioseop_opengraph.php.  However, we need to account for anyone who has selected Blog as their Object Type.  

If I have set one of the Object Type drop downs under Social Meta > Facebook Settings to Blog then removing that as a selection means that the Object Type reverts to the first available type which is Activity.  This is incorrect, if someone had selected Blog then we should change it to Article.

If I set the Object Type on the Social Settings tab of the Edit Post screen to Blog and we remove it then it changes to nothing and I get a PHP Notice:

PHP Notice:  Undefined index: blog in/wp-content/plugins/all-in-one-seo-pack-2.3.15/modules/aioseop_opengraph.php on line 733

Instead we should revert them to Article. 

I'm pushing this to the next release as it doesn't need to hold up this release.

michaeltorbert:
This will probably require a db migration for the entry and/or we may want to just check for that and use website if they have blog.

amostajo:
@wpsmort ready for testing. I have added the migration for when version switches to `2.3.17`. Haven't changed version on code though,

michaeltorbert:
@amostajo What does "Haven't changed version on code though" mean?

amostajo:
Means that I have not changed/pish the version of the plugin to 2.3.17, cuz I thing this is something you usually do.

wpsmort:
@amostajo I tested PR #1112 and found three issues:

1). When I update to this version I get the following error during the DB migration:

WordPress database error You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''s:35:\"aioseop_opengraph_settings_category\"' at line 2 for query UPDATE wp_postmeta
[...]

PR #2232, opened 2019-02-25
Social Meta module - Remove Blog from Object Type drop downs

contactashish13:
#1013

### DIFF
--- a/modules/aioseop_opengraph.php
+++ b/modules/aioseop_opengraph.php
@@ -76,7 +76,6 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Opengraph' ) ) {
 				),
 				'Websites'                   => array(
 					'article' => __( 'Article', 'all-in-one-seo-pack' ),
-					'blog'    => __( 'Blog', 'all-in-one-seo-pack' ),
 					'website' => __( 'Website', 'all-in-one-seo-pack' ),
 				),
 			);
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good. I tested on a large site running PHP 5.2 with over 5000 posts all set to OG:Type of Blog and there was no performance problems.

modules/aioseop_opengraph.php
```diff
@@ -203,6 +203,27 @@ function __construct() {
 			$this->init();
 		}
 
+		/**
+		 * Handle specific meta tags.
+		 *
```
EkoJR: Could you add `@since` to the phpDoc please.

modules/aioseop_opengraph.php
```diff
@@ -203,6 +203,27 @@ function __construct() {
 			$this->init();
 		}
 
+		/**
+		 * Handle specific meta tags.
+		 *
+		 * @param mixed $value The value of the meta tag.
```
EkoJR: Is this a string, or an array...or both? Using 'mixed' doesn't explain the type much, but 'string|array' would be more detailed; assuming it's 'mixed' because of an array.

modules/aioseop_opengraph.php
```diff
@@ -871,10 +892,16 @@ function filter_settings( $settings, $location, $current ) {
 								$flat_type_list[ $k ] = $v;
 							}
 						}
+						$default_fb_type = $this->options[ "aiosp_opengraph_{$current_post_type}_fb_object_type" ];
+						// https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1013
+						// if 'blog' is the selected type but because it is no longer a schema type, we use 'website' instead.
+						if ( 'blog' === $default_fb_type ) {
+							$default_fb_type = 'website';
+						}
+						$default_fb_type = $flat_type_list[ $default_fb_type ];
```
EkoJR: This looks like a potential bug. Variable/Array might not be set.

modules/aioseop_opengraph.php
```diff
@@ -203,6 +203,27 @@ function __construct() {
...
 		}
 
+		/**
+		 * Handle specific meta tags.
+		 *
+		 * @param mixed $value The value of the meta tag.
+		 * @param string $t The type of network.
+		 * @param string $k The name of the meta tag.
+		 *
+		 * @return mixed
```
EkoJR: Same with the `@return`. What would the variable types be?

PR #1892, opened 2018-09-10
Change Help Text Dropdown to Tooltips #1156

EkoJR:
Issue #1156

Issue #1156, opened 2017-09-21
Change behavior of help text area


michaeltorbert:
Instead of moving everything down and inserting the text (which uses a lot of code and has caused bugs in the past) maybe there's another way. 
Popup tooltip? How can we make this look better?

michaeltorbert:
@EkoJr discussed wanting to do this with jQuery UI Dialogs.

EkoJr:
As discussed: Implementing jQuery UI Tooltips instead.

EkoJr:
Both Tooltips and Dialogs were added for testing. Tooltips seem to work quite well, but dialogs seem problematic anyways. Here are some of the notes I added of the changes that were made.

* **Dialogs** were added on the Top Right of the metabox, inside the metabox title container (Next to the Help link).
  * Dialog content is added/included from a separate file that is unique to the metabox.
  * ISSUE: Clicking the Info/Help link multiple times causes multiple dialog boxes to appear.
  * ISSUE: Top bar will occasionally appear out of view if there is a large amount of content in a given width.
* **Tooltips** were added to the option labels on the left side. It appears on hover, rather than click.
  * Adding it to the input element area seemed possible, but obscuring the option was a worry.
  * Possibly able to customize the tooltip to show/hide on click rather on hover.
  * Could add a delay on hide.
  * Option Help Text will need to be changed to remove raw HTML that was used for links (image has page edits showing the intended version).
* jQuery UI Styles were (temporarily) added in order to render properly. It was mainly intended for testing. Unsure on this as a final method, but added as a note.
  * `\all-in-one-seo-pack\admin\aioseop_module_class.php` at `All_in_One_SEO_Pack_Module::enqueue_styles()`
  * It uses a Google CDN that is updated according to what version WP has for the scripts.
  * Uses the same method I applied here [StackOverflow: WordPress, jQuery UI CSS Files?](https://stackoverflow.com/questions/8849684/wordpress-jquery-ui-css-files/11789443#11789443)
* Dependants were added to wp_enqueue_script() for aioseop_module.js
  * In the past I’ve used dependants with WP when enqueueing script, but I noticed AIOSEO is a bit different with jQuery (didn’t seem to use jQuery UI).
  * Added a variable to JS localization in order to identify and initialize the element ui.

wpsmort:
Notes from Skype call with @EkoJr 

[...]

PR #2236, opened 2019-02-28
Change Help Text Dropdown to Tooltips

EkoJR:
Issue #1156

Some changes have occured since PR #1892 was create, and had to track all the changes to add to the new helper file. So, this will probably need testing.


wpsmort:
@EkoJR I tested this and found these items that need fixing:

1. The tooltips for the three options under Webmaster Verification box have two documentation links (see screenshot below)
2. The Specify a Preferred Name tooltip returns an error because there's no help text for this option.  We should add the help text "Enter your site name".

In addition to these, we need to determine what to do regarding Michael's two comments here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1156#issuecomment-423007123

Screenshots:
screen shot 2019-02-28 at 5 22 30 pm

screen shot 2019-02-28 at 5 24 16 pm

EkoJR:
I added the changes, and changed the cursor to the default cursor, since it seemed like most of the styling for the tooltip is done within jQuery UI.

wpsmort:
@EkoJR I tested and the two problems are fixed.  I've added my recommendations for the usability here https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1156#issuecomment-468692707

wpsmort:
This is tested and is good.

admin/aioseop_module_class.php
```diff
@@ -1785,6 +1736,16 @@ function admin_enqueue_styles( $hook_suffix ) {
 			if ( function_exists( 'is_rtl' ) && is_rtl() ) {
 				wp_enqueue_style( 'aioseop-module-style-rtl', AIOSEOP_PLUGIN_URL . 'css/modules/aioseop_module-rtl.css', array( 'aioseop-module-style' ), AIOSEOP_VERSION );
 			}
+
+			// Uses WP Scripts to load the current platform version of jQuery UI CSS .
+			$wp_scripts = wp_scripts();
+			wp_enqueue_style(
+				'aioseop-jquery-ui-css',
+				'https://ajax.googleapis.com/ajax/libs/jqueryui/' . $wp_scripts->registered['jquery-ui-core']->ver . '/themes/smoothness/jquery-ui.css',
```
contactashish13: .min file should be preferred. Google does not host it but cdnjs does.

admin/class-aioseop-helper.php
```diff
@@ -0,0 +1,1128 @@
...
+			default:
+				$this->help_text = array_merge( $this->help_text, $this->help_text_general() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_performance() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_sitemap() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_opengraph() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_robots_generator() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_file_editor() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_importer_exporter() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_bad_robots() );
+				$this->help_text = array_merge( $this->help_text, $this->help_text_post_meta() );
```
contactashish13: isn't this too wordy? array_merge can accept multiple parameters and might have better performance

admin/class-aioseop-helper.php
```diff
@@ -0,0 +1,1128 @@
...
+	 * @return array
+	 */
+	private function help_text_general() {
+		/*
+		 * Consider changing the construction of the macros.
+		 *
+		 * The name of the macro should NOT be inside _e() or __() because it does not make sense as it
+		 * won't change with the language.
+		 *
+		 * Moreover, it will confuse WPCS and it will try to replace %c (as in %category%) to %$1c.
```
contactashish13: since this PR will be run through grunt (if it hasnt already), c will become %c. maybe you should add the codestandardsignore clause so that that does not happen

js/aioseop-helper.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * Use for JavaScript compatability by using anonymous functions; which
+ * are loaded into the footer.
+ *
+ * @summary   AIOSEOP Footer JS
+ *
+ * @since     2.4.2
+ */
+// phpcs:disable PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket
```
contactashish13: we should add these to the phpcs.xml file

js/aioseop-helper.js
```diff
@@ -0,0 +1,60 @@
...
+	 * jQuery UI Tooltips
+	 *
+	 * Initiates the jQuery UI Tooltips on the admin pages with the class and
+	 * title attributes set properly.
+	 *
+	 * @since 2.4.1.1
+	 *
+	 * @link http://api.jqueryui.com/tooltip/
+	 */
+	function aioseopTooltips() {
```
contactashish13: @wpsmort I have faced an (as yet unresolved) conflict with themes that use twitter bootstrap as that interferes with jquery tooltips (https://stackoverflow.com/questions/13731400/jqueryui-tooltips-are-competing-with-twitter-bootstrap). It would be good to test with that as some of the most popular themes use that. I don't remember the theme name. cc @EkoJR

admin/aioseop_module_class.php
```diff
@@ -1829,13 +1791,28 @@ public function admin_enqueue_scripts( $hook_suffix ) {
 				wp_enqueue_media();
 			}
 
+			$helper_dep = array(
+				'jquery',
+				'jquery-ui-core',
+				'jquery-ui-widget',
```
contactashish13: is widget required?

admin/aioseop_module_class.php
```diff
@@ -1829,13 +1791,28 @@ public function admin_enqueue_scripts( $hook_suffix ) {
...
+				'jquery-ui-position',
```
contactashish13: is position required?

admin/aioseop_module_class.php
EkoJR: Good catch. ...and Google does. Changed.

admin/class-aioseop-helper.php
EkoJR: Changed

admin/class-aioseop-helper.php
EkoJR: Added. However, it would be best not to use `@codingStandardsIgnore`, and use `phpcs:disable <ruleset>`.

1. It allows us to identify which rules are being disabled.
2. It allows other rules to still be enabled.

js/aioseop-helper.js
EkoJR: Again, that is why it is added. So that it can be recognized later. This isn't a PR with phpcs.xml changes.

js/aioseop-helper.js
EkoJR: This is something that I am aware of, and tried to keep in mind while coding. It probably would be worth testing.

admin/aioseop_module_class.php
EkoJR: It is required, if you look under Dependencies. http://api.jqueryui.com/tooltip/

admin/aioseop_module_class.php
EkoJR: Yes.

js/aioseop-helper.js
contactashish13: if the changes are needed as part of this PR and they are included in the js file, I don't see why they cannot be included in the XML file. They should be removed from the js file too

admin/class-aioseop-helper.php
contactashish13: ignoring rules is required only for the block that has the sprint so it does not make sense to disable it for the entire file

contactashish13:
@wpsmort this needs testing as per https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2236/#discussion_r261824187

js/aioseop-helper.js
EkoJR: I can't just changed the grunt coding standards, when it's not even part of it. Some of it needs discussing before it can be changed.

admin/class-aioseop-helper.php
EkoJR: It is not disabled for the entire file. It is just for this code block.

Line 426 `//phpcs:enable`

js/aioseop-helper.js
contactashish13: Sure but if you have added this to the JS file means you have already decided to disable that rule from working on that file => no discussion required. Plus if you want to remove these rules from all over the workspace, then they need to be added to the XML file otherwise the current implementation is fine. I'm not sure what your intent is. 

Either ways, you have already decided that these need to be disabled so there isn't much scope left for discussion. Its confusing if you have already added these to the file and then require discussion. If you require a discussion, these should not be added to any code change and instead a new issue should be created. I'm not sure I understand the intent of adding them to a file but not to another file

EkoJR: It is not removed from all over the workspace, and this part of the code it is intended; which are also the coding standards we had issues with when reviewing your grunt PR.

https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage

wpsmort:
I have tested this with multiple Bootstrap themes from the WordPress theme repository as well as some free themes from other locations.  I did not see any conflicts.  

This is tested and is good.

js/aioseop-helper.js
contactashish13: this has been discussed on slack so I'm resolving this conversation

PR #2248, opened 2019-03-05
Update opengraph fb_object_type field to Fix issue 2227 

onetarek:
## Cause of this issue
- Facebook object type ( `aioseop_opengraph_settings_category` ) field causes the issue
- Value of first default item of the field is duplicate of another value. So when user select default value and submit then the selection is overridden by another option. Value of default item should be blank.
- When the value of the field is blank then it is being populated by the default value from the plugin option.

## Solution
- I set blank value for the default item.
- Removed a block of code to stop assigning default value from plugin option.


michaeltorbert:
I'm going to go ahead and close this as it's a complicated issue and this doesn't resolve it.

Issue #2261, opened 2019-03-07
Fix Grunt Failing on node.js test 


EkoJR:
I noticed grunt has been failing due to recent changes, and it found something to cause a warning error with the unit tests for admin_notices class. Looks like it is just a child class that doesn't have params like the parent class.

PR #2262, opened 2019-03-07
Fix warnings with unit tests for admin_notices

EkoJR:
Issue #2261


js/admin-notice.js
```diff
@@ -74,7 +76,9 @@
 	 * @param string noticeSlug
 	 */
 	function aioseop_notice_delay_wp_default_dismiss_ajax_action( noticeSlug ) {
-		var noticeNonce     = aioseop_notice_data.notice_nonce;
+		// JSHint: 'aioseop_notice_data' is not defined. (W117)
```
contactashish13: the recommended way to do this is to mention the below in the JS file

```
/* global aioseop_notice_data */
```

tests/base/class-aioseop-notices-testcase.php
```diff
@@ -490,31 +490,31 @@ protected function data_screens_wp() {
 				'url'       => site_url() . '/wp-admin/post-new.php',
 				'dir'       => WP_DEVELOP_DIR . '/src/wp-admin/post-new.php',
 			),
-//			array(
-//				'screen_id' => 'post',
-//				'url'       => site_url() . '/wp-admin/post.php?post=###&action=edit',
-//				'dir'       => WP_DEVELOP_DIR . '/src/wp-admin/post.php?post=###&action=edit',
-//			),
+			// array(
```
contactashish13: isnt it better to remove this than comment it out?

tests/classes/aioseop-notices/test-delay-time.php
```diff
@@ -90,9 +90,11 @@ protected function mock_notice() {
 	 * Override and skip.
 	 *
 	 * @since 3.0
+	 *
+	 * @dataProvider data_screens
 	 */
-	public function test_enqueue_scripts_on_screens() {
-		$this->markTestSkipped('Skip');
+	public function test_enqueue_scripts_on_screens( $screen_id, $url, $dir ) {
+		$this->markTestSkipped( 'Skip' );
```
contactashish13: maybe markTestIncomplete would be a better choice? same for the ones below. Skipped is better used for a particular condition, say dependency on PHP version.

tests/base/class-aioseop-notices-testcase.php
EkoJR: Not quite yet. I still have other related issues/PRs that I got to add soon, and is still very useful information. It would also save a lot of time from having to investigate all the screens needed.

Issue #2217, opened 2019-02-19
XML Sitemap - indexes should be reverse chronological order


michaeltorbert:
Reported here: https://wordpress.org/support/topic/bugs-with-the-sitemap-please-fix-them/

> Newest posts are always written in the first sitemap. For example, when sitemap is split in 5 chunks, 1000 posts per split, in sitemap5.xml 1000 in 4 1000, 3 1000, 2 1000, Sitemap1 will have the newest posts but when that fills up with 1000, sitemap1 will change name into sitemap 2, 2 will change name into 3, 4 into 5 and so on. This is a very bad practice which according to Google themselves causes deindexing:
https://www.seroundtable.com/google-dynamic-sitemap-file-names-25187.html
Please fix this. YOAST SEO has figured it out, its the only sitemap out there which has the oldest 1000 posts in sitemap1 and gradually the newest posts in the last sitemap, lets say 5. After 5 is full with 1000 newest posts, that will remain the same, unchanged, FOREVER, and sitemap6 will be created and filled with the newest posts. This is the most intelligent way a sitemap should behave.
In short, please put the newest posts at the bottom and the oldest on top or simply just give an option in the settings and let users choose the order.

vikingrunes:
I can't believe this isn't getting any traction and discussion. This is actually true and very important. Sub-sitemaps can't change names all the time, Google will go crazy. Its actually important to start with the oldest posts first and gradually move to the newest. Isn't milestone 3.0 kind of too far ahead? Can't this be implemented in the next update? Shouldn't be that complicated to do.

michaeltorbert:
3.0 is the next release. It’s been marked as high priority.

vikingrunes:
ohh sorry, I didn't know 3.0 is the next update, I thought there would be 2.13, 2.14 and so on, LOL. Well that's great news, I'm really happy this is getting taken care of. Cheers

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2263

PR #2263, opened 2019-03-08
XML Sitemap - indexes should be reverse chronological order

contactashish13:
#2217

### DIFF
--- a/tests/modules/sitemap/test-sitemap.php
+++ b/tests/modules/sitemap/test-sitemap.php
@@ -38,10 +38,10 @@ class Test_Sitemap extends Sitemap_Test_Base {
 
 		$posts = $this->setup_posts( 2 );
 		$pages = $this->setup_posts( 2, 0, 'page' );
-		
+
 		// create a new page with a delay so that we can test if the sitemap is created in ASCENDING order.
 		// @issue https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2217
-		sleep(1);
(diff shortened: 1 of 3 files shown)


contactashish13:
@EkoJR grunt is failing in this PR (and will fail in all PRs) because of a non-grunted change below. Can you please address this? @michaeltorbert will also have to merge this into development after your change so that PRs do not fail.

image

EkoJR:
Yea, I've been trying to address this, and the PR is #2262

tests/modules/sitemap/test-sitemap.php
```diff
@@ -39,6 +39,12 @@ public function test_only_pages() {
 		$posts = $this->setup_posts( 2 );
 		$pages = $this->setup_posts( 2, 0, 'page' );
 
+		// create a new page with a delay so that we can test if the sitemap is created in ASCENDING order.
+		// @issue https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2217
+		sleep( 1 );
```
EkoJR: Are you adding sleep because of the time it takes to process the sitemap? If so, would it be safer to make it 3 seconds?

contactashish13: i'm adding sleep so that the post is created with a different timestamp than setup_posts which might have the same timestamp down to the second depending on how fast it creates posts.

wpsmort:
This is tested and is good.

Issue #1382, opened 2017-11-27
Remove from sitemap when noindex is enabled


michaeltorbert:
Reported by Gordon via email. 

We should consider whether to remove a post from the XML sitemap with the noindex option is enabled. This makes sense, although it may make "exclude from sitemap" redundant. 
If we do this, we should consider whether to remove the latter option.

It was suggested that excluding a post from XML might prevent Google and other crawlers from reading the noindex and therefore continuing to list the sites.

arnaudbroes:
@michaeltorbert good thinking, I'd remove the "exclude from sitemap" setting. Using NOINDEX makes more sense.

contactashish13:
@michaeltorbert would this also need to be backward compatible for users who had excluded pages from sitemap and now find the option gone?

michaeltorbert:
Requested again by Serhat via email


Hello,You need to bring the "exclude from sitemap" option for the Woocommerce categories. I set "No-Index" for a category and now Google Search Console warns me to drop that category URL from the sitemap.See the screenshot: https://www.dropbox.com/s/wlt66xa5f0530ls/screenshot.47.png?dl=0
--

arnaudbroes:
I think we should just remove the Exclude from Sitemap setting altogether then and merge it with noindex.

wpsmort:
Requested again - https://semperplugins.com/support/configuring-all-in-one-seo-pack-pro/since-more-than-week-google-doesnt-index-posts/

yatgirl:
and also product tag archive pages - I have 14 product tag archive pages that I am receiving google errors on because I have set them to no index, but they are still being included in the sitemap. I am hoping this can be resolved so I can clear the errors in google webmaster very shortly.
[https://semperplugins.com/support/configuring-all-in-one-seo-pack-pro/tag-archive-sitemap/](url)

contactashish13:
what's the consensus on this approach?

michaeltorbert:
Reason **_to_** merge-

1)We have basically three places people can tell search engines not to look at a page: Sitemap options page (excluded pages), exclude this page option on the edit page itself, noindex on the edit page itself. This can be confusing.


Reason **_not to_** merge

1) backwards compatibility - as mentioned by Ashish. It would take quite a db migration to move all that over. Otherwise, we have to forever support the old settings (also a possibility).

michaeltorbert:
[...]

PR #2264, opened 2019-03-08
Remove from sitemap when noindex is enabled

contactashish13:
#1382


modules/aioseop_sitemap.php
```diff
@@ -443,13 +443,40 @@ public function __construct() {
...
 		}
 
+		/**
+		 * Adds additional arguments before posts are fetched.
+		 *
+		 * @param array $args Arguments needed by get_posts/WP_Query.
+		 *
+		 * @return array
+		 */
+		function filter_posts( $args ) {
```
EkoJR: This concept would probably be better where the filter is being applied since this seems like this would be core concept. Was there a reason for preferring an action hook?

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/8bdb9d1c564ec607ae10dbe2821fa228e98d094b/modules/aioseop_sitemap.php#L4183

modules/aioseop_sitemap.php
```diff
@@ -443,13 +443,40 @@ public function __construct() {
...
+					'compare'   => 'NOT EXISTS',
+				),
+			);
+			$meta_query = $query;
+			if ( array_key_exists( 'meta_query', $args ) && ! empty( $args['meta_query'] ) ) {
+				$meta_query = $args['meta_query'];
+				$meta_query = array_merge( $meta_query, $query );
+			}
+
+			$args['meta_query'] = $meta_query;
```
EkoJR: Caution when adding meta-queries. There may be no other alternative for this, but this will need to be bulk tested.

tests/modules/sitemap/test-sitemap.php
```diff
@@ -57,6 +57,37 @@ public function test_only_pages() {
 		);
 	}
 
+	/**
+	 * Creates pages, marks some as NOINDEX and tests whether only the non-NOINDEX pages are being shown in the sitemap.
```
EkoJR: Do you think my format might help you with documenting unit tests? Especially for future reference.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/3.0/tests/base/class-aioseop-notices-testcase.php#L319

modules/aioseop_sitemap.php
contactashish13: the filter hook already existed and I'm merely using it. I don't see an action hook. Which one are you referring to?

modules/aioseop_sitemap.php
contactashish13: agreed. maybe you can create a label and tag this PR so that we are aware of what all needs to be specially bulk tested?

tests/modules/sitemap/test-sitemap.php
contactashish13: good idea. have used that.

wpsmort:
@contactashish13 I have tested this and here are my comments:

- If I select NOINDEX on the Blog page, then the page still shows up.  I can see how this should be the case for the Home page but not the Blog page.

- If I have Sitemap Indexes enabled and I select NOINDEX for all of the  posts in a post type then we generate an empty sitemap for that post type.  To reproduce this, enable Sitemap Indexes on a site with only one Post.  Verify the post shows up in sitemap_post.xml.  Then edit the post and select NOINDEX.  Now go to the sitemap index and you'll see sitemap_post.xml is still listed and when you navigtate to it you'll see it's empty.

Ping me on Slack if you have any questions or need help reproducing either item above.

modules/aioseop_sitemap.php
EkoJR: Your using a filter that's applied in a function that is only used by the sitemap module. Which makes the prefix for the filter redundant when it is just the sitemap module. This could be done the same way exclude is being done.

This would further spread out code, and it seems confusing; which the prefix itself makes it increasingly difficult to track.

tests/modules/sitemap/test-sitemap.php
EkoJR: Looks good, but isn't `Actual: As expected; no current issue.` suppose to mention the issue it is fixing? It helps to show why the test was initially made. I thought this was more of a bug where there was an unintended result occuring.

The other PR I linked was new code, so there was no prior actual for me to add.

modules/aioseop_sitemap.php
michaeltorbert: So I'm not a big fan of the way we use a variable prefix due to the decreased readability, but we do need to prefix them so there are no conflicts.

EkoJR: @michaeltorbert The thing is, the function is only used by the sitemap module. There is no other module that uses that filter.

michaeltorbert: @EkoJR Maybe we're not talking about the same thing. Are you talking about `add_filter( $this->prefix . 'post_query', array( $this, 'filter_posts' ), 10, 1 );` and recommending we use an unprefixed filter called post_query?

EkoJR: @michaeltorbert No, but that is the filter. I was recommending not using the filter. Using it just seems increasingly complex; especially with the prefix.

If anything, a prepare function came to mind, but I'm not too sure about it.

contactashish13:
@wpsmort have addressed both issues. Additionally, I have also added support for **Default to NOINDEX** in general settings.

wpsmort:
@contactashish13 I tested this and found that if you have more than one post type in your sitemap then selecting Default to noindex for one post type doesn't work.

Here's the steps to reproduce this:
- Create your sitemap and include posts and pages
- Verify your sitemap shows all your posts and pages
- Go to Noindex Settings and select the checkbox for Posts
- View your sitemap and you'll see posts are still included

It works fine if you only include posts in your sitemap.

It also no longer works if you just select Noindex box on the Edit Post / Edit Page screen.

arnaudbroes:
@contactashish13 for the sake of simplicity, let's keep this with you as the PR is coming from your own forked repository. I'm not really great with CSS anyway.

modules/aioseop_sitemap.php
contactashish13: if there is a filter we should use it - its not only for other people to extend behavior but its for us to use it as well instead of using a prepared function. Any kind of hook does increase complexity (blame it on WP) but it also increases flexibility. I don't see any reason to not use the hook that already exists.

contactashish13:
@wpsmort this is a known bug in this process. I've captured in the code as comments that I'm reproducing here. This case does not work only for non-indexed sitemaps and it is a bug because of the following reasons:

- a single query is run with multiple post types (it is an IN query)
- the value of the post_meta value **_aioseop_noindex** to use is confusing because it can have 2 values at the same time (see the truth table below)
- both values cannot be included in the same query because one value might be relevant for a **post** while another value might be relevant for a **page** (I've explained a potential scenario below) 

```
				Here is the truth table of this behavior:

				NOINDEX (general)	|	NOINDEX (post)	|	value of _aioseop_noindex	|	Expected Behavior
				unchecked				checked				on								exclude
				unchecked				unchecked			does not exist					include
				checked                 default             does not exist                  exclude
				checked					index				off								include
				checked					noindex				on								exclude

				The conflict is that when the value of _aioseop_noindex does not exist, it can mean both include and exclude.
				The actual context can only be determined by knowing what the value is in NOINDEX (general).

				The confusion becomes manifold when multiple post types are being queried e.g.
				POST, NOINDEX (general), INDEX(post), value of _aioseop_noindex = off
				PAGE, INDEX (general), INDEX(post), value of _aioseop_noindex = does not exist
				but the meta_query is run on all the post_types included. This happens in only one case and we will filter that out.
```

@arnaudbroes @EkoJR are welcome to suggest an alternative

EkoJR:
You may have to include all the post_types in the query, but exclude the noindex post types.

contactashish13:
@EkoJR if you look at the truth table you will see that its not that simple. A single query is run for all post types selected. Even if a post_type is NOINDEX, a post under it can be marked as INDEX. The value for an INDEXed post whose post_type is **marked** NOINDEX is **not the same as** the value for a post whose post_type is **not marked** NOINDEX. 

Thus the same query has to include the meta_query with a variable that can have 2 values for the same behavior. We can discuss this further in the meeting.

contactashish13:
I feel how the settings behave here is messed up and contravenes logic. That is the reason I believe this should be fixed instead of trying to introduce a way-around this bug and complicating things.

Here are a few scenarios of this buggy behavior:

##scenario 1
- post_type XXX is **marked** NOINDEX
- a post of type XXX is marked **default**  => value of post_meta **does not exist**
- a post of type XXX is marked **noindex** =>value of post_meta **on** 
- a post of type XXX is marked **index**     =>value of post_meta **off** 

The value of the post_meta in 2 and 3 should be the **same** but they seem to have different values for the **same behavior (noindex the post)**.

##scenario 2
- post_type XXX is **marked** NOINDEX
- post_type YYY is **not marked** NOINDEX
- a post of type XXX is marked **default**  => value of post_meta **does not exist**
- a post of type YYY is marked **noindex**   =>value of post_meta **on** 

The value of the post_meta in 3 and 4 should be the **same** but they seem to have different values for the **same behavior (noindex the post)**.


##scenario 3
- post_type XXX is **marked** NOINDEX
- post_type YYY is **not marked** NOINDEX
- a post of type XXX is marked **index**     =>value of post_meta **off** 
- a post of type YYY is left alone                   =>value of post_meta **does not exist**

The value of the post_meta in 3 and 4 should be the **same** but they seem to have different values for the **same behavior (index the post)**.

In order to fix this behavior and make the value of post_meta consistent, we need to fire a one-time process that goes through each post_type and aligns its posts correctly with the following assumptions in mind:

- the post_meta of a post that is **not** **noindex**ed **should not exist** (this would be the default behavior for most post types/posts over the user base)
- the post_meta of a post that is **noindex**ed should have a value of **on** (this is because someone actually wants a deviation from the norm)
- if someone has chosen to **noindex** a post_type, they would most probably not select **index** or **noindex** in the post settings.
- if someone has chosen **not to** **noindex** a post_type, they would most probably not select **index**.
- assuming all of the above, we fix the existing posts by
- fetching all posts whose post_types are marked as **noindex** and whose post_meta value **does not exist** (they wanted to force-noindex these posts, so this should be a very small number) and making the post_meta value **on**
  - fetching all posts whose post_types are marked as **noindex** and whose post_meta value is **off** (they wanted to force-index these posts, so this should be a small number) and deleting the post_meta value

@semperfiwebdesign/aioseop-core-committers comments/suggestions?

arnaudbroes:
@contactashish13 what do you mean with "one-time process"? A database operation of some sorts? I'm trying to figure out whether this is something we need to do to rectify the post_meta table for our existing users or whether this is something new users will need also. In other words, will each existing user have to run this function just once or whenever they change something to their noindex settings.

contactashish13:
One time means when it will run only once when the plugin is updated similar to all the upgrade related one time functions that run. It is a one time fix to fix wrong settings.

⁣--
Regards,
Ash​

wpsmort:
Closing this in favor of PR #2317

Issue #1361, opened 2017-11-13
Sitemap indexes cannot be viewed when compression is enabled


arnaudbroes:
![image](https://user-images.githubusercontent.com/18156697/32740778-4b1aadd2-c8a4-11e7-8e84-e237a58aa285.png)

contactashish13:
@arnaudbroes can you please let me know the steps to reproduce this issue? Can I also get access to the cloned site and the SSH instance? If you want, you can make this site available to me through cloudways - will send my email ID on slack.

arnaudbroes:
@contactashish13 we don't have a cloned site for this issue. 
Follow these steps to reproduce the issue:

1. Install a fresh, clean test site
2. Install All in One SEO Pack 
3. Create a few dummy posts and pages
4. Activate the XML Sitemap Module in the Feature Manager
5. Check if the 'Create compressed sitemap' setting is enabled
6. Enable the 'Enable sitemap indexes' setting
7. Update and view the sitemap

contactashish13:
What should I find in the sitemap and what is the expected behavior?

arnaudbroes:
@contactashish13 if you go to /sitemap.xml/, you'll see that the browser displays URLs with the .gz extension, which shouldnt be the case. In addition, when you click on any of the URLs in the sitemap, you won't be able to access them (or depending on the browser it will download the .gz file).

Normal behaviour is that the sitemap does not display the .gz extension and every URL can be opened.

contactashish13:
@arnaudbroes I'm not sure I understand the problem. From what I have been reading (e.g. https://www.design19.org/blog/optimize-sitemap-xml-gz-better-seo/) a compressed sitemap is indeed a gzipped file. If thats not the case, then what should compression really do?

michaeltorbert:
@contactashish13 The problem is that when sitemap indexes is enabled, the compressed file is a 404.

tonyzeoli:
Please note that when I disabled compression on https://janetwhitworth.com, then resubmitted the sitemap to GWT, all 6 errors went away. So it's only the compression that throws the 404 errors in GWT.

EkoJr:
This appears to be related to #1428. Which also has a PR.

tonyzeoli:
I just got this on a brand new site I'm setting up - 5pointz.com.

1.  Activated XML Sitemaps.
2. Gzip was already enabled, so I didn't touch that.
[...]

PR #2269, opened 2019-03-09
Sitemap indexes cannot be viewed when compression is enabled

contactashish13:
#1361


michaeltorbert:
Since more discussion is needed about the issue, I'm closing this PR as it's not production ready and shouldn't be tested. https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1361#issuecomment-471150329

Issue #1715, opened 2018-05-26
UI - Add option to disable animations or make transition better


arnaudbroes:
One of our users (Thomas - via email) reported that he occasionally suffers from motion sickness due to the animations that occur when he switches between the SEO and Social Meta tabs. He asked us whether we'd consider adding a setting to turn them off.

[edited by Michael to remove Thomas' email address]

michaeltorbert:
I think this is a great idea. We should probably at a minimum have a way to disable it, even better would be to find a different transition that would make it more accessible for people who may have motion sickness issues.

Someone can feel free to dive into the javascript and come up with something better.

contactashish13:
I've made this use fadein/fadeout instead of show/hide. In medium vs. slow fade, I find slow more graceful.

PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2271

I see that we are using a bespoke implementation of tabs where as now we can use https://jqueryui.com/tabs/ as jqueryui is being used extensively in other parts of the plugin as well. Eventually, tabs should also make it to the main settings page where each module is a tab than a separate menu item.

@EkoJR @arnaudbroes @wpsmort suggestions?

arnaudbroes:
I'm all for switching custom code to framework code. As for the settings, I agree we should move all modules to a single settings page at some point.

wpsmort:
@contactashish13 I've opened #2274 for implementing jQuery Tabs and assigned it to you for this release.  We already have #127 open for your suggestion of using tabs on the main settings screens but that will be done in a future release.

michaeltorbert:
Does https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2274 make this issue obsolete?

michaeltorbert:
Let's reach out to Thomas and see if it's resolved for him.

arnaudbroes:
I've reached out to Thomas today.

arnaudbroes:
Here's what Thomas said (although he no longer uses AIOSEOP):

> If it’s still an animation – as opposed to simply switching – the likelihood of it triggering a spell of dizziness and nausea remains.

michaeltorbert:
Closing as only one person has reported it.

PR #2271, opened 2019-03-11
Add option to disable animations or make transition better

contactashish13:
#1715

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -19,6 +19,12 @@ _Put an `x` in the boxes that apply. You can also fill these out after creating 
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have added necessary documentation (if appropriate)
 
+## Testing instructions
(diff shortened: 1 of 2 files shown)


arnaudbroes:
@wpsmort perhaps assign this and probably all other similar issues to me after you're done testing for Windows browser testing.

wpsmort:
@contactashish13 I tested this in Chrome, Firefox and Safari on Mac and found that the slow fade works when switching from the Main Settings tab to the Social Settings tab but it does not work when switching back to the Main Settings tab.

michaeltorbert:
@arnaudbroes If multiple issues have overlapping code changes we're just going to double, triple, or more our efforts due to merge conflicts.
Let's try to get PRs completed as quickly as possible.

contactashish13:
I think this PR no longer makes sense if https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2274 is implemented. Let me know how to proceed.

michaeltorbert:
Are the transitions in #2274 improved for motion sickness?

Issue #2190, opened 2019-02-12
Non-existing /sitemap_XXX URLs navigate to empty sitemap instead of 404


arnaudbroes:
Some examples:

https://semperplugins.com/sitemap__arnaudiscool.xml
https://semperplugins.com/sitemap_weneedmorearnaud.xml
https://semperplugins.com/sitemap_arnaudforpresident2020.xml

The underscore seems to matter as this works fine and redirects to a 404:
https://semperplugins.com/sitemapredirectto404.xml

michaeltorbert:
Is this new?

arnaudbroes:
I checked. Been since sitemap indexes were introduced.

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2272

PR #2272, opened 2019-03-11
Non-existing /sitemap_XXX URLs navigate to empty sitemap instead of 404

contactashish13:
#2190


arnaudbroes:
This is good.

modules/aioseop_sitemap.php
```diff
@@ -2033,6 +2049,9 @@ public function do_build_sitemap( $sitemap_type, $page = 0, $filename = '', $com
 			if ( ( 'root' === $sitemap_type ) && ! empty( $this->options[ "{$this->prefix}indexes" ] ) ) {
 				return $this->build_sitemap_index( $sitemap_data, sprintf( $comment, $filename ) );
 			} else {
+				if ( empty( $sitemap_data ) ) {
+					return;
```
EkoJR: For consistency, this should probably return an empty value.

modules/aioseop_sitemap.php
```diff
@@ -1514,7 +1515,22 @@ public function sitemap_output_hook( $query ) {
...
+				$content = $this->do_rewrite_sitemap( $sitemap_type, $page );
+
+				// if the sitemap has no content, it's probabaly invalid and is being called directly.
+				// @issue https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2190
+				if ( empty( $content ) ) {
+					$query->set_404();
+					status_header( 404 );
+					header( "Content-Type: text/html; charset=$blog_charset", true );
+					nocache_headers();
+					include( get_404_template() );
```
EkoJR: Would it be better to use `get_template_part()`? 

https://developer.wordpress.org/reference/functions/get_template_part/

contactashish13: this is the recommended method

modules/aioseop_sitemap.php
contactashish13: done

modules/aioseop_sitemap.php
```diff
@@ -1987,6 +2003,9 @@ public function do_build_sitemap( $sitemap_type, $page = 0, $filename = '', $com
 			if ( ( 'root' === $sitemap_type ) && ! empty( $this->options[ "{$this->prefix}indexes" ] ) ) {
 				return $this->build_sitemap_index( $sitemap_data, sprintf( $comment, $filename ) );
 			} else {
+				if ( empty( $sitemap_data ) ) {
+					return null;
```
EkoJR: This is not an empty variable.

contactashish13: is_empty works on both empty or null

EkoJR: Yea, but it's not the same case type; this creates a void return.

contactashish13: as discussed on slack, I've changed the return type to `string|null`. Your preference is returning a string and mine is to return a null, which is not an empty string. Both work because `empty()` supports both. Its down to personal preference. Even WP Core (https://github.com/WordPress/WordPress/blob/master/wp-includes/author-template.php#L21) exhibits this behavior so there's not much to choose.

EkoJR: @michaeltorbert For consistency, this should return an empty string. Instead, it's checking if the variable is an empty string, and then returning null for some odd reason. Null is not the same case type as a string, and if the person coding doesn't expect it, it could lead to issues in the future.

Initially this was a void return, but essentially the PHPDoc is being changed as a personal preference to return null; when everything else returns a string.

I believe it would be better to keep code consistent, and just return an empty string.

contactashish13: Actually, its checking for empty() (http://php.net/manual/en/function.empty.php) and this behavior is fine as per WP Core (its there in multiple core files). Its checking for empty() even before returning a null, after which also its checking for empty. If it were checking for 1!== ''` I would return a string instead.

michaeltorbert: I'm not sure I understand why it's an issue, as long as the return doc specifies that it could return null.

EkoJR: @michaeltorbert There is no practical or logical reason to change the return type. That just forces checking all throughout the code base. With a string, you know your getting the return type, and then now all of a sudden we have to check, **every single time**. Something that I would consider as a recipe for disaster.

contactashish13:
I again say there is no type checking happening so the question of checking type doesn't arise.

michaeltorbert:
> I again say there is no type checking happening so the question of checking type doesn't arise.
> […](#)
> On Mar 16, 2019, 15:13, at 15:13, EkoJR ***@***.***> wrote: EkoJR commented on this pull request. > @@ -1987,6 +2003,9 @@ public function do_build_sitemap( $sitemap_type, $page = 0, $filename = '', $com if ( ( 'root' === $sitemap_type ) && ! empty( $this->options[ "{$this->prefix}indexes" ] ) ) { return $this->build_sitemap_index( $sitemap_data, sprintf( $comment, $filename ) ); } else { +	if ( empty( $sitemap_data ) ) { +	return null; @michaeltorbert There is no practical or logical reason to change the return type. That just forces checking all throughout the code base. With a string, you know your getting the return type, and then now all of a sudden we have to check, **every single time**. Something that I would consider as a recipe for disaster. -- You are receiving this because you were assigned. Reply to this email directly or view it on GitHub: [#2272 (comment)](https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2272#discussion_r266196059)

modules/aioseop_sitemap.php
michaeltorbert: This is what I'm seeing:

@contactashish13 - Either option A or option B is the same for practical purposes in this instance, and I have a personal preference for option A, but it doesn't ultimately doesn't really matter from a practical/functional standpoint.
@EkoJR - Either option A or B may be fine, but there's at least a reasonable possibility that B is a better choice on down the road, so therefore we should choose option B to be on the safe side.

Please let me know if I'm missing something or if that pretty much sums it up. If that's accurate, we should obviously go with B right?

EkoJR: I see no reason for A, and we should continue using B. This is a simple change request too just to change the code to `return '';`

I've shown logical concern, and I'm a bit surprised changing the return type would be a personal preference. That is something that should be avoided; all it does is muddle up the code. Making return values inconsistent and more complex to handle.

contactashish13: From my point of view, the logic is pedantic and does not currently and will never in the future affect functionality. Nevertheless, in order to help move things forward I have fixed the code as requested.

EkoJR: With the way the Sitemap has caused a lot problems for others from commits that were made, and the amount of work it took to restore. There's good reason to be concerned. 

You may consider it pedantic, but I'm just trying to be mindful of the function's concept. If our return types are less complex, they are easier to handle; rather than having to double check the work a function just did.

modules/aioseop_sitemap.php
```diff
@@ -1468,7 +1472,22 @@ public function sitemap_output_hook( $query ) {
 				if ( $gzipped ) {
 					ob_start();
 				}
-				$this->do_rewrite_sitemap( $sitemap_type, $page );
+
+				$content = $this->do_rewrite_sitemap( $sitemap_type, $page );
+
+				// if the sitemap has no content, it's probabaly invalid and is being called directly.
+				// @issue https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2190
+				if ( empty( $content ) ) {
```
EkoJR: Was there any other way to make certain the post exists?

I'm not sure if you have a post id available to use with [post_exists()](https://developer.wordpress.org/reference/functions/post_exists/).

Issue #2274, opened 2019-03-11
Use jQuery Tabs on Edit Post screen


wpsmort:
From #1715:

I see that we are using a bespoke implementation of tabs where as now we can use https://jqueryui.com/tabs/ as jqueryui is being used extensively in other parts of the plugin as well.

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2290

wpsmort:
@contactashish13 I don't like the design of the new tabs.

I'd like to see us improve on how the tabs currently look.  Here is an example of a design I like that's simple, clean and looks good - https://tympanus.net/Blueprints/FullWidthTabs/

contactashish13:
I'm reopening this in view of https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2290#issuecomment-475834342

PR #2288, opened 2019-03-13
Travis build fails for specific WP_VERSIONs because of PHPUnit version

contactashish13:
#2287

PR #2290, opened 2019-03-13
Use jQuery Tabs on Edit Post screen

contactashish13:
#2274


contactashish13:
This PR obviates the need for https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2271 as the transition is built into this as well.

PS - The PR will faill till #2288 is merged.

contactashish13:
yes it incorporates the motion sickness related transitions

wpsmort:
@contactashish13 I've tested and found the following:

- Line 1235 of aioseop_module.css changes the styling of the help icons, preview snippet h3 and the Debug This Post button (Social Settings tab)
- The style of the tabs have changed.  I've commented here https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2274#issuecomment-472511002 on suggestion for improving the design

EkoJR:
Not sure why I'm being requested a review when this isn't ready for review.

michaeltorbert:
@EkoJR That was prior to Steve sending it back to development.

EkoJR:
@michaeltorbert Yea, it was prior. Lately I've been getting a lot of requested reviews when there hasn't even been any testing. I can understand if it is to request feedback on something being worked on, but there is no mention why I'm needed so soon.

contactashish13:
@wpsmort I'm not sure what aspects of https://tympanus.net/Blueprints/FullWidthTabs/ to follow so I've tried to bring the style as close to the old style as possible and also addressed your other concern. Please review. If you see anything awry can you post a screenshot so that I see exactly what you see?

@EkoJR I think https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2290#issuecomment-472677869 is my fault. I saw that I could add a reviewer (separate from assigning the PR; I wonder how to use that "add a reviewer" feature though) and I did that without realizing that you'd get a notification immediately. I'll desist now :smile:

wpsmort:
This is tested in Chrome, Safari and Firefox on Mac and is good.  @arnaudbroes Can you test on Windows please.

arnaudbroes:
@contactashish13 this will need some additional CSS changes:

- background of tabs shouldn't be an image, just a solid colour
- active tab should have bold text
- move tabs a few pixels to the right
- see if it's possible to remove the border underneath the tab one way or another; I realize this is actually coming from the box underneath but it would be great if they flow into each other

contactashish13:
@arnaudbroes do you want to take a crack at it since this is just CSS?

contactashish13:
@arnaudbroes have incorporated your suggestions. Let me know if its alright now.

css/modules/aioseop_module.css
```diff
@@ -1264,6 +1227,57 @@ div.aioseop_notice a.aioseop_dismiss_link {
...
+.aioseop_tabs .ui-widget-content .ui-state-default,
+.aioseop_tabs .ui-widget-header .ui-state-default {
+	background-image: none !important;
+}
+
+.aioseop_tabs .ui-state-active a {
+	font-weight: bold !important;
+}
+
+.aioseop_tabs.ui-tabs .ui-tabs-nav li {
```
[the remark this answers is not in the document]
contactashish13: @arnaudbroes as discussed with @wpsmort it seems you were seeing something different from him and thats why you suggested the last point. I misunderstood it (that's why in the future please attach screenshots so that we see the same thing) and removed the bottom border from all tabs. Steve was trying to contact you to see if you can reproduce and fix this at your end because what he sees is what I see, which is probably different from what you see.

arnaudbroes:
@contactashish13 I'd make one more change and that is removing the border in .aioseop_header_tabs li a.aioseop_header_tab so that it looks like b) instead of a).

a) with border
image

b) without border
image

wpsmort:
I think it need the border under the inactive tab, which is how the current tabs are styled.  See below:

Screen Shot 2019-03-18 at 10 46 01 AM

arnaudbroes:
@wpsmort agreed.
@contactashish13 please implement that too, and I'd also reduce the line height a bit so that the tab looks as close to the original as possible.

So the required changes are:

1.  Remove border on tab (see above for exact class)
2. Add bottom border to inactive tab (see Steve's screenshot)
3. Reduce line height so it looks as close to the original, current design

contactashish13:
@arnaudbroes @wpsmort done

arnaudbroes:
@contactashish13 thanks, looks much better now.

admin/aioseop_module_class.php
```diff
@@ -1737,15 +1737,12 @@ function admin_enqueue_styles( $hook_suffix ) {
...
-				'https://ajax.googleapis.com/ajax/libs/jqueryui/' . $wp_scripts->registered['jquery-ui-core']->ver . '/themes/smoothness/jquery-ui.min.css',
-				false,
-				AIOSEOP_VERSION,
-				false
-			);
+			// Uses WP Scripts to load the current platform version of jQuery UI CSS.
+			// Loads only if it is not already loaded in order to not cause a conflict between 2 different versions.
+			if ( ! wp_style_is( 'jquery-ui', 'registered' ) && ! wp_style_is( 'jquery-ui', 'enqueued' ) ) {
+				$wp_scripts = wp_scripts();
+				wp_enqueue_style( 'jquery-ui', '//ajax.googleapis.com/ajax/libs/jqueryui/' . $wp_scripts->registered['jquery-ui-core']->ver . '/themes/smoothness/jquery-ui.min.css', false, AIOSEOP_VERSION, false );
```
EkoJR: I believe `jquery-ui` needs to be prefixed to prevent conflict with other plugins. I've had experiences in the past where other plugins use the same id to load a different file/version. Either ours before theirs, or theirs before ours.

The line should also be broken down to multiple lines (like it was); especially for long lines like this.

admin/aioseop_module_class.php
```diff
@@ -1770,6 +1767,7 @@ function admin_enqueue_styles( $hook_suffix ) {
 		public function admin_enqueue_scripts( $hook_suffix ) {
 			wp_enqueue_script( 'sack' );
 			wp_enqueue_script( 'jquery' );
+			wp_enqueue_script( 'jquery-ui-tabs' );
```
EkoJR: Is `jquery-ui` no longer required?

css/modules/aioseop_module.css
```diff
@@ -1264,6 +1225,48 @@ div.aioseop_notice a.aioseop_dismiss_link {
 	vertical-align: bottom;
 }
 
+.aioseop_tabs.ui-widget-content {
```
EkoJR: Do you think this would a good time to create a *-ui.css file for jquery-ui stuff?

js/modules/aioseop_module.js
```diff
@@ -540,31 +540,10 @@ jQuery( document ).ready(
...
-				var stringref = jQuery( this ).attr( "href" ).split( '#' )[1];
-				jQuery( '.aioseop_tab:not(#' + stringref + ')' ).hide( 'slow' );
-				jQuery( '.aioseop_tab#' + stringref ).show( 'slow' );
-				jQuery( '.aioseop_header_tab[href!="#' + stringref + '"]' ).removeClass( 'active' );
-				jQuery( '.aioseop_header_tab[href="#' + stringref + '"]' ).addClass( 'active' );
-				return false;
-			}
-		);
-
+        jQuery('.aioseop_tabs').tabs({
```
EkoJR: Even though this is small, do you think this would also be a good time to have a seperate js file for jquery-ui?

admin/aioseop_module_class.php
contactashish13: I don't agree that we should be in a position to load 2 different CSS for jquery-ui. That's why the check for whether the style has been loaded or not. Unless you have an example I can take a look at to understand this better.

admin/aioseop_module_class.php
contactashish13: according to https://developer.wordpress.org/reference/functions/wp_enqueue_script/ it does not seem to be a needed dependency

css/modules/aioseop_module.css
contactashish13: Maybe if there are more rules. But it would also be a performance hit by loading 2 resources so I don't think it is required at this point in time.

js/modules/aioseop_module.js
contactashish13: again, same as above.

admin/aioseop_module_class.php
EkoJR: Loading 2 jQuery UI is still going to happen regardless when plugins use a different name/id. I had also thought about loading 2, but I've found it has been better to load the version we intend to use.

Also, using a generic name is better left with WordPress; which would be the most likely scenario. So, WP would have priority in that case.

Plus, this PR isn't intended to fix an issue like that. So, it's a bit odd why this was even changed when there is no reported issue. The IF check is good, but it's name/id that has me concerned.

contactashish13: The PR didn't "fix" it. I made my changes first and when I synced your changes also appeared. So I decided to use the correct method. I would go with the generic naming than a plugin specific name unless you have a specific real-life example of a problem that this can cause.

EkoJR: Ok, I see where you made changes.

It's been awhile, but here on some of the issues I had tracked...

https://github.com/Advanced-Post-List/advanced-post-list/issues/26
https://github.com/Advanced-Post-List/advanced-post-list/issues/44
https://github.com/Advanced-Post-List/advanced-post-list/issues/74

1) Multiple versions are still loaded, regardless.
2) If there is a different script loaded before aioseop's with the same id, it will prevent ours.
3) Generic names are better left with WP, and even WP says to "prefix everything".
https://www.sitepoint.com/enqueuing-scripts-styles-wordpress/
https://make.wordpress.org/core/2018/12/06/javascript-packages-and-interoperability-in-5-0-and-beyond/
https://developer.wordpress.org/plugins/plugin-basics/best-practices/#avoid-naming-collisions
https://wpgurus.net/enqueue-scripts-style-sheets-on-shortcode-pages/
4) The most likely scenario is WP enqueueing the script; if and when it does get added. All other plugins are expected to prefix everything.
5) When debugging the script, it would help to prefix it to ensure we are enqueueing and can check if it did enqueue in WP_Scripts.

contactashish13: That is your plugin so I was looking for behavior that someone else has implemented just to understand how others go about it. Your implementation will not differ from your suggestion :)

We have 2 alternatives:

If we prefix:
- we don't have to use the `wp_style_is`
- 2 different CSS can be loaded. How will that affect the UI?

If we don't prefix
- we will retain `wp_style_is`
- if someone else has loaded another CSS theme different from ours with the generic prefix, it will be loaded. How will that affect the UI?

I can't find any source that says one method is better over the other. It would have been easier to decide.

@michaeltorbert will need your inputs here.

EkoJR: You asked for real life examples, and I thought you wanted my past examples. 

I still added 4 other links, 2 of them are from WP. Which basically reiterates what I've been saying.

> For vendor scripts in particular, there is a chance for conflict if a plugin or theme registers their own copy of the dependency with the same un-prefixed name. There is a greater likelihood if there is a large version mismatch between the registered scripts.
>
> In general, a plugin should seek to avoid naming collisions by following the best practice of “Prefix Everything” as prescribed in the Plugin Handbook. While this can sometimes result in multiple copies of a script being present in a page, it reduces the chance for breakage and grants WordPress the option to use the unprefixed name in the future.

https://make.wordpress.org/core/2018/12/06/javascript-packages-and-interoperability-in-5-0-and-beyond/

Which then links to this https://developer.wordpress.org/plugins/plugin-basics/best-practices/#avoid-naming-collisions.

contactashish13: Makes sense. I've made this change.

contactashish13:
@michaeltorbert some changes were made by @EkoJR after approving the code and I don't agree with those changes. I pinged him on slack but didn't get a response. I don't think the changes made make sense and they should be reverted.

admin/aioseop_module_class.php
```diff
@@ -1737,15 +1737,18 @@ function admin_enqueue_styles( $hook_suffix ) {
...
-			wp_enqueue_style(
-				'aioseop-jquery-ui-css',
-				'https://ajax.googleapis.com/ajax/libs/jqueryui/' . $wp_scripts->registered['jquery-ui-core']->ver . '/themes/smoothness/jquery-ui.min.css',
-				false,
-				AIOSEOP_VERSION,
-				false
-			);
+			// Uses WP Scripts to load the current platform version of jQuery UI CSS.
+			// Loads only if it is not already loaded in order to not cause a conflict with a generic version.
+			if ( ! wp_style_is( 'aioseop-jquery-ui', 'registered' ) && ! wp_style_is( 'aioseop-jquery-ui', 'enqueued' ) ) {
```
EkoJR: Do you mean this line that you forgot to change?

contactashish13:
I didn't forget to change it. It did not need to be changed. Why are we checking whether our own resource is loaded already? Who else but us will load our own resource? It's totally redundant. My comment above the code makes it clear that we check whether the default style is loaded or not and only then load our style.

Also, I'm not sure it is good practice to change code after approving a code review. I realize it must be inadvertent but I wanted to point it out that had it been discussed with me prior to approval I would have had the same point of view - that it's not required.

michaeltorbert:
@contactashish13 @EkoJR 

Regarding making code changes after code review I agree, this shouldn't happen. If changes are made to the code, it needs to be reviewed, tested, etc again.

For this instance, I'm not sure checking to see if our style is already registered is useful for the reasons Ashish noted. 
On a similar topic, why are we loading our own jquery anyway since it ships with WP?

EkoJR:
@michaeltorbert The script does, **style doesn't** though, and it's been like this for years now. https://stackoverflow.com/questions/8849684/wordpress-jquery-ui-css-files

So, there is no default style...

We also perform this check concept with our scripts. Where we prevent loading our resources multiple times.

On Ashish's note, prior to the merge, **there was no check**, but it doesn't make sense to check for others; all that's going to do is cause potential glitches for not loading the style we intend.

Which WP also states. 

> For vendor scripts in particular, there is a chance for conflict if a plugin or theme registers their own copy of the dependency with the same un-prefixed name. There is a greater likelihood if there is a large version mismatch between the registered scripts.
>
> In general, a plugin should seek to avoid naming collisions by following the best practice of “Prefix Everything” as prescribed in the Plugin Handbook. While this can sometimes result in multiple copies of a script being present in a page, it reduces the chance for breakage and grants WordPress the option to use the unprefixed name in the future.

https://make.wordpress.org/core/2018/12/06/javascript-packages-and-interoperability-in-5-0-and-beyond/

So, we could remove the IF check altogether, OR we can follow the id we are checking for. Why check for other plugins when WP states to "prefix everything"?

contactashish13:
Removing the if check will cause us to load our styles + the styles if/when they get introduced in core. The if check is mandatory to avoid that. The if check should be corrected back to checking core styles, not our own styles.

In all our code, we don't check whether we have loaded our own styles or scripts, so this change is obviously incorrect.

EkoJR:
The StackOverflow question, is a questions I answered way back in 2012, and it still hasn't changed. Do you expect WP to change that any time soon?

The last time this trac ticket was active was 2 years ago. [Bundled jQuery UI should have CSS](https://core.trac.wordpress.org/ticket/18909)

Also, where is does it say that it is mandatory to add an IF check? WP says the opposite in their article regarding vender resources, and that's recent too. 

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/f609f84486a12c505476ed1db57f2f2f06a519a4/admin/class-aioseop-notices.php#L574

I thought I applied a similar check when working on the base module as well, but I guess I was thinking of something else.

michaeltorbert:
I agree with Ashish that if there is to be a check, it onlymakes sense to check for the core style, not our own. 
However, the check seems unneeded. It seems like something we should get rid of the check and then take care of if/when WP ever includes the style.

EkoJR:
That is true, it would make sense to check if there was a core style, but there isn't. There is actually a greater chance to be a conflict with a plugin that didn't follow the "prefix everything" guideline.

With the Notices class where there is a similar check, I used it to prevent processing more than once. 

In this case, I'm ok with removing the check. The 3.0 branch didn't have a check until this PR.

Issue #1852, opened 2018-08-17
Change/Consolidate upgrade to pro notice


EkoJR:
Related to #43 

This will also require the notices object/class #1850

PR #2292, opened 2019-03-13
 Change upgrade to Pro notice to AIOSEOP_Notices

EkoJR:
Issue #1852


arnaudbroes:
@EkoJR this notice is triggered even when WooCommerce isn't enabled or installed.

image

As for wording, I'd use something like this:

"We've detected you're using WooCommerce. Upgrade to All in One SEO Pack Pro to unlock our advanced eCommerce support, including SEO for Product Categories and more."

@michaeltorbert @wpsmort any thoughts?

wpsmort:
@arnaudbroes I recommend three changes:

- Change "We've" to "We have"
- Change "you're" to "you are"
- Change "eCommerce support" to "e-commerce features"

Issue #1854, opened 2018-08-17
Change/Consolidate Blog visibility w/ search engines notice


EkoJR:
Related to #43 

This will also require the notices object/class #1850

PR #2294, opened 2019-03-13
 Change Blog Search Engine Visibility Notice to AIOSEOP_Notices class

EkoJR:
Issue #1854

### DIFF
--- a/admin/functions-notice.php
+++ b/admin/functions-notice.php
@@ -85,4 +85,84 @@ if ( class_exists( 'AIOSEOP_Notices' ) ) {
 		global $aioseop_notices;
 		$aioseop_notices->deactivate_notice( 'woocommerce_detected' );
 	}
+
+	/**
+	 * Set Notice for Disabled Public Blog
+	 *
+	 * Admin Notice when "Discourage search engines from indexing this site" is
+	 * enabled in Settings > Reading.
+	 *
+	 * @since 2.4.2
+	 *
+	 * @global AIOSEOP_Notices $aioseop_notices
+	 *
+	 * @param boolean $update Updates the notice with new content and configurations.
+	 * @param boolean $reset  Notice are re-initiated.
+	 */
+	function aioseop_notice_set_blog_public_disabled( $update = false, $reset = false ) {
+		global $aioseop_notices;
+
+		$notice = aioseop_notice_blog_visibility();
+
+		if ( ! $aioseop_notices->insert_notice( $notice ) ) {
+			if ( $update ) {
+				$aioseop_notices->update_notice( $notice );
+			}
+			if ( $reset || ! isset( $aioseop_notices->active_notices[ $notice['slug'] ] ) ) {
+				$aioseop_notices->activate_notice( $notice['slug'] );
+			}
+		}
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


arnaudbroes:
This is good, but again the wording might need some tweaking. I'd also remove the button and link to options-reading.php in the text. 

More importantly, I'd make sure people know this notice is being displayed by All in One SEO Pack so they're aware we are providing them with this valuable insight.

image

Here's what I suggest:

Warning: All in One SEO Pack has detected that you're blocking access to search engines. You can change this in <a href="/wp-admin/">Reading Settings</a> if this is unintended.

Remind me later.

@michaeltorbert @wpsmort what do you think?

wpsmort:
@arnaudbroes Looks good, the only changes I'd make are:

- Change "you're" to "you are".  I believe contractions make translation for some languages more difficult (although I may be wrong about this)
- Change "Reading" to "Settings > Reading" which hopefully makes it clearer where they should navigate to

EkoJR:
@arnaudbroes @wpsmort Regarding "Settings > Reading", I noticed that is for WP versions under 3.5, and we're long past that. So we can just use what WP currently calls it, which is "Reading Settings"

arnaudbroes:
image
@EkoJR link is no longer clickable.

wpsmort:
@EkoJR My suggestion of Settings > Reading was in reference to how you reach that screen but going to Settings and then Reading.

EkoJR:
I believe I mentioned a limitations with notice messages, and one of them is the message itself can't have links, which is part of the reason why I made action_options to act as links as well.

So, I had to add the link back to the action_option where "Remind me later" is located, and it's changed to "Settings > Reading"

arnaudbroes:
@EkoJR I wasn't aware notices can't contain links. Can you quickly change those action links to buttons (as I think that will look a bit better)? Then this should be good to go.

admin/functions-notice.php
```diff
@@ -85,4 +85,84 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
 		$aioseop_notices->deactivate_notice( 'woocommerce_detected' );
 	}
+
+	/**
+	 * Set Notice for Disabled Public Blog
+	 *
+	 * Admin Notice when "Discourage search engines from indexing this site" is
+	 * enabled in Settings > Reading.
+	 *
+	 * @since 2.4.2
```
contactashish13: shouldn't this (and others below) be since 3.0?

admin/functions-notice.php
```diff
@@ -85,4 +85,84 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
+	}
+
+	/**
+	 * Notice - Blog Visibility
+	 *
+	 * Displays when blog disables search engines from indexing.
+	 *
+	 * @since 2.4.2
+	 *
+	 * @return array
```
contactashish13: a description of what it returns would help

admin/functions-notice.php
EkoJR: Yea it should. Changed

admin/functions-notice.php
EkoJR: Added

PR #2295, opened 2019-03-13
 Add review plugin notice to AIOSEOP Notices class

EkoJR:
Issue #61

### DIFF
--- a/admin/functions-notice.php
+++ b/admin/functions-notice.php
@@ -132,7 +132,7 @@ if ( class_exists( 'AIOSEOP_Notices' ) ) {
 			'target'         => 'user',
 			'screens'        => array(),
 			'class'          => 'notice-info',
-			'message'        => __( 'Looks like you\'ve been using All in One SEO Plugin for awhile now, and that\'s awesome! We are an open source community built from other\'s contributions. By helping us with a 5-star review, it also helps us to reach out to more people.', 'all-in-one-seo-pack' ),
+			'message'        => __( 'You have been using All in One SEO Plugin for awhile now. That is awesome! If you like our plugin please leave us a 5-star rating. Huge thanks in advance!', 'all-in-one-seo-pack' ),
 			'action_options' => array(
 				array(
 					'time'    => 0,
--- a/admin/functions-notice.php
+++ b/admin/functions-notice.php
@@ -132,7 +132,7 @@ if ( class_exists( 'AIOSEOP_Notices' ) ) {
 			'target'         => 'user',
 			'screens'        => array(),
 			'class'          => 'notice-info',
-			'message'        => __( 'You have been using All in One SEO Plugin for awhile now. That is awesome! If you like All in One SEO Pack please leave us a 5-star rating. Huge thanks in advance!', 'all-in-one-seo-pack' ),
+			'message'        => __( 'You have been using All in One SEO Pack for awhile now. That is awesome! If you like All in One SEO Pack, then please leave us a 5-star rating. Huge thanks in advance!', 'all-in-one-seo-pack' ),
 			'action_options' => array(
 				array(
 					'time'    => 0,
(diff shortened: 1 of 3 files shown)


arnaudbroes:
This is good, but the wording needs some tweaking. 

I'd recommend using something like this:

You've been using All in One SEO Pack for a while now. That's awesome! By giving us a 5-star review, you help us reach out to others.

Add a review.
Remind me later.
No, thanks.

What are your thoughts @michaeltorbert @wpsmort?

EkoJR:
I actually should have added the Needs Discussion label when I created this PR.

wpsmort:
@arnaudbroes I'd avoid contractions here are well, so change "You've" to "You have" and change "That's" to "That is"

arnaudbroes:
image

@EkoJR some last changes:

- "Plugin" needs to be "Pack"
- Change the third sentence to: "If you like All in One SEO Pack**, then** please leave us a 5-star rating review."
- Change the third action link to "No, thanks". No capital needed.

Don't forget to run Grunt again.

admin/functions-notice.php
```diff
@@ -85,4 +85,75 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
 	}
+
+	/**
+	 * Set Notice on Activation to Review Plugin
+	 *
+	 * A delayed notice that is set during activation, or initialization (old installs),
+	 * to later display a review/rate AIOSEOP plugin. Delay time: 12 days.
+	 * Delay "...give me a week." 5 days
+	 *
+	 * @since 2.4.2
```
contactashish13: should this be since 3.0?

admin/functions-notice.php
```diff
@@ -85,4 +85,75 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
+				// aioseop_footer_remove_review();
+				// aioseop_footer_set_review();
+			}
+		}
+	}
+
+	/**
+	 * Notice - Review Plugin
+	 *
+	 * @return array
```
contactashish13: a description of what it returns would help

admin/functions-notice.php
```diff
@@ -85,4 +85,75 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
+	 * @param boolean $reset  Notice are re-initiated.
+	 */
+	function aioseop_notice_set_activation_review_plugin( $update = false, $reset = false ) {
+		global $aioseop_notices;
+
+		// TODO Optimize - Create a callback function/method to store most of the configurations (Avoid Database concept).
+		// Dynamic variable could be stored in the database. Config functions could go into a config file/folder.
+		$notice = aioseop_notice_review_plugin();
+
+		if ( $aioseop_notices->insert_notice( $notice ) ) {
```
contactashish13: can't this be simplified to 

```
$aioseop_notices->insert_notice( $notice );
if ( $update ) {
...
}
```
as the first `if` is not really doing anything

admin/functions-notice.php
```diff
@@ -85,4 +85,75 @@ function aioseop_notice_disable_woocommerce_detected_on_nonpro() {
...
+	 * @return array
+	 */
+	function aioseop_notice_review_plugin() {
+		return array(
+			'slug'           => 'activation_review_plugin',
+			'delay_time'     => 1036800,
+			'target'         => 'user',
+			'screens'        => array(),
+			'class'          => 'notice-info',
+			'message'        => __( 'You have been using All in One SEO Pack for awhile now. That is awesome! If you like All in One SEO Pack, then please leave us a 5-star rating. Huge thanks in advance!', 'all-in-one-seo-pack' ),
```
contactashish13: change `awhile` to `a while`

admin/functions-notice.php
EkoJR: `$aioseop_notices->insert_notice( $notice )` returns a boolean, if it returns false, then it updates.

Insert and Update are 2 different things.

admin/functions-notice.php
EkoJR: Changed

admin/functions-notice.php
EkoJR: Added

admin/functions-notice.php
EkoJR: Changed

admin/functions-notice.php
contactashish13: yeah, I see. But if you are doing nothing when insert_notice returns true, maybe just simplify it to the below? 
`
if ( ! $aioseop_notices->insert_notice( $notice ) && $update ) {
 // do the update stuff
`

I just feel that commented out code is adding nothing of value but if you intend to put something in there, then your current code is fine.

EkoJR: Yea, I do intend to add a footer notice. I forgot this was the notice that was designed differently with an empty IF to use later. I left it there to preserve the alternate code design.

Issue #2296, opened 2019-03-14
twitter:title should be truncated


wpsmort:
Whilst discussing #808 with @contactashish13, we agreed that we should open a separate issue for truncating the twitter:title.

Twitter states that the title should be a maximum of 70 characters - https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup.html

This does not, and should not, affect the og:title meta tag as Facebook does not truncate the title. I confirmed this with a title that was over 100 characters long and Facebook displayed the entire title.

@contactashish13 As discussed, we should have a filter that enables users to disable truncation of the twitter:title meta tag.

Related to #2139

contactashish13:
@wpsmort as discussed PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2297 addresses this

arnaudbroes:
This also affects the Twitter:description which is truncated to 200 character (maximum). Just adding a note here because this was confusing me.

Issue #808, opened 2017-03-09
OG descriptions and twitter descriptions are too big


michaeltorbert:
With "Use Content For Autogenerated OG Descriptions" and "Autogenerate Meta Descriptions" unchecked, no character count filter is applied, and the length of the descriptions is too much. (This may also apply to other settings as well.)

I've verified the issue.

I believe this was introduced in 2.3.12, but we should confirm this.

Reported by Barry via email.

arnaudbroes:
On it.

michaeltorbert:
It looks like the OG desc and twitter desc are coming in at ~1000 characters max.
According to their docs, Facebook can do up to 300 and Twitter can do up to 200.

arnaudbroes:
@michaeltorbert 
Tested with about 8 versions, I can confirm this isn't a new.
.

contactashish13:
@michaeltorbert this seems to be a problem with other settings as well. While testing, I tried a long title and the entire title was printed out without any truncation. Do you want to create another issue to deal with this or should this done as part of the same issue?

michaeltorbert:
@contactashish13 Since that's a separate issue, it's best to go ahead and create a separate Github issue.

contactashish13:
### Facebook

I've been searching about this and it seems there are multiple schools of thought without a consensus:

1. https://stackoverflow.com/a/8916327 says provide whatever and it will be suitably truncated
1. The official site http://ogp.me/ says very vaguely "A one to two sentence description of your object" without a hard limit
1. https://stackoverflow.com/a/9552296 (same post) refers to a link that is dead but the content from there is what shows up when you google "og:description length"
1. https://stackoverflow.com/questions/8914476/facebook-open-graph-meta-tags-maximum-content-length#comment76342330_8916327 (same post) says there is a 10,000 character limit which seems to be something we don't cross.

I'm not sure we should be imposing any limits here.

### Twitter 
On the other hand, Twitter is explicit https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/markup.html

### Conclusion
I think we should create a separate issue for twitter meta tags limits and address them all in one issue. For Facebook, we should create a separate issue and conclude there, if we can.

suggestions @arnaudbroes @wpsmort @wpsmort?

arnaudbroes:
I suggest that we use the optimal OG lengths I brought up in #2139. Wait on what Michael and Steve have to say about this though.

[...]

PR #2297, opened 2019-03-15
OG descriptions and twitter descriptions are too big + twitter:title should be truncated

contactashish13:
as discussed with Steve on the call, I'm creating 1 PR for 2 issues as the code is 95% identical - #808 and #2296


wpsmort:
@contactashish13 I tested this and found two problems:

1. The og:description should be truncated to 200 characters not 55 characters as per my comment here https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/808#issuecomment-472449868

2. We should not be truncating an og:description, twitter:title or twitter:description when it is manually entered in the Social Settings tab of the Edit Post screen.  Only auto generated OG and Twitter meta should be truncated.

contactashish13:
@wpsmort if I don't have "Use Content For Autogenerated OG Descriptions" checked, is it considered auto generated? 

It would be great if you could mention the combination of settings that should result in a truncation.

wpsmort:
@contactashish13 You have the truncation working correctly for auto generated OG and Twitter meta.  There's no problem there (with the exception that og:description's should be truncated at 200 characters).  

The problem is that you're truncating the og:description, twitter:title or twitter:description that someone manually types into the Title and Description fields on the Social Settings tab in AIOSEOP on the Edit Post screen.  

We never truncate what someone types into these fields, if they want to type a long title and description then we let them.  We output exactly what they've typed as the meta tags.

Screen Shot 2019-03-16 at 9 36 04 AM

contactashish13:
@wpsmort I've addressed these issues.

wpsmort:
This is tested and is good.

modules/aioseop_opengraph.php
```diff
@@ -1,1832 +1,1874 @@
...
-		}
-
-		function save_tax_data( $term_id, $tt_id, $taxonomy ) {
-			static $update = false;
-			if ( $update ) {
-				return;
-			}
-			if ( $this->locations !== null ) {
-				foreach ( $this->locations as $k => $v ) {
-					if ( isset( $v['type'] ) && ( $v
```
EkoJR: Nice work updating the variables and description. Looks good.

The `since` tag might have been misplaced though. Normally it's suppose to be...

```
/**
 * Summary.
 *
 * @since x.x.x
 *
 * @param type $var Description.
 * @param type $var Optional. Description. Default.
 * @return type Description.
 */
```
https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods

EkoJR: This filter needs to have documentation.

https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

EkoJR: I believe all these will need to use aioseop's str* functions. Which those functions are intended to also be used with mb_str* functions. I believe it's in the base module, or functions.

EkoJR: Params were added, and public documentation may need to be updated.

This filter could also be php documented too.

contactashish13: done

contactashish13: done

contactashish13: good point!

contactashish13: done

EkoJR: Looks good.

@wpsmort is there public documentation?

EkoJR: Looks nice, thanks.

wpsmort: I've opened #2308 to add the API documentation for the new filter

michaeltorbert:
@contactashish13 The merge conflict needs to be resolved, then it needs to be tested and code reviewed again.

wpsmort:
@contactashish13 I found one more instance where the og:description, twitter:title and twitter:description are being truncated when it should not be.  This occurs when you have manually enter a title and description in the fields on the Main Settings tab on the Edit Post screen.  When an SEO title and description has been entered, we use that as the OG and Twitter title and description and we should not truncate this.  

Screen Shot 2019-03-21 at 10 02 38 AM

wpsmort:
This is tested and is good.

Issue #1848, opened 2018-08-16
Facebook SSL og:image only works if they have a non-SSL version cached


jimkeir:
**Describe the bug**
If the site is using entirely HTTPS, Facebook does not retrieve the og:image if they have not cached a non-HTTP version.

**To Reproduce**
1. Create an entirely new page that FB has never seen before, served on HTTPS only.
2. Set the SEO "og:image" tag to an https:// URL as appropriate for that page.
3. Use FB's sharing debugger (https://developers.facebook.com/tools/debug/sharing/) to retrieve the page.
4. Note that they say that the og:image "could not be retrieved".
5. Edit modules/aioseop_opengraph.php line 1362 to send "og:image:secure_url" instead of "og:image" .
6. Re-scrape in FB's sharing debugger.
7. Note that the correct image now appears, but with a warning that "og:image" should be specified even it can be derived.
8. Revert the change in modules/aioseop_opengraph.php and re-scrape.
9. Note that the correct image *still* shows, and that the warning has now gone.

**Expected behavior**
When the resulting URL for og:image is over SSL, send both og:image and og:image:secure_url tags.

**Version Info (please complete the following information):**
 - WordPress version: 4.9.8
 - All in One SEO Pack version: 2.7.3

arnaudbroes:
I can't say that the above steps are fully correct, but what I can say is that Facebook definitely isn't able to pick up the image when only using SSL and that online documentation shows we need to add the `og:image:secure_url` tag when the site supports https.

arnaudbroes:
@contactashish13 I assigned this one to you for 3.0 as it shouldn't be too difficult to solve. Detect if the site is using HTTPS and if so, output the `og:image:secure_url` tag in addition to the regular `og:image` tag.

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2299

PR #2299, opened 2019-03-18
Facebook SSL og:image only works if they have a non-SSL version cached

contactashish13:
#1848


contactashish13:
@arnaudbroes I cannot reproduce this. Do you want to check again?

image

arnaudbroes:
@contactashish13 I might have tested wrong, shows up now, thanks!

Issue #2305, opened 2019-03-20
Make it easy to test PRO features in the free plugin with minimal changes


contactashish13:
I thought about this change while working on https://github.com/semperfiwebdesign/aioseop-pro/issues/34. 

The traditional way of testing PRO features are:

**Method 1**
- make changes in the PRO plugin code
- once everything works, migrate the code to the equivalent FREE files
- be careful **not to forget** anything

**Method 2**
- copy the PRO headers to the FREE plugin
- change value of `AIOSEOPPRO` to true
- copy `pro` folder to the free plugin
- test

**Method 3**
- change value of `AIOSEOPPRO` to true
- copy `pro` folder to the free plugin
- test

The problems with method 1 are the below:
- forgetting the files one worked on (because they are in the pro plugin folder during development)

The problems with method 2 are the below:
- the headers might be checked in accidentally

The problems with method 3 are the below:
- does not work because the free plugin is automatically deactivated 

I discovered that instead of the above, we can instead make subtle quantifiable changes to the FREE plugin so that it behaves like the PRO plugin.

This issue and its PR deal with those small changes that will help in development till both plugins are merged.

contactashish13:
PR https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2306

comments welcome @semperfiwebdesign/aioseop-core-committers 

The steps to test PRO features using the FREE plugin are:
- change value of `AIOSEOPPRO` to true
- change value of `AIOSEOPPRO_MIMIC` to true
- copy `pro` folder to the free plugin
- test

The advantages I see are:
- 2 switches that need to be switched on
- no copying the header
- no forgetting files in some other folder
- no inadvertent commits or forgetting of commits

michaeltorbert:
Even better is to use action/filter hooks so that Pro functionality doesn't require changes in free.

PR #2306, opened 2019-03-20
Make it easy to test PRO features in the free plugin with minimal changes

contactashish13:
#2305

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ ehthumbs.db
 #############
 .htaccess
 
+pro/
--- a/all_in_one_seo_pack.php
+++ b/all_in_one_seo_pack.php
@@ -38,6 +38,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 if ( ! defined( 'AIOSEOPPRO' ) ) {
 	define( 'AIOSEOPPRO', false );
 }
+// To develop/test PRO features while using the FREE plugin...
+// 1. change the value of this and AIOSEOPPRO above to true
+// 2. copy the pro folder here
+// 3. test
+if ( ! defined( 'AIOSEOPPRO_MIMIC' ) ) {
+	define( 'AIOSEOPPRO_MIMIC', false );
+}
 if ( ! defined( 'AIOSEOP_VERSION' ) ) {
 	define( 'AIOSEOP_VERSION', '2.12' );
 }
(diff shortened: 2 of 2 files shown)


michaeltorbert:
What is the benefit of this over just having `define( 'AIOSEOPPRO', true );` in wp-config.php?

contactashish13:
As I mentioned on the call, the free plugin deactivates itself when it sees pro is true. The mimic constant prevents that from happening.

arnaudbroes:
@contactashish13 does this need discussion or testing now? There's no stage label.

contactashish13:
@arnaudbroes Actually, can you test the below 2 ways just to verify that you see the same behavior that I do? 

**Old method**
1. Make `define( 'AIOSEOPPRO', true );`
2. Copy pro folder
3. Test
(here the free plugin gets deactivated automatically after 1-2 refreshes)

**New method**
1. Make `define( 'AIOSEOPPRO_MIMIC', true );`
2. Copy pro folder
3. Test

arnaudbroes:
@contactashish13 that isn't a full solution since there are a number of checks in the plugin for the AIOSEOPPRO variable. For instance, AIOSEOPPRO_MIMIC doesn't disable the upgrade links throughout the plugin nor does it activate Pro features like the Video XML Sitemap. 

What's the advantage for us of having both the free and Pro version activated? Also, I don't think this variable would allow that as the folder name also needs to be different.

michaeltorbert:
Yeah I must admit I'm not sure I see the benefit either. Creating Pro is extremely easy. You just drop in the Pro directory and change the constant + plugin name.

arnaudbroes:
@contactashish13 feel free to reopen this if you can think of a good reason why we need to make this "easier" (as it is already extremely easy like Michael said) but in the meanwhile I'm going to close this PR.

contactashish13:
No offence but I feel like no one's actually tried this :) I have tried the "easy" method and it deactivates the plugin. It makes it impossible, not merely difficult, to test a change. I have tried to list down the different scenarios and out of them the easiest is this. Nevertheless, if you guys feel differently, you can keep this closed.

michaeltorbert:
@contactashish13 It won't deactivate if you change the directory name, for instance appending "-pro".  The deactivation code specifically looks for a plugin directory named "all-in-one-seo-pack". https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/all_in_one_seo_pack.php#L510

contactashish13:
Ok, so that doesn't make it the easiest method then. Because then development is not seamless. I have to change the folder name, copy the headers - don't you think from a development point of view that is too cumbersome compared to what I'm suggesting? The likelihood of checking in the main file with wrong headers is also high.

michaeltorbert:
They're both about the same number of steps, but one doesn't require permanent changes to the code.

contactashish13:
Let me list it down:

1. Copy headers
2. Change define
3. Change folder name
4. Make/test changes
5. Revert folder name so that commits can happen
6. Change define
7. Commit


Suggested method:

1. Change define
2. Make/test changes
3. Change define
4. Commit

PR #2321, opened 2019-03-28
XML Sitemap - Can't exclude custom taxonomies

contactashish13:


contactashish13:
@wpsmort this is ready for testing. I'm still working on the unit test cases.

I've tested this for the following scenarios:

1. Using previous version of the plugin, select some category terms to exclude.
2. Apply PR
3. Settings now show the excluded terms as selected in step 1
4. Save the settings and load the sitemap. The sitemap does not include the posts corresponding to the terms.
5. Add all terms from a selected taxonomy.
6. Indexed sitemap does not contain the taxonomy sitemap. The post sitemap does not contain the posts corresponding to these terms.
7. Simple sitemap does not contain the posts corresponding to these terms. It also does not contain the links to the terms themselves.

Things to note:
- Terms are searched using a `LIKE %%` clause. Thus searching for `s` will find `search` as well as 'fester`.
- Search is triggered with a single character input.
- Terms are only searched from within the selected taxonomies
- If a search text is not found, the dropdown results will not display anything

wpsmort:
@contactashish13 I tested this by following the steps in your comment above and when I apply the PR code, any category I excluded in step 1 was no longer excluded.  I tested on a new site as well as an existing site.   The category I excluded didn't appear in the Excluded Terms field and the category and its posts appear in the sitemap.

I also get this PHP notice:

Undefined index: aiosp_sitemap_excl_terms in all-in-one-seo-pack-pro-240/modules/aioseop_sitemap.php on line 561

contactashish13:
@wpsmort as mentioned on slack, please provide more information such as test site details.

wpsmort:
@contactashish13 I've tested on a brand new site by creating a new test post and assigning it to a new category.  I then excluded this category from the sitemap by using the checkboxes in the Excluded Items section (as per step 1 of your testing instructions).  

I then installed and activated the PR code and refreshed the sitemap.  The category I had excluded was now showing up in the sitemap and when I view the new Excluded Terms field it is empty.

If I look in the database, I see the ID for the category is present under aiosp_sitemap_excl_categories.

Here's the site I'm testing with - https://wordpress-138706-813687.cloudwaysapps.com/wp-admin/

I'll send the login to you via Slack.

contactashish13:
@wpsmort I will check this issue. In case there is a confusion, there are 2 **independent** test scenarios for this PR and are not dependent on each other to function so they can be tested independently:

1. old users upgrading from prev version (which is the issue you have reported)
2. new users on a fresh site (steps 5-7 that I have mentioned but have not been reported by you as working or not working)

contactashish13:
@wpsmort I was able to reproduce the issue on your test site and fix it.

wpsmort:
I've updated the issue with feedback from my testing on a large site.

arnaudbroes:
Quoting Steve here:

> I tested PR semperfiwebdesign/all-in-one-seo-pack#2321 and found one problem. On a site with a lot of taxonomy terms, when you type in the name or a term, the list of matching terms refreshes and you lose what you started typing. I've taken a screen recording of this which shows it happening. You can view the screen recording here - https://www.dropbox.com/s/q5xz8yk25rvxzui/Testing%20PR%202321.mov?dl=0

This is caused by `$(".aioseop-chosen").trigger("chosen:updated");` on line 51 where it is triggering some kind of handler that clears the user input in the input field after the database query returns results. I'm assuming the idea behind this was that the search term can be cleared after the lists with results has been populated, however, this causes two issues: 

(1) it looks like it starts querying immediately after you enter a single character. This is a problem because if you continue entering characters, it will clear whatever you've already entered into the search input field once the results for the initial query are returned. This can be solved by adding a timeout to the function that triggers the query so that the database is only queried once the user has actually stopped typing. This way, we also avoid unnecessary queries and improve performance.

(2) users might want to change their search term or "refine" the results even further by entering more characters after results from the first query are returned. By clearing the search input field, we would be forcing them to start over again, which is counter-productive. 

In short, the following has to happen:

1. Add timeout to function that triggers database query so that we only query the database once the user has finished typing;
2. Remove `$(".aioseop-chosen").trigger("chosen:updated");` both on line 51 and line 45. If a query doesn't return any results, then the user should be able to adjust the search term instead of wiping it.

@contactashish13 let me know if that makes sense and if you have any questions. Apart from that, you've done really good work on this. This is exactly how we wanted this to be implemented.

contactashish13:
@arnaudbroes  I don't agree with your solutions because

1) the user can keep refining the search as you say. This means "stopping" is not defined. The search acts on the whole string the user has typed not incremental. That means if you type "hel" and then add "lo"  to make it "hello" the search looks for hello not results with "lo" in the "hel" results.

2) `$(".aioseop-chosen").trigger("chosen:updated");` is important

arnaudbroes:
@contactashish13 there doesn't have to be a character threshold per se.

1. If you would have looked at the video from Steve, you would have noticed that the code in its current state doesn't allow you to refine your search or even finish your search, as it starts querying right away and clears your input when results are returned. Adding a character threshold won't fix that; it only moves the issue from a single to 1+ characters. 

2. I don't see why that specific line of code is needed anywhere. I suggest that you try this for yourself, but if you comment out lines 41, 45 and 51 through the browser's dev tools, then the code works just fine (and this solves (1) as well).

contactashish13:
@arnaudbroes If you would have looked through the official documentation you would have understood the need for that piece of code. Please search for "Updating Chosen Dynamically" in https://harvesthq.github.io/chosen/. 

Also, if you have access to the code, it might be a better idea to change the actual JS file instead of through dev tools to actually reproduce what I see as the behavior. Line 41 is a comment so I'm not sure what you are referring to.

PS - It's good that you are looking at the code from a developer's point of view but it would save us time if you changed the code yourself and tested it instead of us going back and forth on it.

EkoJR:
@contactashish13 It seems like "Observing Form Field Changes" might be the function you are looking for to update the data, and the "Updating Chosen Dynamically" might be for updating the element's options instead.

arnaudbroes:
@contactashish13 we discussed this PR during today's meeting and we believe you're right that those lines can't be removed; however, that still leaves us with the initial issue (see Steve's video) where the field gets cleared whilst typing. Can you see if you can find a solution for this problem and implement it? 

Ben (see above) might be on the right track. Ideally, there's also a timeout so we don't start querying right away but only after the user finishes typing.

contactashish13:
@EkoJR I'm using the correct event. 

image

Let me explain what is at play here. There are 2 components: jQuery Chosen and jQueryUI Autocomplete (http://api.jqueryui.com/autocomplete/). Chosen is responsible only for showing the drop down as a fancy text box. Autocomplete is the one responsible for searching. Once the search fires and results are obtained, chosen has to be updated. That is why that is the correct function.

The problem that we see is because 
- the user keeps on typing while the search is in progress
- the search fires immediately after the user's keystroke
- the search starts with even 1 character

The problem therefore is with Autocomplete and not with Chosen. The solution, I think, is to add a sensible combination of http://api.jqueryui.com/autocomplete/#option-delay and http://api.jqueryui.com/autocomplete/#option-minLength. I've used 500 and 3 respectively viz. the search will be fired 500ms after the user's last keystroke and after at least 3 characters have been inputted.

@wpsmort I have tested by adding some 500+ categories to my dev instance. Can you please test?

wpsmort:
@contactashish13 To confirm our phone call today, here are the problems I experienced when testing the latest code in this PR:

1.  After typing the name of a tax term and selecting the matching term, if I them try to type a new term, the list populates with terms that match my previous query.  For example, I type cat 1 and select the matching category.  I then type tag 1 but the list shows terms matching cat 1 still.

2.  After the problem above happens a few times, I can no longer type in the field.  It won't let me type until I select a term from the list.

3.  If I then try to type a new term such as drama, the list shows the matching term after I type the first three letters, but then the typed entry disappears and the list returns to displaying terms matching cat 1 and tag 1.

I uploaded a video of this here - https://www.dropbox.com/s/cxrvc7l9spkmwjn/PR%202321.mov?dl=0

I have copied my large test site to Cloudways and sent you the logins via Slack.

contactashish13:
@wpsmort I've addressed these issues. The new matching terms will now be added to the top of the list rather than at the bottom.

wpsmort:
@contactashish13 I have tested this and the field behavior is good.  There is one problem which is, when you exclude a term and there are no other terms in the taxonomy then an empty sitemap is created when sitemap indexes is enabled.  

Steps to reproduce this:
1. Create a taxonomy with one term and add a post to that term
2. Enable sitemap indexes and verify that a sitemap is created for the taxonomy in the sitemap index and that the sitemap contains the one post
3. Now exclude that term from the sitemap
4. View the sitemap index and the sitemap for the taxonomy should still be present and when you view it you will get a 404 page not found error

Let me know if you have any questions.

wpsmort:
This is tested and is good.  I've tested on a large site with a complex combination of taxonomies and terms and the new field worked well.

admin/aioseop_module_class.php
```diff
@@ -821,11 +821,14 @@ function get_object_labels( $post_objs ) {
 		 *
 		 * @return array
 		 */
-		function get_term_labels( $post_objs ) {
+		function get_term_labels( $post_objs, $prefixed_with_taxonomy = false ) {
```
EkoJR: PHPDoc needs to be updated, and noting a change in `since`.

admin/aioseop_module_class.php
```diff
@@ -855,8 +858,8 @@ function get_taxonomy_titles( $args = array() ) {
 		 *
 		 * @return array
 		 */
-		function get_category_titles( $args = array() ) {
-			return $this->get_term_labels( get_categories( $args ) );
+		function get_category_titles( $args = array(), $prefixed_with_taxonomy = false ) {
```
EkoJR: PHPDoc needs to be updated, and noting a change in `since`.

css/aiosp_admin.css
```diff
@@ -132,3 +132,8 @@ li#wp-admin-bar-aioseop-pro-upgrade a.ab-item {
 	color: #d54e21;
 	font-size: 105%;
 }
+
+/* ensure the chosen auto-complete is visible */
+#aiosp_sitemap_excl_terms_wrapper .aioseop_option_div {
+	overflow-y: visible;
+}
```
EkoJR: Might be good measure to add a new line at end of file with this PR

js/modules/sitemap.js
```diff
@@ -0,0 +1,76 @@
+/* global sitemap */
```
EkoJR: This will need documentation for the file header. https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#file-headers

Might also help to link to JS chosen somewhere (either file header of function).

js/modules/sitemap.js
```diff
@@ -0,0 +1,76 @@
...
+
+(function($, sitemap){
+    
+    $(document).ready(function(){
+        initAll();
+    });
+
+    function initAll(){
```
EkoJR: Function(s) will need documentation https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/javascript/#functions

js/modules/sitemap.js
```diff
@@ -0,0 +1,76 @@
...
+                        }));
+                        $(".aioseop-chosen").trigger("chosen:updated");
+                        $(".aioseop-chosen").enableTextBox();
+                    }
+                });
+            }
+        });
+    }
+
+})(jQuery, sitemap);
```
EkoJR: Missing new line at end of file.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
 			// TODO is this required for dynamic sitemap?
 			add_action( 'transition_post_status', array( $this, 'update_sitemap_from_posts' ), 10, 3 );
 			add_action( 'after_doing_aioseop_updates', array( $this, 'scan_sitemaps' ) );
+			add_action( 'after_doing_aioseop_updates', array( $this, 'upgrade_excluded_categories_to_excluded_terms' ) );
 			add_action( 'all_admin_notices', array( $this, 'sitemap_notices' ) );
+			add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue_scripts' ), 20 );
```
EkoJR: Unsure if you are aware, but adding this hook is usually already handled by the parent class; when using the same function/method name.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+			add_action( "wp_ajax_{$this->prefix}", array( $this, 'ajax' ) );
+		}
+
+		/**
+		 * Migrate the values in "Excluded Categories" to "Excluded Terms".
+		 *
+		 * @since 3.0
+		 */
+		function upgrade_excluded_categories_to_excluded_terms() {
```
EkoJR: Is this a database update? If so, this would probably be better placed in the AIOSEOP's updater.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+			}
+		}
+
+
+		/**
+		 * The single entry function to handle all ajax stuff.
+		 *
+		 * @since 3.0
+		 */
+		function ajax() {
```
EkoJR: This is a really broad function name. Wouldn't it be better for a more defined name for its purpose? Like, `ajax_terms_search`.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+
+					$all_terms = array();
+
+					foreach ( $taxonomies as $taxonomy ) {
+						$terms = get_terms(
+							array(
+								'taxonomy' => $taxonomy,
+								'fields' => 'id=>name',
+								'hide_empty' => false,
+								'name__like' => $_POST['term'],
```
EkoJR: Curious, did you try using `search` at all?

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+		}
+
+		/**
+		 * Load resources.
+		 *
+		 * @since 3.0
+		 *
+		 * @param string $hook_suffix The current admin page.
+		 */
+		function admin_enqueue_scripts( $hook_suffix ) {
```
EkoJR: This would either need to be renamed, or call the parent method; this would break the Design Pattern when working in conjunction with the parent class.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+					'nonce' => wp_create_nonce( $this->prefix ),
+				),
+			);
+			if ( strpos( $hook_suffix, '/modules/aioseop_sitemap' ) !== false ) {
+				wp_register_script( 'aioseop-chosen', AIOSEOP_PLUGIN_URL . 'js/lib/chosen.jquery.min.js', array( 'jquery', 'jquery-ui-autocomplete' ), AIOSEOP_VERSION );
+				wp_register_style( 'aioseop-chosen', AIOSEOP_PLUGIN_URL . 'css/lib/chosen.min.css', array(), AIOSEOP_VERSION );
+				wp_enqueue_script( 'aioseop-chosen' );
+				wp_enqueue_style( 'aioseop-chosen' );
+				$deps[] = 'aioseop-chosen';
+				$options['l10n'] = array(
```
EkoJR: Wouldn't it be more optimal & secure for clients/browsers to handle only the data that is needed, rather than sending/localizing the whole database options? If it is just translated text & a couple other values, then it would probably be better to just use those for localizing.

modules/aioseop_sitemap.php
```diff
@@ -403,7 +446,175 @@ public function __construct() {
...
+					'choose_terms' => __( 'Choose specific terms to exclude.', 'all-in-one-seo-pack' ),
+					'term_not_found' => __( 'Term not found', 'all-in-one-seo-pack' ) . ':',
+					'term_searching' => __( 'Searching for', 'all-in-one-seo-pack' ) . ':',
+				);
+				wp_enqueue_script( 'aioseop-sitemap', AIOSEOP_PLUGIN_URL . 'js/modules/sitemap.js', $deps, AIOSEOP_VERSION );
+				wp_localize_script( 'aioseop-sitemap', 'sitemap', $options );
```
EkoJR: It would be a good idea to prefix the localized variable to prevent any potential conflicts.

modules/aioseop_sitemap.php
```diff
@@ -3849,10 +4068,10 @@ public function get_tax_args( $taxonomy, $page = 0 ) {
 		 *
 		 * @return mixed
 		 */
-		public function set_post_args( $args ) {
-			if ( $this->option_isset( 'excl_categories' ) ) {
+		public function set_post_args( $args, $only_tax = false ) {
```
EkoJR: PHPDoc will need to be updated, and a `since` added noting the change.

modules/aioseop_sitemap.php
```diff
@@ -4249,6 +4520,8 @@ public function get_all_post_type_data( $args ) {
 			}
 			$this->excludes = array_merge( $args['exclude'], $exclude_slugs ); // Add excluded slugs and IDs to class var.
 
+			$args = $this->add_tax_query( $args );
```
EkoJR: DEV NOTE: Potential conflict with #2317 Same function is being modified.

admin/aioseop_module_class.php
contactashish13: done

admin/aioseop_module_class.php
contactashish13: done

css/aiosp_admin.css
contactashish13: done

js/modules/sitemap.js
contactashish13: done

js/modules/sitemap.js
contactashish13: done

js/modules/sitemap.js
contactashish13: done

modules/aioseop_sitemap.php
contactashish13: have renamed the function

modules/aioseop_sitemap.php
contactashish13: it is a database update but it is very specific to this module so I feel better having it in the module than in a generic place.

modules/aioseop_sitemap.php
contactashish13: it is a single entry function for all ajax requests for this module as the comment indicates.

modules/aioseop_sitemap.php
contactashish13: no but it seems to be doing the same thing

modules/aioseop_sitemap.php
contactashish13: have renamed the function

modules/aioseop_sitemap.php
contactashish13: I'm not sure what you mean. `$options` is a different variable here

modules/aioseop_sitemap.php
contactashish13: I don't see any potential for conflicts. Plus the variable is available only in the IIFE context of the function so conflicts are not likely.

modules/aioseop_sitemap.php
contactashish13: done

modules/aioseop_sitemap.php
EkoJR: I was mistaken. I thought I saw an `$aioseop_options` variable, and the variable name threw me off.

modules/aioseop_sitemap.php
EkoJR: I tested operations with the parent method, and this can be removed. This works with the parent class admin_enqueue_scripts as intended.

I've uploaded a commit showing the changes https://github.com/EkoJR/all-in-one-seo-pack/commit/998cd74b2a6ba25dfdf45b63f4401c8ec61d8019

modules/aioseop_sitemap.php
EkoJR: @michaeltorbert We put all update operations in the updater class, correct?

modules/aioseop_sitemap.php
EkoJR: That is not good idea. If more AJAX is added, then this will have to be renamed, or repurposed, but as time progresses, sticking all ajax functionality in 1 function only increases its complexity, and unnecessary processes with unrelated operations. It also helps to use different nonces

modules/aioseop_sitemap.php
EkoJR: You didn't test the comparison? The reason why I mention it, is because I believe the LIKE query is heavy on the database, and using `search` might be using a different SQL string.

modules/aioseop_sitemap.php
EkoJR: I just tested this in conjunction with the parent method, and it works as intended. It would be better to use that to keep everything tied together rather than invent an alternate method.

I've uploaded a commit showing the changes https://github.com/EkoJR/all-in-one-seo-pack/commit/998cd74b2a6ba25dfdf45b63f4401c8ec61d8019

modules/aioseop_sitemap.php
EkoJR: Anytime you use variable that other plugins can use, you should always prefix your code. If you are unsure, prefix your code. It's a WP standard for good reason.

contactashish13: Other plugins can't use the variable. Its scope is only in the function - that's what IIFEs are. If you can provide me practical example, I'll change it otherwise it doesn't make sense to do it.

modules/aioseop_sitemap.php
contactashish13: not relevant now as the function is not being modified.

modules/aioseop_sitemap.php
contactashish13: I think its alright this way without violating any conventions. Interestingly, this has been used in many other classes. I understand you would have done it that way instead of this way but that's fair.

modules/aioseop_sitemap.php
EkoJR: > The name of the variable which will contain the data. **Note that this should be unique to both the script and to the plugin or theme.** Thus, the value here **should be properly prefixed** with the slug or another unique value, to prevent conflicts. However, as this is a JavaScript object name, it cannot contain dashes. Use underscores or camelCasing.

https://codex.wordpress.org/Function_Reference/wp_localize_script

modules/aioseop_sitemap.php
EkoJR: There's no reason to create an alternate function like this. I tested and it does work with the parent as intended, and also follows the encapsulation pattern.

modules/aioseop_sitemap.php
EkoJR: It even says for the $name param.

> The name of the variable which will contain the data. **Note that this should be unique to both the script and to the plugin or theme.** Thus, the value here **should be properly prefixed** with the slug or another unique value, to prevent conflicts. However, as this is a JavaScript object name, it cannot contain dashes. Use underscores or camelCasing.

https://codex.wordpress.org/Function_Reference/wp_localize_script

EkoJR:
Closing this in favor of https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2379

@contactashish13 I noticed there was a number of other issues I hadn't noticed earlier.

* This breaks the Help Text by adding old code back in; we should be using the new helper class for the info boxes.
* This should of also been using & calling the parent's enqueue function through inheritance.
* Did not need AJAX, nor did it need to be queried by name_like, but I assume that's because of ChosenJS.
* This should of used the Updater class to push an update.

Issue #2324, opened 2019-04-01
Sync AIOSEOP_Helper class with Pro


EkoJR:
Some changes were added to pro to make the Helper class more extended.

Related to https://github.com/semperfiwebdesign/aioseop-pro/pull/478

PR #2325, opened 2019-04-01
Sync with Pro & Add Filter for extended help text

EkoJR:
Sync AIOSEOP_Helper class with Pro #2324

This updates the Helper class to extend the help-text/tooltips for Pro.

PR to test https://github.com/semperfiwebdesign/aioseop-pro/pull/478

## Filter Added

Used to modify the help text. Intended for internal purposes with Pro.

Tag: `aioseop_helper_set_help_text`
Param1: (array) `$help_text` - Contains an array of help text for each setting.
Param2: (string) `$module`  - Shows which class module is using the function.

### DIFF
--- a/admin/class-aioseop-helper.php
+++ b/admin/class-aioseop-helper.php
@@ -46,7 +46,7 @@ class AIOSEOP_Helper {
 	 * no class name in $module, then this Help Text will add all module help texts.
 	 *
 	 * @ignore
-	 * @since 2.4.2
+	 * @since 3.0
 	 * @access private
 	 *
 	 * @param string $module All_in_One_SEO_Pack module.
@@ -79,21 +79,17 @@ class AIOSEOP_Helper {
 			case 'All_in_One_SEO_Pack_Bad_Robots':
 				$this->help_text = $this->help_text_bad_robots();
 				break;
-			default:
-				$this->help_text = array_merge(
-					$this->help_text,
-					$this->help_text_general(),
-					$this->help_text_performance(),
-					$this->help_text_sitemap(),
-					$this->help_text_opengraph(),
-					$this->help_text_robots_generator(),
-					$this->help_text_file_editor(),
-					$this->help_text_importer_exporter(),
-					$this->help_text_bad_robots(),
-					$this->help_text_post_meta()
-				);
-				break;
 		}
+
+		/**
+		 * Set Help Text
+		 *
+		 * @since 3.0
+		 *
+		 * @param array  $this->help_text Contains an array of help text for each setting.
+		 * @param string $module          Shows which class module is using the function.
+		 */
+		$this->help_text = apply_filters( 'aioseop_helper_set_help_text', $this->help_text, $module );
 
 		if ( AIOSEOPPRO ) {
 			$help_text = aioseop_add_pro_help( $this->help_text );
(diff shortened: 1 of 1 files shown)


arnaudbroes:
@EkoJR we'll need some instructions to test this.

EkoJR:
@arnaudbroes added a redirect to https://github.com/semperfiwebdesign/aioseop-pro/pull/478 since this is just to sync with that PR.

wpsmort:
I believe this should be code reviewed at the same time as https://github.com/semperfiwebdesign/aioseop-pro/pull/478

EkoJR:
Changed title since the modifications isn't isolated to just a filter, but also includes edits that occured in Pro.

admin/class-aioseop-helper.php
```diff
@@ -79,31 +79,17 @@ private function _set_help_text( $module ) {
...
 
+		/**
+		 * Set Help Text
+		 *
+		 * @since 3.0
+		 *
+		 * @param array  $this->help_text Contains an array of help text for each setting.
+		 * @param string $module          Shows which class module is using the function.
+		 */
+		$this->help_text = apply_filters( 'aioseop_helper_set_help_text', $this->help_text, $module );
```
contactashish13: instead of https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/3.0/admin/class-aioseop-helper.php#L57-L81 it would be more elegant to use the filter and let each module populate its own help texts.  You are already passing the `$module` in the filter and this can be used to invoke a common method on each module that will populate the text.

This would ensure that any changes to a help text in a particular module only affect that module file and not this file as well, thereby preventing potential merge conflicts. Also, if the module defines its own fields I don't think some other file should define the help text corresponding to those fields unless there was a very strong reason to shift all the help text outside. Currently this behaves like a procedural function instead of OOP.

EkoJR: I'm not so sure I follow...the filter still has to check ([pro code](https://github.com/semperfiwebdesign/aioseop-pro/blob/d8ccb7bf9e66c0fda675b69e685813b53110468e/pro/admin/aioseop-helper-filters.php#L8)), and is mirrored in the same way to limit which module help text is loaded. The filter is mostly meant for extended purposes to replace what would of been.

```
if ( AIOSEOPPRO ) {
	$help_text = aioseop_add_pro_help( $this->help_text );
	// TODO - Get Prefix from PRO Version.
	foreach ( $help_text as $key => $text ) {
		$this->help_text[ $key ] = $text;
	}
}
```

Plus, I believe calling the method directly would make it more simple. You can go straight to the method/function rather than having to search through the hooks to get to the function. Those methods are also made to be private functions as well.

I can see it making more sense if the hook had a suffix, and might help for extending purposes. Within the same object seems a bit much, and require a lot more changes when I'm trying to keep the changes minimal in this PR in relation to the [Pro PR](https://github.com/semperfiwebdesign/aioseop-pro/pull/478).

If anything, security is more of a concern with that filter (since it renders in the admin area), but I already sanitize in a public function get_help_text.

contactashish13: I've created a new issue https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2329 to discuss this further.

PR #2334, opened 2019-04-13
Video XML Sitemap - Remove taxonomy pages from video sitemap

contactashish13:
https://github.com/semperfiwebdesign/aioseop-pro/issues/246

This should be tested in conjunction with https://github.com/semperfiwebdesign/aioseop-pro/pull/474


wpsmort:
Whilst testing this I found a bug that causes a 404 error when viewing the Video XML Sitemap, see issue #2339.  That bug needs to be fixed before I can test this PR.

wpsmort:
See latest comments here - https://github.com/semperfiwebdesign/aioseop-pro/issues/246#issuecomment-484209544

arnaudbroes:
This is looking good now. Both issues have been fixed and these fixes don't affect other features.

modules/aioseop_sitemap.php
```diff
@@ -1875,7 +1875,7 @@ public function get_sitemap_index_filenames() {
 				$options[ "{$this->prefix}taxonomies" ] = array();
 			}
 			$options[ "{$this->prefix}posttypes" ]  = array_diff( $options[ "{$this->prefix}posttypes" ], array( 'all' ) );
-			$options[ "{$this->prefix}taxonomies" ] = array_diff( $options[ "{$this->prefix}taxonomies" ], array( 'all' ) );
+			$options[ "{$this->prefix}taxonomies" ] = apply_filters( "{$this->prefix}show_taxonomy", array_diff( $options[ "{$this->prefix}taxonomies" ], array( 'all' ) ) );
```
EkoJR: Filter is missing documentation; there are other implementations of this filter missing docs.

Instead of having multiple apply_filters of the same filter, would it be better to reduce it to calling a function to apply the filter?

modules/aioseop_sitemap.php
```diff
@@ -3822,7 +3822,7 @@ public function parse_content_for_images( $content, &$images ) {
 		 *
 		 * @return array
 		 */
-		public function get_tax_args( $page = 0 ) {
+		public function get_tax_args( $taxonomy, $page = 0 ) {
```
EkoJR: Documentation needs to updated, and the `since` tag could also reflect a change to the param(s).

Issue #2282, opened 2019-03-12
Upgrade to Pro link has hardcoded path


michaeltorbert:
if the directory is anything other than all-in-one-seo-pack these don't show up.

<img width="452" alt="Screen Shot 2019-03-11 at 8 41 14 PM" src="https://user-images.githubusercontent.com/17364998/54167001-2f13d200-443e-11e9-8e0d-b8882c34a393.png">

michaeltorbert:
<img width="382" alt="Screen Shot 2019-03-11 at 8 51 55 PM" src="https://user-images.githubusercontent.com/17364998/54167336-98481500-443f-11e9-814a-ac3792d63317.png">

michaeltorbert:
The shield is in https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/css/aiosp_admin.css#L5 For example, the all-in-one-seo-pack in #toplevel_page_all-in-one-seo-pack-aioseop_class is a hardcoded URL.

The upgrade link is https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/admin/display/menu.php#L41 all-in-one-seo-pack/aioseop_class.php is the hardcoded path.

onetarek:
Changing the plugin folder name is not only causing the missing icon but it also changes all menu slugs. 
For example : I changed the folder name to `all-in-one-seo-pack2`. Now the icon is missing and menu slugs are changed to 
- `all-in-one-seo-pack2/aioseop_class.php`
- `all-in-one-seo-pack2/modules/aioseop_performance.php`
- etc.

## Cause of issue :
All menu slugs are defined based on plugin_basename( __FILE__ )
example : https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L4710

## Possible solution : 
 Make new function called `aioseop_menu_slug` that will take `__FILE__` as parameter.
Process of this function : 
- calculate $menu_slug =  plugin_basename( __FILE__ ) 
- calculate $plugin_folder_name = //calculate some how
- Replace $plugin_folder_name with `all-in-one-seo-pack` from $menu_slug
- return $menu_slug

Use this aioseop_menu_slug( __FILE__ ) function in all places where you define menu slug.

@michaeltorbert 
If you think this solution is perfect then I could send a pull request.

michaeltorbert:
@onetarek No, we're not going to do that, and please stop posting irrelevant comments on issues.
[...]

PR #2350, opened 2019-04-25
Remove hardcoded path for Upgrade link in admin menu #2282

arnaudbroes:
#2282.


wpsmort:
This is tested and is good.

Issue #2088, opened 2018-12-05
Use double quotes for pagination links


wpsmort:
Requested here - https://wordpress.org/support/topic/single-quotes-on-prev-next-links-in-archives/

We currently output the link rel=prev and link rel=next with single quotes.  We should consider changing this to double quotes like so:

```
<link rel="prev" href="https://www.example.com/category/recipes/page/2/" />
<link rel="next" href="https://www.example.com/category/recipes/page/4/" />
```

michaeltorbert:
I don't see any compelling reason for either, but sure, we can do double.

EkoJR:
They way it currently is, is incorrect. WP Coding standards say we should be using single quotes to wrap strings, and this is the other way around and should be fixed.

The only time I've seen double quotes being useful is with SQL strings. Otherwise WP standards tend to be better.

michaeltorbert:
@EkoJR You're thinking about PHP coding standards. HTML coding standards are either: https://make.wordpress.org/core/handbook/best-practices/coding-standards/html/#quotes

PR #2352, opened 2019-04-25
Use double quotes for pagination links

arnaudbroes:
#2088.


wpsmort:
This is tested and is good.

PR #2355, opened 2019-04-25
Use AIOSEOP_PLUGIN_NAME constant for admin menu

arnaudbroes:
This relates to https://github.com/semperfiwebdesign/aioseop-pro/issues/482.


arnaudbroes:
@wpsmort follow these instructions to test:

1. Add Pro folder to the PR version
2. Set AIOSEOPPRO to true
3. Set $aioseop_plugin_name to 'All in One SEO Pack Pro' 

The two first steps are obvious but the last step is important because that value is used in Pro.

wpsmort:
@arnaudbroes Changing this has the negative effect of the menu label wrapping to two lines which looks terrible.  I don't think we should make this change and should stay with the menu lable we have now.

Issue #1972, opened 2018-10-12
Add Pro to sitemap source code as appropriate


michaeltorbert:
We should add Pro to AIOSEOP ~~(otherwise the version number is confusing)~~.

![](https://user-images.githubusercontent.com/18156697/46801662-94fb1100-cd5b-11e8-98d6-54effae01e9b.png)

From https://github.com/semperfiwebdesign/aioseop-pro/issues/444

PR #2358, opened 2019-04-25
Add Pro to sitemap source code

arnaudbroes:
#1972.


arnaudbroes:
@wpsmort follow these instructions to test:

1. Add Pro folder to the PR version
2. Set AIOSEOPPRO to true
3. Set $aioseop_plugin_name to 'All in One SEO Pack Pro'

The two first steps are obvious but the last step is important because that value is used in Pro.

wpsmort:
This is tested and is good.

Issue #1271, opened 2017-10-04
Change blog_xxxx macro to site_xxxx macro


michaeltorbert:
~~This would require a db upgrade.~~ Let's continue to support blog_title unofficially.

arnaudbroes:
At some point we'll probably want to do a db migration - #2361

PR #2360, opened 2019-04-26
Add %site_title% and %site_description% macros

arnaudbroes:
#1271.


michaeltorbert:
This should have unit testing for the old macros as well as the new in order to verify the behavior.

arnaudbroes:
We should have for each and another to compare them and see if the result is the same.

wpsmort:
This is tested and is good.

arnaudbroes:
@EkoJR reach out to me first before looking at this since I'll need to add a unit test.

aioseop_class.php
```diff
@@ -919,6 +919,9 @@ public function get_title_format( $args ) {
 		$w            = $info['w'];
 		$p            = $info['p'];
 
+		if ( strpos( $title_format, '%site_title%' ) !== false ) {
```
EkoJR: The parent class method for strpos should be used instead. (multiple occurances)

aioseop_class.php
```diff
@@ -919,6 +919,9 @@ public function get_title_format( $args ) {
...
+			$title_format = str_replace( '%site_title%', get_bloginfo( 'name' ), $title_format );
+		}
 		if ( strpos( $title_format, '%blog_title%' ) !== false ) {
```
EkoJR: It might help to add a dev note here that %blog_title% is deprecated.

aioseop_class.php
```diff
@@ -1462,7 +1465,13 @@ function get_aioseop_title( $post, $use_original_title_format = true ) {
 		} elseif ( is_search() && isset( $s ) && ! empty( $s ) ) {
 			$search = esc_attr( stripslashes( $s ) );
 			$title_format = $aioseop_options['aiosp_search_title_format'];
-			$title        = str_replace( '%blog_title%', $this->internationalize( get_bloginfo( 'name' ) ), $title_format );
+			$title        = str_replace( '%site_title%', $this->internationalize( get_bloginfo( 'name' ) ), $title_format );
+			if ( strpos( $title, '%blog_title%' ) !== false ) {
```
EkoJR: The parent class method for strpos should be used instead. (multiple occurances)

Using Yoda operations is also a good habit to get into. `if ( 'value === $variable ) {`. It's a WP Standard and values are generally shorter than variable & function names. (multiple occurances)

aioseop_class.php
EkoJR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/53b0a5f9c74941979d645caa4866145287f7e944/admin/aioseop_module_class.php#L302

arnaudbroes: Done.

aioseop_class.php
arnaudbroes: Done.

aioseop_class.php
arnaudbroes: Done.

arnaudbroes:
@EkoJR I've addressed all comments. Now I am waiting on either you or @contactashish13 to help me finalize my PHPUnit setup so I can add the unit tests that were requested.

arnaudbroes:
@EkoJR this can be code-reviewed again. I implemented all the comments and added the required unit tests. I'm going to expand those unit tests to include all pages (homepage, search page, 404 page, ...) and also test all scenario's (homepage title, aioseop title, homepage as static page, etc ..)

Issue #875, opened 2017-04-10
Add Yandex and Baidu Webmaster Tools verification


michaeltorbert:
and any others we're missing including:

Baidu

wpsmort:
Requested by kemal durmu by email today.

RUSxDeviL1979:
please tell me when will add Yandex webmaster ?

wpsmort:
@RUSxDeviL1979 This is on our roadmap but we don't have a release date as yet.  We'll update this issue when we do have one.

FYI... you can add the HTML code for Yandex Webmaster verification in the Additional Headers fields under All in One SEO > General Settings > Advanced Settings - https://semperplugins.com/documentation/all-in-one-seo-pack-advanced-settings/

PR #2362, opened 2019-04-27
Add Yandex and Baidu Webmaster Tools verification

arnaudbroes:
#875.


wpsmort:
This is tested and is good.

Issue #2357, opened 2019-04-25
Change 'All in One SEO' in sitemap


arnaudbroes:
This is yuck and should use the constant AIOSEOP_PLUGIN_NAME.
![image](https://user-images.githubusercontent.com/18156697/56774772-20853e00-6792-11e9-9d87-2d42708e2d4f.png)

PR #2365, opened 2019-04-28
Use AIOSEOP_PLUGIN_NAME in sitemap instead of string

arnaudbroes:
#2357.

I also removed a redundant link and made the existing ones open as new tabs.


wpsmort:
This is tested and is good.

PR #2366, opened 2019-04-28
Change social meta home title field to text input 

arnaudbroes:
#1336.


wpsmort:
This is tested and is good.

Issue #1790, opened 2018-07-17
css improvement for text input fields


michaeltorbert:
Reported via email by Mark:

_There needs a change to the size (height) if the input field boxes used on each post or page editor, where we can enter Title, Keywords, etc.

The letters placed into the input box are cut-off at the bottom. Some letters are indistinquishable because the bottom portion is cut off.

This is the CSS file that needs changing;

Location of the CSS File: /wp-content/plugins/all-in-one-seo-pack/css/modules/aioseop_module.css_

```
.aioseop input[type="text"] {
color: #515151;
height: 45px; /*this one, changed from 35px to 45px*/
padding: 10px 0 10px 10px;
font-size: 14px;
width: 95%;
max-width: 600px;
}
```

_With that change, the letters are not cut off._

arnaudbroes:
@wpsmort can you see if you're able to reproduce this? I'm not able to.

wpsmort:
@arnaudbroes I'm not able to reproduce this in any browser on Mac.  However, we can certainly improve the CSS for text input fields by setting a min-height instead of a fixed height.  I've created PR #2367 for this change.

PR #2367, opened 2019-04-29
CSS improvement to text input fields

wpsmort:
#1790 

This changes the fixed height to a min-height


arnaudbroes:
This is tested and is good.

PR #2369, opened 2019-04-30
Add separate mailing list for Pro users

arnaudbroes:
https://github.com/semperfiwebdesign/aioseop-pro/issues/307.


arnaudbroes:
@wpsmort you should be able to tell the difference by looking at the signup forms; the one for free has already been given a new look; the one for Pro needs still needs to be redesigned.

wpsmort:
This is tested and is good.

Issue #2373, opened 2019-05-02
Help icons alignment is off in WP v5.2 


wpsmort:
WordPress v5.2 changes CSS line-height values to be unitless - https://core.trac.wordpress.org/ticket/44643

This means that the line-height for .postbox .inside is changed from 1.4em to 1.4 and as a result, our help icon alignment in the 3.0 branch is affected.  See screenshot below:

<img width="583" alt="Screen Shot 2019-05-02 at 11 45 34 AM" src="https://user-images.githubusercontent.com/17548525/57088234-d212fa80-6ccf-11e9-8a73-0fd8c1634917.png">

michaeltorbert:
When testing, we need to make sure the new CSS is appropriate in older supported versions of WP as well.

michaeltorbert:
Also, since 3.0 may not be out before WP 5.2, we may need to consider a 2.12.1 release for this.

wpsmort:
There are also problems with RTL styling:

<img width="593" alt="Screen Shot 2019-05-02 at 12 45 48 PM" src="https://user-images.githubusercontent.com/17548525/57091911-3b970700-6cd8-11e9-9bbb-73802b1abf52.png">

PR #2378, opened 2019-05-02
Fixes for WP v5.2 alignment issues

wpsmort:
#2373 

This is the PR for the 3.0 branch


arnaudbroes:
This is tested and is good.

michaeltorbert:
We don't do two identical commits to different branches in the repository. That isn't have it works and would cause a merge conflict.

PR #2379, opened 2019-05-03
Fix exclude terms from custom taxonomies

EkoJR:

Adds custom taxonomy support to Exclude Categories (Exclude Terms) in Sitemap settings.

Ran grunt, and found errors unrelated to this PR/Issue; which will need to be addressed in a separate PR.

# Update function added
`aiosp_sitemap_excl_categories` is converted to `aiosp_sitemap_excl_terms`. 

Note: The 3.0 branch is already set to `3.0-dev`; `last_active_version` will need to be reverted in the database.

# To Reproduce
1) Have free plugin active.
2) Go to AIOSEOP Sitemap XML settings page.
3) Navigate to Exclude Items > Exclude Terms
4) Enter or select term name.

**Actual**: Only builtin taxonomy (categories) terms are displayed.
**Expected**: All taxonomy terms should be displayed. **Unless taxonomy is unselected in XML Sitemap > Taxonomies**


arnaudbroes:
@EkoJR I'm not able to access the XML Sitemap module menu on two different test sites due to a fatal error. Here's the error from the log file:

> Got error 'PHP message: PHP Warning:  array_search() expects parameter 2 to be array, string given in /home/225863.cloudwaysapps.com/rswarvawya/public_html/wp-content/plugins/all-in-one-seo-pack-0d52edd054ee0d3e0768a2b1133533d21a4f3d90/modules/aioseop_sitemap.php on line 724
> 
> PHP message: PHP Fatal error:  Uncaught Error: Cannot unset string offsets in /home/225863.cloudwaysapps.com/rswarvawya/public_html/wp-content/plugins/all-in-one-seo-pack-0d52edd054ee0d3e0768a2b1133533d21a4f3d90/modules/aioseop_sitemap.php:726. Stack trace:\n#0 /home/225863.cloudwaysapps.com/rswarvawya/public_html/wp-content/plugins/all-in-one-seo-pack-0d52edd054ee0d3e0768a2b1133533d21a4f3d90/modules/aioseop_sitemap.php(806): All_in_One_SEO_Pack_Sitemap->add_post_types()\n#1 /home/225863.cloudwaysapps.com/rswarvawya/public_html/wp-includes/class-wp-hook.php(286): All_in_One_SEO_Pack_Sitemap->add_page_hooks('')\n#2 /home/225863.cloudwaysapps.

EkoJR:
I have it fixed, but I'm not completely sure why `aiosp_sitemap_taxonomies` wasn't being set as an array as expected.

arnaudbroes:
@EkoJR the search field works well but there's one bug and there should be one UI change as well:

1. When a Category term is excluded, all Posts belonging to that term are excluded as well. That's good. However, when a term of any other taxonomy (I tested both Tags and WooCommerce Product Categories) is excluded, the posts belonging to that term are not excluded from the sitemap.

2. You already brought this up on Slack with me, but it's not clear to which taxonomy each term belongs so ideally the parent taxonomy's name would be suffixed in brackets after the term. Ashish was able to do that in his PR. Can you look into that and see if it's possible with the JQuery plugin you're using?

EkoJR:
Regarding the first issue. It looks like it is an issue with get_posts() referring to categories. I suspect it's more than just WC Products. So, I'll have to 1) convert the function to use WP_Query, including other functions that use it. 2) Change the data in the elements. 3) I have to change the database setting to also store the taxonomy the term is from as well; will resemble a tax_query which uses it.

The second issue is resolved though.

EkoJR:
I just realized I have to add the changes to the updater.

EkoJR:
Updated description to include some details on the update.

If you are attempting to trigger the update, you will need to change the version number in the database.

If you don't update, you may run into an error on page load; which may resolve on save.

arnaudbroes:
@EkoJR excluding terms doesn't seem to have an affect on the sitemap at the moment.

EkoJR:
@arnaudbroes It's working for me. There was a force push though; to resolve a merge conflict that would have occured. So, are you sure you have the recent changes?

arnaudbroes:
@EkoJR so it's now correctly excluding all posts that are assigned to a certain term, but it's no longer excluding that term itself in the sitemap.

arnaudbroes:
I've tested this and it looks good. However, I'd like @wpsmort to do a round of testing as well since this is an important feature. 

@wpsmort if you install the plugin on an existing website, you might see a notice at the top of the XML Sitemap menu. Users who update shouldn't see this and you should be able to resolve it by simply adding an excluded term and then saving the options.

wpsmort:
I have tested this on my large site with lots of terms and it works great.  I only found one problem which is when installing the plugin on a brand new site, I get this notice on activation:

PHP Notice:  Undefined index: modules in /wp-content/plugins/all-in-one-seo-pack/inc/aioseop_updates_class.php on line 265

wpsmort:
This is tested and is good.

Issue #2387, opened 2019-05-07
Unit Testing - Title format unit test not triggering wp_title()


arnaudbroes:
The title format unit test in test-rewrite-titles.php is not triggering the wp_title() function in aioseop-class.php which is why the title isn't being rewritten by the plugin. Currently, the tests aren't throwing any errors because the WP title is the same as the expected result, but when we'd expand the test with more macros (#2384), they'll start to fail.

@contactashish13 seems to have run into the same issue in test-meta.php but didn't troubleshoot further or hasn't been able to resolve it so far.

arnaudbroes:
This should looked after https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2394 is done.

arnaudbroes:
Let's wait till the title code is moved to its own file.

Issue #2384, opened 2019-05-05
Add additional scenarios to rewrite title unit tests


arnaudbroes:
Once https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2360 is merged, I should expand those unit tests to include all page types (homepage, search page, 404 page, etc) as well as all scenarios (homepage title, aioseop title, homepage as a static page, etc)

Issue #1754, opened 2018-06-28
add %currentyear% title format


jayyap24:
I'm trying to figure out how to pull up the "current year" using a string so I don't have to update the SEO title anymore. For example, Yoast has "%%CURRENTYEAR%%" and it will get whatever is the current year.

Do you guys have that in All in One?

michaeltorbert:
We have current_date, but not current year. 
However now that there's an issue open for it, we'll be happy to add current year as well. :)

michaeltorbert:
We should even go so far as to have an article written on "How to add the current year to your title".

arnaudbroes:
This will need a unit test as well.

PR #2388, opened 2019-05-07
Add %current_year% macro

arnaudbroes:
#1754.

This macro will be incorporated in the existing unit test for title format macros as part of #2384 but first we need to get #2387 out of the way which is rather complex.

@wpsmort I've added this macro for the following title format fields:

- Homepage
- Post
- Custom post types
- Page
- Category
- Custom taxonomies
- Description
- Search page

It has not been added for:

- Archive page
- Date archive page
- Author archive page
- 404 page
- Paginated page

If it should be added for any of the above, then let me know, but I don't see a scenario where these need this macro and they don't seem to have much macros anyway.


wpsmort:
This is tested and is good.

Issue #2121, opened 2018-12-24
Remove Google+ (and two filter hooks)


michaeltorbert:
Google+ is shutting down in a few months, so we should be prepared to remove any associated functionality such as rel="publisher".
https://9to5google.com/2018/12/21/google-apis-shut-down-date/

I've tentatively assigned a 3.3 milestone, but we can adjust that up or down as appropriate.

PR #2390, opened 2019-05-07
remove Google+ (and two filter hooks)

michaeltorbert:
issue #2121 

## Testing instructions
@wpsmort When testing, you may also want to review the code since you're pretty familiar with it. Still assign it for secondary code review.

## Code review instructions
Make sure none of the code removed is used anywhere else.

# Filters Removed 

* `aioseop_google_publisher`
* `aioseop_google_author`


wpsmort:
This is tested and is good.

wpsmort:
There is no API docs for aioseop_google_publisher and aioseop_google_author

michaeltorbert:
@EkoJR Good catch. I've modified the title so we can remember to include a mention of it in the changelog.

Issue #2241, opened 2019-03-02
Make remaining strings translatable


arnaudbroes:
![image](https://user-images.githubusercontent.com/18156697/53683396-a344cc00-3d00-11e9-93bb-95efc37ba2fc.png)

arnaudbroes:
Here's another one:
![image](https://user-images.githubusercontent.com/18156697/53683748-bbb6e580-3d04-11e9-8b33-98880d6ee087.png)

PR #2391, opened 2019-05-07
Make remaining strings translatable

arnaudbroes:
#2241.

@wpsmort to test this, just go through all menus and check if everything looks good. However, you also need to look at the sitemap since these changes affect the strings there especially.


wpsmort:
@arnaudbroes I tested this and found two problems as follows:

1. Line 3667 of aioseop_class.php needs a space before "by".  Currently the text looks like this:

All in One SEO Pack 3.0-devby Michael Torbert of Semper Fi Web Design

2. The word "URLs" is now missing when you view the sitemap.  Currently the text looks like this:

"This sitemap contains 54 ."

arnaudbroes:
This is fixed and can be tested again.

wpsmort:
This is tested and is good.

admin/display/general-metaboxes.php
```diff
@@ -187,7 +187,7 @@ static function pro_meta_content() {
 
 		echo '</ul>';
 
-		echo '<a href="https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/new" />Click here</a> to file a feature request/bug report.';
+		echo __( '<a href="https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/new" />Click here</a> to file a feature request/bug report.', 'all-in-one-seo-pack' );
```
michaeltorbert: Use a placeholder for the URL.

aioseop_class.php
```diff
@@ -240,7 +240,7 @@ function __construct() {
 			'paged_format'                => array(
 				'name'     => __( 'Paged Format:', 'all-in-one-seo-pack' ),
 				'type'     => 'text',
-				'default'  => ' - Part %page%',
+				'default'  => __( 'Part %page%', 'all-in-one-seo-pack' ),
```
michaeltorbert: Use placeholder for the "%page%" so that doesn't get translated.

aioseop_class.php
```diff
@@ -3661,11 +3661,10 @@ function wp_head() {
...
 		// Logging - rewrite handler check for output buffering.
 		$this->check_rewrite_handler();
-		if ( AIOSEOPPRO ) {
-			echo "\n<!-- All in One SEO Pack Pro $this->version by Michael Torbert of Semper Fi Web Design";
-		} else {
-			echo "\n<!-- All in One SEO Pack $this->version by Michael Torbert of Semper Fi Web Design";
-		}
+
+		echo "\n";
+		printf( '<!-- %1$s %2$s %3$s Michael Torbert %4$s Semper Fi Web Design', AIOSEOP_PLUGIN_NAME, $this->version, __( 'by', 'all-in-one-seo-pack' ), __( 'of', 'all-in-one-seo-pack' ) );
```
michaeltorbert: This should probably be flipped, so that translators see the entire string, even if part of it is placeholders. 
Then we should use translator comments so they have context for the two words they're translating.

modules/aioseop_sitemap.php
```diff
@@ -251,7 +251,7 @@ public function __construct() {
 			$this->prio     = array_merge( $arr_no, $prio );
 
 			$freq = array();
-			foreach ( array( 'always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never' ) as $f ) {
+			foreach ( array( __( 'always', 'all-in-one-seo-pack' ), __( 'hourly', 'all-in-one-seo-pack' ), __( 'daily', 'all-in-one-seo-pack' ), __( 'weekly', 'all-in-one-seo-pack' ), __( 'monthly', 'all-in-one-seo-pack' ), __( 'yearly', 'all-in-one-seo-pack' ), __( 'never', 'all-in-one-seo-pack' ) ) as $f ) {
```
michaeltorbert: Are these settings? If so, we don't want to translate option values.

inc/sitemap-xsl.php
```diff
@@ -110,8 +110,8 @@
 		<table cellpadding="3">
 			<thead>
 			<tr>
-				<th width="50%">URL</th>
-				<th>LastChange</th>
+				<th width="50%"><?php echo __( 'URL', 'all-in-one-seo-pack' ); ?></th>
+				<th><?php echo __( 'Last Change', 'all-in-one-seo-pack' ); ?></th>
```
michaeltorbert: We probably don't want to translate sitemap data.

modules/aioseop_sitemap.php
arnaudbroes: These can be translated but there's a small issue that has to be addressed first. I've opened https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2401 for that. Once that issue is resolved, the plugin will be 100% translatable.

arnaudbroes:

wpsmort:
This is tested and is good.

Issue #825, opened 2017-03-16
fix missing spaces for translation


arnaudbroes:
Right now some strings start with a space, but these are not visible on the translation platform, causing some strings to be pasted together and look like this: (click hereto upgrade to Pro).

We should correct these for:

**aioseop_sitemap.php** line 480 & 481
' Post Type'
' Taxonomy'

**aioseop_sitemap.php** line 620 & 622
' Your sitemap has been created'
' and changes are automatically submitted to search engines'

**general-metaboxes.php** line 42
' to upgrade to Pro Version and get:'

arnaudbroes:
I believe we can either 

(1) add non-breaking space HTML entities (&nbsp;) to partial strings so that our translators take these into account; 

or (2) remove the spaces from those partial strings and concatenate them as separate strings (for example "I am" . " " . "Arnaud Broes" or "I am" . "&nbsp;" . "Arnaud Broes"). This makes the code a little messier but also ensures translators can't make any mistakes.

@michaeltorbert what do you think?

michaeltorbert:
Probably best to use placeholders https://codex.wordpress.org/I18n_for_WordPress_Developers#Placeholders

PR #2397, opened 2019-05-08
Remove spaces from translatable strings

arnaudbroes:
#825.

@wpsmort you'll just want to have a look at the strings this PR addresses and see if they still look good.


wpsmort:
This is tested and is good.

admin/display/general-metaboxes.php
```diff
@@ -43,7 +43,7 @@ static function display_extra_metaboxes( $add, $meta ) {
 							<strong>
 								<?php
 								echo aiosp_common::get_upgrade_hyperlink( 'side', __( 'Pro Version', 'all-in-one-seo-pack' ), __( 'CLICK HERE', 'all-in-one-seo-pack' ), '_blank' );
-								echo __( ' to upgrade to Pro Version and get:', 'all-in-one-seo-pack' );
+								printf( ' %1$s', __( 'to upgrade to Pro Version and get:', 'all-in-one-seo-pack' ) );
```
michaeltorbert: When there's only one placeholder, there's no need to use %1$s, just %s is fine.

modules/aioseop_sitemap.php
```diff
@@ -649,8 +649,8 @@ public function add_post_types() {
 			$this->default_options['taxonomies']['default']              = array_keys( $this->default_options['taxonomies']['initial_options'] );
 			$this->default_options['excl_categories']['initial_options'] = $this->get_category_titles();
 
-			$post_name = __( ' Post Type', 'all-in-one-seo-pack' );
-			$tax_name  = __( ' Taxonomy', 'all-in-one-seo-pack' );
+			$post_name = sprintf( ' %1$s', __( 'Post Type', 'all-in-one-seo-pack' ) );
```
michaeltorbert: These are overly complicated and difficult to read. Something like the following should work just as well.

`$post_name = ' ' . __( 'Post Type', 'all-in-one-seo-pack' ) );`

arnaudbroes:

wpsmort:
This is tested and is good.

Issue #2218, opened 2019-02-19
Convert - XML Sitemap - Remove "Only show 1000 posts per sitemap file" message


michaeltorbert:
Reported here: https://wordpress.org/support/topic/bugs-with-the-sitemap-please-fix-them/

> Even though you give the option to choose the split number, if I choose 2000 per split instead of default 1000, I am haunted by a pop-up message showing up everywhere annoying me to set it to 1000 because this is how its recommended! There is a little X but if I close the popup and simply navigate to a another page, it will show up again and again and again. There is no way to permanently disable this annoying notification to force me to have 1000 posts per split.

arnaudbroes:
Shouldn't we also just make that notification appear once?

wpsmort:
I cannot reproduce this behavior.  If I click the X to dismiss the message and then navigate away and then back again, the message does not come back.  I've tried but cannot make that message reappear.

@arnaudbroes Can you try to reproduce this.

arnaudbroes:
I'll take a look. I believe it's been reported a number of times now.

arnaudbroes:
@wpsmort I'm not able to reproduce this either.

michaeltorbert:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/7550dd8f46ce436216288c3dada2e71ad9929f12/modules/aioseop_sitemap.php#L458 is where the magic happens. It is indeed written so that the message doesn't show again. While this will be obsoleted once we have the new notification system in place, perhaps @EkoJR or @contactashish13 could have a look at it and see if anything jumps out in our current notice code that could cause the notices to keep coming back.
Recall we used to have this problem with the WooCommerce notice a few years ago where it wouldn't go away for some people and we could never manage to reliably reproduce it.

contactashish13:
@michaeltorbert the only reason that I see for this is the ajax method not being fired that updates the user meta. This may happen because of a JS error because of some other component. That is why the user is being redirected to another page when clicking on the 'X' instead of it firing the ajax method. Can we get a screenshot of the JS console?

michaeltorbert:
@contactashish13 Unfortunately we're not able to reliably reproduce it on our site or a user's site.

contactashish13:
I don't think we can reproduce it on another instance as it will be specific to the components installed on that site. The user says it happens all the time so it seems like it's consistent behavior for him. We can't get a screenshot of his console?

[...]

PR #2403, opened 2019-05-09
Dev 3.0 notices

EkoJR:
Dev branch with admin notices

 ## Testing

Sitemap 1000+ warning - will need 1000+ posts to trigger.

Blog Search Engine Visibility Notice - will need the blog visibility setting disabled in WP's reading settings.

### DIFF
--- a/admin/display/notices/blog-visibility.php
+++ b/admin/display/notices/blog-visibility.php
@@ -60,14 +60,14 @@ function aioseop_notice_blog_visibility() {
 		'action_options' => array(
 			array(
 				'time'    => 0,
-				'text'    => __( 'Settings > Reading', 'all-in-one-seo-pack' ),
+				'text'    => __( 'Update Reading Settings', 'all-in-one-seo-pack' ),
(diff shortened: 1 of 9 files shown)


michaeltorbert:
#1854 
#2218

michaeltorbert:
The 1000 posts notice doesn't show up for me.

michaeltorbert:
Screen Shot 2019-05-09 at 12 19 21 PM


The blue button should probably be an action instead of a vague reference to the location it's sending you to.
I've included ones from WooCommerce for comparison.

arnaudbroes:
@michaeltorbert agreed. I think "Update settings" should work for both those notices. Also, let's remove that dot behind "Remind me later" in the first notice so they are uniform.

wpsmort:
I recommend that the button label should describe the specific action, i.e.

Update Sitemap Settings
Update Reading Settings

michaeltorbert:
I like the Update xxx Settings idea, as well as removing the .

arnaudbroes:
This has been tested and looks good.

Issue #1963, opened 2018-10-09
Update plugin icon to modern gear logo


michaeltorbert:
We also will want to do this for wp.org in a separate issue.

arnaudbroes:
Related to #1574.

Issue #1861, opened 2018-08-22
Menu Icon reverts to default cogwheel w/ alt. plugin folder name


EkoJR:
**Describe the bug**
The icon used within the wp-admin menu uses the default cogwheel image/dashicon when the plugin directory is changed to a name different from what is registered with WP (possibly as well as coded in AIOSEOP); which is `all-in-one-seo-pack`.

**To Reproduce**
1. Change the plugin folder name to something other than `all-in-one-seo-pack`.
2. Activate the plugin (if deactivated)
3. Should see AIOSEOP's menu icon is WP's default; instead of the shield icon.

**Expected behavior**
The menu icon should resemble a shield.

**Screenshots**
Broken
![menu_icon](https://user-images.githubusercontent.com/2794445/44441307-c7b93200-a580-11e8-8209-3b29ae7ae97b.png)
Expected
![menu_icon_fix](https://user-images.githubusercontent.com/2794445/44441310-cab42280-a580-11e8-8aee-1407245dda58.png)

**Additional context**
The Icon is applied using CSS, and uses the toplevel id that is supplied by WP by default. It's possible there may be an alternative with CSS classes that is also supplied by WP.

michaeltorbert:
BTW this should be the newest gear icon

arnaudbroes:
@EkoJR I came to the same conclusion in #2349, however, the toplevel ID is generated by our plugin based off the plugin directory name. I'm guessing JavaScript is the only thing that will solve the trick here? Unless I'm mistaken, we can't dynamically change the target ID inside our CSS code.

Issue #1917, opened 2018-09-17
Change Admin Menu icons to an Icon-Font


EkoJR:
![title menu icon](https://user-images.githubusercontent.com/2794445/45653306-5a6bc480-ba8c-11e8-8018-0aaadc82d59a.png)

Related to #289

michaeltorbert:
This should be the newest gear icon

michaeltorbert:
I don't remember now why we did it this way, but here's the shield: https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/css/aiosp_admin.css#L5

PR #2407, opened 2019-05-09
 Update plugin icon to modern gear logo

arnaudbroes:
@wpsmort this PR fixes both #1963 and #1861 at the same time so you'll also want to test this by changing the plugin directory name.

You'll notice that the plugin icon turns white only after a second; this is not a bug but expected behaviour. It turns white once the page is fully loaded. You'll notice the same behaviour when you compare this with our competitor.


wpsmort:
This is tested and is good.

aioseop_class.php
```diff
@@ -4717,7 +4717,8 @@ function admin_menu() {
 			$menu_name, $menu_name, apply_filters( 'manage_aiosp', 'aiosp_manage_seo' ), $file, array(
 				$this,
 				'display_settings_page',
-			)
+			),
```
EkoJR: Since you're modifying this call function, could you change it to...

```
add_menu_page(
	$menu_name,
	$menu_name,
	apply_filters( 'manage_aiosp', 'aiosp_manage_seo' ),
	$file,
	array( $this, 'display_settings_page' ),
	aioseop_get_menu_icon()
);
```

It'll make it easier to read. I was trying to double check the function params, and I think you mentioned using the wrong line at one point; not sure if this is the same set of lines.

inc/aioseop_functions.php
```diff
@@ -1081,3 +1081,80 @@ function aioseop_formatted_date( $date = null, $format = null ) {
...
+ * Returns a base64 data URI for the SVG icon in the menu.
+ *
+ * @since 3.0
+ *
+ * @return string
+ */
+if ( ! function_exists( 'aioseop_get_menu_icon' ) ) {
+
+	function aioseop_get_menu_icon() {
+		$svg = '<svg version="1.1" xmlns="http://www.w3.org/2000/svg"
```
EkoJR: Is it possible to make this an SVG file? If so, you might be able to eliminate this function by linking the SVG file directly.

aioseop_class.php
```diff
@@ -4717,7 +4717,8 @@ function admin_menu() {
...
+			aioseop_get_menu_icon()
```
EkoJR: Was it possible to use a URL?

This could also use the class name of font icon; which would be more a future issue.

aioseop_class.php
arnaudbroes: Same set of lines. I was calling aioseop_get_menu_icon() within the array above it which is why it wasn't working.

inc/aioseop_functions.php
arnaudbroes: Has to be a base64-encoded SVG using a data URI which needs to begin with 'data:image/svg+xml;base64,' so I'm afraid that linking it directly is not possible (I tried).

aioseop_class.php
arnaudbroes: As far as SVG files go, it has to be a base64-encoded SVG using a data URI. However, using Dashicons here is possible: 
> "Pass the name of a Dashicons helper class to use a font icon, e.g. 'dashicons-chart-pie'. "
Feel free to open an issue for that.

arnaudbroes:

wpsmort:
This is tested and is good.

inc/aioseop_functions.php
EkoJR: Ok, I was just curious if you had explored it.

arnaudbroes:
This also made #1917 redundant.

Issue #1337, opened 2017-11-04
Add character counter for homepage OG title & meta description


arnaudbroes:
Since it doesn't have one yet.

arnaudbroes:
#2366 needs to be merged first.

PR #2408, opened 2019-05-09
 Add character counter fields to OG Home Title & Description

arnaudbroes:
#1337.


wpsmort:
This is tested and is good.

modules/aioseop_opengraph.php
```diff
@@ -335,6 +335,10 @@ public function init() {
 					'name'     => __( 'Home Title', 'all-in-one-seo-pack' ),
 					'default'  => '',
 					'type'     => 'text',
+					'class'    => 'aioseop_count_chars',
+					'count'    => true,
+					'count_desc' => $count_desc,
```
EkoJR: When you change the `=>` alignment, you usually want to align the rest. Otherwise WPCS would flag it.

For example

```
'hometitle'              => array(
	'name'       => __( 'Home Title', 'all-in-one-seo-pack' ),
	'default'    => '',
	'type'       => 'text',
	'class'      => 'aioseop_count_chars',
	'count'      => true,
	'count_desc' => $count_desc,
	'size'       => 95,
	'condshow'   => array(
		'aiosp_opengraph_setmeta' => array(
			'lhs' => 'aiosp_opengraph_setmeta',
			'op'  => '!=',
			'rhs' => 'on',
		),
	),
),
```

arnaudbroes: Done.

arnaudbroes:

wpsmort:
This is tested and is good.

Issue #2139, opened 2019-01-10
Add recommended and/or max. amount of chars for og:title and og:description


arnaudbroes:
We might want to consider this:

> As with metadata, the lengths of your OG tag are defined by the platform that they will be viewed on. For example, a link shared to Facebook will create a card that looks different on mobile and desktop Facebook and will look different again on LinkedIn.
> 
> As such, the optimum length is more defined by what is most likely to be seen by all users across any platform and, even then, you should keep your key messages upfront in the copy to reduce the risk of them being truncated.
> 
>* Optimum OG Title length: 55 characters or under. This ensures that your title will be visible on desktop and mobile devices and also allows for your description to be visible.
> * Maximum recommended OG Title length: 95 characters or under.
>
> * Optimum OG Description length: 55 characters or under. Combined with a 55 character title, your copy will be visible on all devices.
> * Maximum recommended OG Description length: 200 characters or under.

wpsmort:
@arnaudbroes Are you recommending we change the text and the character counter (yellow and red levels) under the Title and Description fields on the Social Settings tab?  If not, what are you recommending?

arnaudbroes:
@wpsmort I recommend the following:

1. OG:title
< 55 => green
>= 55 and <= 95 => yellow
> 95 => red

2. OG:description
< 55 => green
>= 55 and <= 200 => yellow
> 200 => red

Also, we need to add the counter to the Home Title and Home Description fields in the Social Meta menu. Let me know what you think. We might want to separate this into two specific issues.

wpsmort:
@arnaudbroes I agree with your recommendations.  

We have #1337 open for adding the counter to the Home Title and Home Description fields under the Social Meta settings.

wpsmort:
This is related to #808

wpsmort:
We agreed in the dev meeting today that, in the absence of any official documentation from Facebook, we'll go with the information in the thoughtshift.co.uk article that Arnaud found and we'll make it clear in our documentation and help that this is just a guide.

PR #2409, opened 2019-05-09
 Change recommended OG description length

arnaudbroes:
#2139.

@wpsmort this should also affect the auto-generated OG:descriptions so don't forget to test those too.


wpsmort:
@arnaudbroes I found two problems as follows:

1. This trims the og:description meta tag to 200 characters for manually entered Open Graph descriptions.  It should not affect manually entered descriptions, only auto generated descriptions.

2. The character counter does not turn yellow at 55 characters as per your recommendation here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2139#issuecomment-453830817

arnaudbroes:
@wpsmort we are also trimming user input for the OG:description field because Open Graph allows up to 1.000 characters as a max., so I put that back in with a comment to make it more clear in the code.

wpsmort:
@arnaudbroes This works fine English language text but breaks when I use a language such as Chinese.  Here's the steps to reproduce:

1. Create a post that has a long excerpt and long content containing Chinese characters
2. Check the auto generated og:description using the current live version of AIOSEOP.  You'll see it gets output and trimmed to 1000 characters.  The same with the twitter:description
3. Now activate the code from this PR and check the og:description and you'll find it is gone.  However, the twitter:description is present and is trimmed to 200 characters

See screenshot:

Screen Shot 2019-05-10 at 2 47 58 PM

arnaudbroes:
This is happening because @contactashish13's orginal code does not truncate words for the OG:description but looks for the next space character and trims the description up to that point. However, since Chinese (and other foreign languages) do not use spaces, the code simply returns an empty string. In order to solve this, I'll have to look at what we're doing for the regular meta description.

arnaudbroes:
@wpsmort I believe this fixes the original code. It runs the OG:description through the filter function for the regular SEO description.

wpsmort:
@arnaudbroes I tested the latest code and there are still two problems:

1. The manually entered OG Description is being truncated to 1000 characters.  We should not be truncating manually entered descriptions.

2.  There's still no og:description meta tag output for my Chinese language test page.  I've tested with other languages and they work fine, it's only Chinese where the og:description does not get output.

arnaudbroes:
@wpsmort I've created an issue for (1) since we've been doing that for a while now; this PR didn't start doing that - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2429

For (2), does your test page contain one big blob of Chinese text or is it separated in paragraphs? If it doesn't have paragraphs, add one and see if you then get a OG:description. There's not much we can do to fix this besides truncating at 200 characters, which is why I opened https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2424. This PR didn't start this either but made the issue more apparent because we lowered the amount from 1,000 to 200 characters.

wpsmort:
@arnaudbroes As we discussed last week regarding (2), this is something that has to be fixed.  We cannot just stop outputting the og:description for customers with Chinese language sites.  We either need to fix it in this PR so that the og:description is output, or a new PR needs to be created that fixes this.  As you know from our testing last week, adding a paragraph break does fix this but that's not the point, this PR doesn't output an og:description when we absolutely should.

arnaudbroes:
@wpsmort I believe this should do the trick.

wpsmort:
@arnaudbroes This fixes the problem with Chinese characters, however, it introduces a new problem.

When the excerpt or content contains an image before the text, no og:description or twitter:description meta tags are output.  

Steps to reproduce this:

1.  Check the box for Use Content for Autogenerated OG Descriptions under Social Meta > Advanced Settings
2. Create a post and add an image in the content before any text
3. Check the source code for the post and you'll see that the og:description and twitter:description are missing
4. Remove the image from the content and check the source code and you should see meta tags are now present.

This also happens if you are using the excerpt and you enter the HTML for an image in the Excerpt before any text.  It also happens if the image is in the middle of the content, in this instance only the text before the image is used for the autogenerated descriptions.

I believe that this commit broke this - https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2409/commits/b71a3a357e48c3195f0260a85aaa5760ed635c71.  This was originally fixed in PR #1420

arnaudbroes:
@wpsmort great catch! I fixed it.

wpsmort:
@arnaudbroes This now truncates og:descriptions to 160 characters instead of 200 characters.

wpsmort:
This is tested and is good.

Issue #2257, opened 2019-03-07
Default to 1000 posts per sitemap file


michaeltorbert:
This just needs to be changed from 10k to 1000 in the settings code.

arnaudbroes:
Relevant to #2218 .

PR #2414, opened 2019-05-10
Default to sitemap indexes and 1000 posts per sitemap file

michaeltorbert:
## Issue 

#1929 #2257 

## Proposed changes

For some large sites, our sitemap code can't efficiently run. As we add more functionality to the sitemap, this will only get worse. The best solution to this is to use sitemap indexes.
Long term, we'll probably want to use indexes exclusively, but for now, we can just default to them on and use only 1000 posts per sitemap file.

## Types of changes

This should only change the defaults. If a person has saved new settings, it shouldn't change anything. 

## Testing instructions
- Make sure to test on an existing site as well as a brand new site.
- The defaults should be index=on and posts per file = 1000.

##Further comments
- The behavior of our settings are that defaults aren't saved to the db if the site admin hasn't clicked "update options". This means that even on an existing install, the defaults will change if they haven't ever clicked the update button. This isn't necessarily desired behavior, but it is expected behavior.

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -181,7 +181,10 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 					),
 					'default'         => 0,
 				),
-				'indexes'     => array( 'name' => __( 'Enable Sitemap Indexes', 'all-in-one-seo-pack' ) ),
+				'indexes'     => array(
+					'name' => __( 'Enable Sitemap Indexes', 'all-in-one-seo-pack' ),
+					'default' => 'on',
+				),
 				'max_posts'   => array(
 					'name'     => __( 'Maximum Posts Per Sitemap Page', 'all-in-one-seo-pack' ),
 					'type'     => 'text',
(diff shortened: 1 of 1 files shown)


arnaudbroes:
@michaeltorbert it's not possible to override the amount of URLs per index anymore.

arnaudbroes:
@michaeltorbert this also had to be reverted; otherwise all indexes were still being populated with a max. of 1,000 URLs. I've already made the change and ran Grunt.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/233387cc9a34f4ef63e388df1375f242edeb416e/modules/aioseop_sitemap.php#L58-L66

wpsmort:
I tested this and found one problem when testing on a large site:

When the maximum posts per sitemap is 1000 or greater, then sitemaps are not paginated to 1000 URLs each.  Instead, all URLs are output in each sitemap.  However, if I set the maximum posts per sitemap to less than 1000, then the sitemaps are paginated to that number.

This happens on a new install and on an existing site.

See screenshot:

Screen Shot 2019-05-10 at 4 27 27 PM

arnaudbroes:
@wpsmort can be tested again.

We only had to change the defaults at the top of the file. I'm not sure where my head was at.

wpsmort:
This is tested and is good.

Issue #534, opened 2016-10-03
XML Sitemap module - Remove the Create Compressed Sitemap option


wpsmort:
**_Because this has issues on nginx_**, this option can be completely removed.  There is no need for a filter.  This should be enabled for all.

wpsmort:
See #945

michaeltorbert:
When would someone not want this available? We just need to make sure they know it exists (in the new indicator).

Is it one or the other... is this checked, is there still a non-compressed version?

wpsmort:
I don't believe there's a case where someone would not want this available.  It's something we should just do and then indicate that it's been done.

michaeltorbert:
I think we should remove this option, and it should be "enabled" in that if someone views your ...com/sitemap.xml.gz they get the compressed version, and if they visit ...com/sitemap.xml they get the uncompressed version, which I think is basically how it works now.
We're having so many issues with compression, we should make the default sitemap the non-compressed version, but where they can provide Google with the compressed link if they prefer.
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+compressed
https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+compression

michaeltorbert:
What we discussed is that we'll remove the option and all checks for it, and leave the functionality such that either the compressed or uncompressed URL works. We'll just explicitly supply the uncompressed version though, and update the documentation to make it clear we offer both.

michaeltorbert:
Considering we're moving in the direction of indexes, we should probably just remove all compression functionality since it doesn't really benefit us at all.

PR #2417, opened 2019-05-10
remove sitemap compression #534

michaeltorbert:
## Issue

#534 

## Proposed changes

Because compression doesn't really add much value, and has caused nothing but problems (for instance with nginx), we've decided to remove the functionality completely.
Browsing to a compressed URL should fall back to the uncompressed sitemap.

## Types of changes

-Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing instructions
-existing installs with compression enabled with indexes
-existing installs with compression enabled without indexes
-existing installs w/o compression enabled with indexes
-existing installs w/o compression enabled without indexes
-new installs
-full testing of the sitemap functionality

## Further comments

We need to make a note to include mention of this in the release post, as people may want to re-submit their sitemap to Google (though it isn't necessary).

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -204,10 +204,6 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 				'archive'     => array( 'name' => __( 'Include Date Archive Pages', 'all-in-one-seo-pack' ) ),
 				'author'      => array( 'name' => __( 'Include Author Pages', 'all-in-one-seo-pack' ) ),
 				'images'      => array( 'name' => __( 'Exclude Images', 'all-in-one-seo-pack' ) ),
-				'gzipped'     => array(
-					'name'    => __( 'Create Compressed Sitemap', 'all-in-one-seo-pack' ),
-					'default' => 'On',
-				),
 				'robots'      => array(
 					'name'    => __( 'Link From Virtual Robots.txt', 'all-in-one-seo-pack' ),
 					'default' => 'On',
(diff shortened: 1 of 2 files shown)


michaeltorbert:
When testing, we need to remember to do this from @wpsmort's comment https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1361#issuecomment-471545441

arnaudbroes:
I've tested this and it looks good but I'd like @wpsmort to also have a look since this is rather important.

wpsmort:
This is tested  and is good.

modules/aioseop_sitemap.php
```diff
@@ -1521,19 +1495,15 @@ public function log_start() {
 		 * Stop timing and log memory usage for debug info.
 		 *
 		 * @param string $sitemap_type
-		 * @param bool   $compressed
 		 * @param bool   $dynamic
 		 */
-		public function log_stats( $sitemap_type = 'static', $compressed = false, $dynamic = true ) {
```
EkoJR: phpDoc `since` noting a change to params

modules/aioseop_sitemap.php
```diff
@@ -1824,33 +1771,26 @@ public function add_xml_mime_type( $mime ) {
...
-		 * Write a single sitemap to the filesystem, handle compression.
+		 * Write a single sitemap to the filesystem.
 		 *
 		 * @param      $filename
 		 * @param      $contents
-		 * @param bool $gzip
 		 *
 		 * @return bool
 		 */
-		public function write_sitemap( $filename, $contents, $gzip = false ) {
```
EkoJR: phpDoc `since` noting a change to params.

Issue #39, opened 2016-03-17
Make user profile meta fields obvious


michaeltorbert:
It should be obvious that these fields are for AIOSEOP and not built into WP or from another plugin. related to #40

wpsmort:
Here's a screenshot of how this should potentially look:

![screen-capture](https://cloud.githubusercontent.com/assets/17548525/14157709/4d92cd3e-f69a-11e5-986f-d8b76ddafbbd.png)

iko-dev:
The image given in description is repeating the "Social Media Info" i.e. Google+, Twitter and Facebook. Attached is our understanding, please confirm or explain if we understood otherwise.

![user profile - issue 39](https://cloud.githubusercontent.com/assets/18193659/14583598/5eaae954-0440-11e6-8afd-92b823592f16.png)

wpsmort:
Your understanding is correct and your mockup matches what we want.

iko-dev:
thanks for reply, we will let you know after reviewing it.

Question: should we confirm each step with you before applying? or can we use our judgement to save time. And then explain you the reason.

wpsmort:
You can just go ahead and submit your code for review once you have finished.

michaeltorbert:
@iko-dev There's nothing in the pull request. You have to commit your changes to your fork first.

iko-dev:
Please check now.

Little Explanation about the task:
Its not possible to add title directly into this section. Because filters are used in this contact section and filters only provide fields and labels. So we added "All in one SEO" as field and hide out the input field in css.

The other option is to bring this "All in One SEO" section down in the end, in this way it disappears from users eye on first look.

michaeltorbert:
@iko-dev Pull requests need to be into the iko-dev branch, not master.

iko-dev:
Do you have access to this location
https://github.com/iko-dev/all-in-one-seo-pack/pull/1

michaeltorbert:
Whoever takes this over can look at the various closed PRs for guidance.

PR #2421, opened 2019-05-12
make social user profile fields more obviously from aioseop 

michaeltorbert:
Issue #39  

## Proposed changes

It's not always obvious that the Facebook/Twitter fields are from All in One SEO Pack. This PR makes a section header "All in One SEO Pack".

## Types of changes

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have added necessary documentation, or have created an issue for docs (if appropriate).

## Testing instructions
- To produce this functionality, enable the Twitter/Facebook profile options in the Social Meta Module Settings. You should then see the fields in your profile, along with the new AIOSEOP section heading.
- It should show up if either Twitter _or_ Facebook _or_ both fields are configured to show up on the profile, but _not_ if neither is. (ie you shouldn't see the heading but not either field.)

## Further comments

For the code review- my first thought was to hook into some relevant action or filter, but _edit_user_profile_ was the closest. To use this, we'd have to migrate our current fields to the new custom ones, since the current fields use the _user_contactmethods_ filter. I'm sure we also could have used javascript to hijack the page as well.
What I did instead was to add another profile field called "aioseop_title" and just use CSS to hide the input field and increase the font-size so it only _looks_ like it's an H2.

### DIFF
--- a/css/aiosp_admin.css
+++ b/css/aiosp_admin.css
@@ -76,11 +76,11 @@ li#wp-admin-bar-aioseop-pro-upgrade a.ab-item {
 	font-size: 105%;
 }
 
-label[for=aioseop_title]
+label[for=aioseop_edit_profile_header]
 {
 	font-size: 1.3em;
 }
 
-#aioseop_title {
+#aioseop_edit_profile_header {
 	display: none;
 }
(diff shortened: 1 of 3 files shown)


michaeltorbert:
@arnaudbroes @wpsmort 
You can decide amongst yourselves which of you will Test and who will Code Review (although each of you should probably both peek at the code and take a look at the output.

wpsmort:
This is tested and is good.

arnaudbroes:
This looks good. @michaeltorbert clever solution btw.

Issue #1757, opened 2018-07-03
change "media" to "attachment" 


michaeltorbert:
This can be confusing for some, so we should discuss whether to change it.

arnaudbroes:
We're only using one instance of Media (see below) on line 844 of /modules/aioseop_sitemap.php. Apart from that, all other occurences of Media aren't (hard)coded in the plugin but fetched from the list of Custom Post Types that the plugin detects. 

`$post_type_titles['attachment'] = __( 'Media / Attachments', 'all-in-one-seo-pack' );`

@michaeltorbert I think we'd be better off closing this issue as we would only be able to change it to "Attachment" in the XML Sitemap module, which may lead to even more confusion. If you disagree, feel free to reopen this issue.

arnaudbroes:
I realize we might be able to change this everywhere using the above code snippet but that seems like a lot of work for something so minor.

michaeltorbert:
In general settings "media" is listed, which some users take to mean the photos themselves, rather than the attachment post type.

PR #2426, opened 2019-05-13
Change 'Media' to 'Media / Attachments'

arnaudbroes:
Issue #1757

## Proposed changes

'Media' can be confusing so we're changing the post label to 'Media / Attachments' in our UI to make it more clear.

## Types of changes

## Checklist

## Testing instructions

Check all modules to make sure that we no longer use 'Media' but use 'Media / Attachments' instead (General Settings, Social Meta, XML Sitemap, Video Sitemap, Importer/Exporter). 

Also, uncheck/check the checkboxes for 'Media / Attachments' and make sure that the options are still being saved to the database.

## Further comments

None.


michaeltorbert:
My $.02 for testing... we need to make sure this is just changing it cosmetically and not anything in the database.

wpsmort:
This is tested and is good.  I have confirmed that this only affects the option labels, not the settings in the database.

Issue #1688, opened 2018-05-13
Notice: Undefined variable: tag in aioseop_opengraph.php


nikkiskene:
I am looking to resolve the following issue when I activate the all-in-one-seo-pack.  The following line is displayed on the header of my page:

Notice: Undefined variable: tag in /home/content/n3pnexwpnas04_data01/49/41464749/html/wp-content/plugins/all-in-one-seo-pack/modules/aioseop_opengraph.php on line 1204

michaeltorbert:
@nikkiskene Thanks for reporting this. We'll take care of it in a future release. 
In the meantime, as it's only a notice rather than an error, you can safely ignore it. Make sure to not have debug logging turned on for your live site.

michaeltorbert:
Reported again here: https://wordpress.org/support/topic/undefined-variable-error-in-aioseop_opengraph-php/

arnaudbroes:
I have been able to reproduce this with these settings:

![image](https://user-images.githubusercontent.com/18156697/57648178-cb6b6980-7592-11e9-9a1d-d45e2f2ea67f.png)

![image](https://user-images.githubusercontent.com/18156697/57648145-b4c51280-7592-11e9-9cae-1cc4087c875a.png)

PR #2427, opened 2019-05-13
 Fix undefined $tag in aioseop_opengraph.php

arnaudbroes:
Issue #1688

## Proposed changes

Fixes a user-reported bug where the 'tag' variable isn't defined in aioseop_opengraph.php. 

This happens when "Automatically Generate Article Tags" in the Social Meta menu is enabled but "Use Keywords In Article Tags" isn't.

## Types of changes

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

## Testing instructions

1. Go to the General Settings menu and enable Meta Keywords.
2. Create a new Post, assign it a Category, Tag and give it one or more meta keywords.
3. Go to the Social Meta menu and enable  "Automatically Generate Article Tags" but disable "Use Keywords In Article Tags".
4. Enable WP_DEBUG in the wp-config file.
5. View the Post and make sure that the "undefined tag variable in aioseop_opengraph.php" notice is no longer present.

## Further comments

You can reproduce the notice using the same steps and the current release of the free version.


wpsmort:
I've tested this and it is good.  

I did find a bug with the incorrect output of the article:tags meta tag and opened issue #2430, however, this bug is not related to this PR and exists in the current version of the plugin.

modules/aioseop_opengraph.php
```diff
@@ -1342,6 +1342,7 @@ function add_meta() {
 
 			if ( $type === 'article' && ! empty( $post ) && is_singular() ) {
 				if ( ! empty( $this->options['aiosp_opengraph_gen_tags'] ) ) {
+					$tag = '';
```
EkoJR: This will need to be moved higher up in the function.

1) it could potentually overwrite a variable above https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/03ea3f9130ccfa95dc3d9b96b8be7c696d0d195a/modules/aioseop_opengraph.php#L1274
2) and there would still be an undefined error below with https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/03ea3f9130ccfa95dc3d9b96b8be7c696d0d195a/modules/aioseop_opengraph.php#L1371

You might have to move it to the top of the function where all the other variables are being set.

arnaudbroes: You are right; good catch!

arnaudbroes:
Comments have been addressed so this can be tested again;

wpsmort:
This is tested and is good.

Issue #2405, opened 2019-05-09
Make WC upgrade button orange


michaeltorbert:
<img width="1098" alt="Screen Shot 2019-05-09 at 12 19 57 PM" src="https://user-images.githubusercontent.com/17364998/57474615-8973b800-7260-11e9-9d84-fe73671104ce.png">

It should be the same orange as the others, which I think is the orange from here but whoever works on this will want to doublecheck.
 https://make.wordpress.org/design/handbook/design-guide/foundations/colors/

michaeltorbert:
This is blue again...

<img width="992" alt="Screen Shot 2019-05-30 at 9 38 17 PM" src="https://user-images.githubusercontent.com/17364998/58675689-6a2def00-8323-11e9-9449-bbe4ff2d1571.png">

PR #2431, opened 2019-05-13
Make WC upgrade button orange

arnaudbroes:
Issue #2405

## Proposed changes

Makes the Upgrade button for the WooCommerce notice orange. Also affects Upgrade button on the Welcome screen as I've improved the styling a bit so it looks less like a flat pancake.

## Types of changes

## Checklist

## Testing instructions

Check the Upgrade button both on the Welcome screen and in the WooCommerce notice on a new* website and make sure it looks good. 

_*on an existing website the notice might have been generated already which is why the Upgrade button will be missing the .button-orange class._

## Further comments

None.


wpsmort:
This is tested and is good.

Issue #2435, opened 2019-05-14
Improve help text and formatting for tooltips


arnaudbroes:
> After we've improved the look, we should also go through all tooltips and add line breaks or other formatting where needed.

The tooltips for title format fields could look like this for example:

![image](https://user-images.githubusercontent.com/18156697/57723145-02ef1a00-7656-11e9-9c3d-f5707474c4fe.png)

PR #2443, opened 2019-05-15
Improve help text for new tooltips

arnaudbroes:
Issue #2435

## Proposed changes

This PR significantly improves the look of our help text for title format fields.

I've also made two minor changes to two other help text strings, namely for the "Create RSS Sitemap" setting and "Rewrite Titles" setting, and removed the %category% macro in the help text for the Post Title Format field since it's deprecated.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

You should look at all tooltips for the Title Format fields and make sure they look good.
You should also look at the tooltips for the "Create RSS Sitemap" and "Rewrite Titles" setting.

## Further comments

For the macros within the title format field tooltips, we now use HTML description lists - https://www.w3schools.com/tags/tag_dt.asp (courtesy of @EkoJR).

I'd also like to improve all help text in the General Settings menu by removing all the "Check this option to" and "This option does" parts but that's something to look at later.


wpsmort:
@arnaudbroes There are some errors and changes needed here as follows:

-  The help text for the %cf_fieldname% macro should read "The name of a custom field".  This applies to Home Page, Page, Post and Description format fields
- The help text for the %category_title% macro should read "The original title of the category".  This applies to Category Title Format
- Should the macros for %day% and %year% in Date Archive Title Format also say (localized)?
- There should be a space after "Meta Descriptions." in the opening paragraph for the Description Format
- The text for the Title Format fields for custom post types is messed up after %post_title%.  See screenshot below
- Is it possible to output the actual name of the post type in the help text.  I ask because in some of the help text it says page and in other parts it says post.  See screenshot below

Screen Shot 2019-05-15 at 3 43 50 PM

arnaudbroes:
@wpsmort:

> The help text for the %cf_fieldname% macro should read "The name of a custom field". This applies to Home Page, Page, Post and Description format fields.

This has been taken care of.

> The help text for the %category_title% macro should read "The original title of the category". This applies to Category Title Format.

This has been taken care of.

> Should the macros for %day% and %year% in Date Archive Title Format also say (localized)?

No, because those output numbers. Good question though!

> There should be a space after "Meta Descriptions." in the opening paragraph for the Description Format.

This has been taken care of.

> The text for the Title Format fields for custom post types is messed up after %post_title%. See screenshot.

This has been taken care of, as well as the bug we found where non-public taxonomies with non-unique names were also being added as macros, in particular the ones from WooCommerce.

> Is it possible to output the actual name of the post type in the help text. I ask because in some of the help text it says page and in other parts it says post. 

I've done this not only for custom posts and taxonomies, but also for the other "static" title format fields since it looks a bit better and those strings have to be retranslated now anyway, so we might as well. 

I've optimized the strings as much as possible in order to reduce the amount of strings that have to be translated (again).

wpsmort:
@arnaudbroes There are some more changes needed to the hep text as follows:

- Post Title Format:  %category_title% should read "The (main) Category of the Post"
- I preferred it when the text "(localized)" was at the end of the help text rather than in the middle.  It make it easier to read the help text.  For consistency, there's one instance in Description Format where (localized) appears at the end instead of in the middle
- The code to get the post types for the help text in the Content Types section is getting the post type name which in many cases will be the plural name.  We should get the singular name, i.e. Movie instead of Movies (see example here - https://developer.wordpress.org/reference/functions/get_post_types/)

arnaudbroes:
> Post Title Format: %category_title% should read "The (main) Category of the Post".

I took care of this.

> I preferred it when the text "(localized)" was at the end of the help text rather than in the middle. It make it easier to read the help text. For consistency, there's one instance in Description Format where (localized) appears at the end instead of in the middle.

I've taken care of the instance in the Description Format helptext but also moved "(localized)" back to the end of the string.

> The code to get the post types for the help text in the Content Types section is getting the post type name which in many cases will be the plural name. We should get the singular name, i.e. Movie instead of Movies (see example here - https://developer.wordpress.org/reference/functions/get_post_types/).

I've checked this and get_post_types is returning the singular names so no changes needed. I've confirmed it not only by checking the variables in the code but also by creating a custom post type "Movie" with "Movie" as the singular name and "Movies" as the plural name.

image

image

wpsmort:
@arnaudbroes As discussed, this is still using $post_type for the post type name in the help text rather than singular_name.

arnaudbroes:
@wpsmort I've now made sure to use the singular_name attribute. :-)

wpsmort:
This is tested and is good.

arnaudbroes:
If anyone is wondering like @EkoJR why I'm using sprintf() in strings where the value does not change, here's the reasoning behind it:

I took words out in a way so that the amount of strings that have to be translated is much smaller, as our locales would otherwise be taking quite a hit.

Consider the following string:

```
sprintf(
    __( 'The %1$s when the %2$s was published (localized)', 'all-in-one-seo-pack' ), 
    __( 'date', 'all-in-one-seo-pack' ),
    /* translators: "Post/Page" are the two main content types in WordPress. */
    __( 'Post/Page', 'all-in-one-seo-pack' )
```

If I wouldn't have added those placeholders, 9 different combinations (where most of the words are the same) would have to be translated:

1. The date when the Post was published (localized)
2. The year when the Post was published (localized)
3. The month when the Post was published (localized)
4. The date when the Page was published (localized)
5. The year when the Page was published (localized)
6. The month when the Page was published (localized)
7. The date when the Post/Page was published (localized)
8. The year when the Post/Page was published (localized)
9. The month when the Post/Page was published (localized)

Using placeholders cuts it back to one sentence "`The %1$s when the %2$s was published (localized)`", and three short strings, "`Post`", "`Page`", and "`Post/Page`". However, "`Post`" and "`Page`" should have already been translated in WordPress Core so it's effectively only two new strings that have to be translated.

admin/class-aioseop-helper.php
```diff
@@ -129,300 +129,239 @@ private function help_text_general() {
...
+					'<dt>%site_title%</dt>' .
+					'<dd>' . __( 'Your site title', 'all-in-one-seo-pack' ) . '</dd>' .
+					'<dt>%site_description%</dt>' .
+					'<dd>' . __( 'Your site description', 'all-in-one-seo-pack' ) . '</dd>' .
+					'<dt>%page_title%</dt>' .
+					/* translators: %s is replaced with a content type such as Post, Page, etc. */
+					'<dd>' . sprintf( __( 'The original title of the %s', 'all-in-one-seo-pack' ), __( 'Homepage', 'all-in-one-seo-pack' ) ) . '</dd>' .
+					'<dt>%page_author_login%</dt>' .
+					/* translators: Example sentence: "The first name of the author of the Post" */
+					'<dd>' . sprintf( __( 'The %1$s of the author of the %2$s', 'all-in-one-seo-pack' ), __( 'username', 'all-in-one-seo-pack' ), __( 'Homepage', 'all-in-one-seo-pack' ) ) . '</dd>' .
```
EkoJR: These should probably be new lines since they extend too far; I think 120 is usually the character limit

admin/class-aioseop-helper.php
```diff
@@ -129,300 +129,239 @@ private function help_text_general() {
...
+					'<dt>%page_author_nicename%</dt>' .
+					'<dd>' . sprintf(
+						__( 'The %1$s of the author of the %2$s', 'all-in-one-seo-pack' ),
+						/* translators: The "nicename" is the sanitized version of a username. */
+						__( 'nicename', 'all-in-one-seo-pack' ), __( 'Homepage', 'all-in-one-seo-pack' )
```
EkoJR: New line for 'Homepage'

admin/class-aioseop-helper.php
```diff
@@ -129,300 +129,239 @@ private function help_text_general() {
...
+					'<dd>' . sprintf( __( 'The original title of the %s', 'all-in-one-seo-pack' ), __( 'Post', 'all-in-one-seo-pack' ) ) . '</dd>' .
+					'<dt>%wp_title%</dt>' .
+					'<dd>' . __( 'The original WordPress title', 'all-in-one-seo-pack' ) . '</dd>' .
+					'<dt>%current_date%</dt>' .
+					'<dd>' . sprintf( __( 'The current %s (localized)', 'all-in-one-seo-pack' ), __( 'date', 'all-in-one-seo-pack' ) ) . '</dd>' .
+					'<dt>%current_year%</dt>' .
+					'<dd>' . sprintf( __( 'The current %s', 'all-in-one-seo-pack' ), __( 'year', 'all-in-one-seo-pack' ) ) . '</dd>' .
+					'<dt>%post_date%</dt>' .
+					'<dd>' . sprintf(
+						__( 'The %1$s when the %2$s was published (localized)', 'all-in-one-seo-pack' ), __( 'date', 'all-in-one-seo-pack' ),
```
EkoJR: For readability, could you add a new line for the translated string 'date'.

css/aiosp_admin.css
```diff
@@ -98,3 +98,7 @@ label[for=aioseop_edit_profile_header]
 	color: #0073AA;
 	text-decoration: none;
 }
+
+.ui-tooltip.ui-widget.ui-corner-all.ui-widget-content.aioseop-ui-tooltip dt {
+	font-weight: bold;
+}
```
EkoJR: EOF error. Needs new line.

arnaudbroes: Done.

arnaudbroes:
@EkoJR I added a new line at the end of the CSS file. As far as readibility goes, I'm not going to make any changes there as discussed on Slack.

Issue #2469, opened 2019-05-22
Rename clean_url() to avoid false deprecation positives 


wpsmort:
When running the plugin files through https://wpseek.com/pluginfilecheck/ this deprecated function was identified:

clean_url is deprecated since WordPress 3.0.0
esc_url() all-in-one-seo-pack-3.0/modules/aioseop_sitemap.php

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/53b0a5f9c74941979d645caa4866145287f7e944/modules/aioseop_sitemap.php#L3387

(edited by Michael to add link to code)

michaeltorbert:
This is actually our own internal function, not WP's. We can just rename it to avoid false positives.

PR #2486, opened 2019-05-24
rename clean_url to aioseop_clean_url 

michaeltorbert:
Issue #2469 

## Proposed changes

We have an internal function, clean_url(), which causes false positives in tools thinking its deprecated.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- There's not too much to test, but you can take a peek at the code to identify which functionality this function is used on.


arnaudbroes:
This looks good and Grunt returns no errors.

Issue #2480, opened 2019-05-23
fix filter hook parameters


michaeltorbert:
In https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/7c504495b31e1bfac7e3bb6fdc3b2d8d2bb99be8 10, 1 shouldn't be there.

PR #2445, opened 2019-05-15
Add filter hook for adding custom title formatting 

michaeltorbert:
Issue #942

## Proposed changes

Users and developers of other plugins have requested the ability to programmatically add their own title template tags.
I've also added some action hooks at the beginning and end of each relevant function.

## Types of changes

What types of changes does your code introduce?

- Adds new API hooks
- Improves existing functionality

## Checklist

- [x] I've selected the appropriate branch (ie x.y rather than master).
- [x] I've created an appropriate title, descriptive of what the PR does.
- [x] Travis-ci runs with no errors.
- [x] I have added tests that prove my fix is effective/my feature works or have created an issue for it (if appropriate). #2447
- [x] I have added necessary documentation, or have created an issue for docs (if appropriate). #2446

## Testing instructions
- You'll want to peak at the code for a general overview of what's happening.
- You can find examples here: https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2446#issue-444640702
- Make sure that all titles are as expected with and without the use of this filter. 
- I'm pretty sure this only applies to post (type) titles, and probably the homepage as well. It likely doesn't work for taxonomy titles.

## Further comments

Tester, code reviewer, etc... feel free to suggest a better filter name if you have one.

PR #2490, opened 2019-05-24
remove extra filter hook arguments

michaeltorbert:
Issue #2480 

## Proposed changes

In #2445 we added some new filter hooks. I inadvertently added some arguments that shouldn't be there. 

## Types of changes

- Removing old code/functionality
- Improves existing code

## Checklist

## Testing instructions
- There's not much to really test here other than that the filters still work. 

## Further comments

It worked before, and this shouldn't break anything. We're just removing two arguments that shouldn't have been there.


arnaudbroes:
This looks good.

Issue #2456, opened 2019-05-16
Remove newline from translatable string


michaeltorbert:
There should be a newline, it just shouldn't be translatable. Either use a placeholder or concatenate earlier.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/53b0a5f9c74941979d645caa4866145287f7e944/admin/aioseop_module_class.php#L1238

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/53b0a5f9c74941979d645caa4866145287f7e944/admin/aioseop_module_class.php#L1243

and many more

PR #2491, opened 2019-05-24
extract newline out of translatable string

michaeltorbert:
Issue #2456 

## Proposed changes

We should remove \n from translatable strings so that it doesn't confuse translators who may remove them.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Take a peak at the code to see what triggers the strings to show up, and make sure when they display the newline comes along with it as before.

## Further comments

This isn't necessarily every location in the entire plugin, but it was the strings we modified in this release (and a few others in the same file).


arnaudbroes:
This looks good. Had to mess around with chmod to change folder permissions in order to reproduce these error notices:

image

Issue #2546, opened 2019-06-01
Content not being correctly excluded from sitemap


wpsmort:
**Describe the bug**
In PR #2317 we added support for excluding content from the sitemap when NOINDEX is enabled.  The bug here is that individual posts set to NOINDEX or Exclude from Sitemap are not being excluded from the sitemap when Indexes is enabled and the sitemap is being paginated.  

**To Reproduce**
1. On a new site, install the master branch and enable the sitemap module
2. Now create a hundred posts in the Uncategorized category
3. Next, edit three posts  as follows:

- One post should be set to Noindex
- One post should be set to Exclude from Sitemap
- One post should be removed from Uncategorized and added to a new category that is excluded from the sitemap using the new Excluded Terms option in the sitemap module

4. View the sitemap, the three posts that have been excluded above should not appear in the sitemap_post.xml
5. Now go to the sitemap settings and set pagination to 50 and view the sitemap
6. You should see two post sitemaps (sitemap_post_1 snd sitemap_post_2). View each one and check to see if the posts set as Noindex and Exclude from Sitemap are excluded.  You should see that the only post that is excluded is the one assigned to the category that has been excluded using Excluded Terms

Additionally, if you create three new posts and make one Noindex and another Exclude from Sitemap, all three show up in sitemap_post_3. 

**Expected behavior**
Posts set to Noindex or Exclude from Sitemap should be excluded from the sitemap regardless of the pagination setting.  However, posts are not excluded when pagination is set to a number lower than the number of published posts.

**Screenshots**
<img width="749" alt="Screen Shot 2019-06-01 at 1 28 13 PM" src="https://user-images.githubusercontent.com/17548525/58751799-ba1acc00-8471-11e9-8184-9d1c568a56df.png">

<img width="976" alt="Screen Shot 2019-06-01 at 1 28 35 PM" src="https://user-images.githubusercontent.com/17548525/58751802-c0a94380-8471-11e9-9d35-844434c5aa36.png">

**Version Info:**
 - WordPress version: 5.2.1

 - All in One SEO Pack version:
 - PHP version: 7.x

This is a new bug as it affects a new feature added in v3.0.

michaeltorbert:
When I install the original PR where this should have been fixed, I'm still getting the URLs included in the sitemap. https://github.com/semperfiwebdesign/all-in-one-seo-pack/tree/04899b3f30e65956b35aac93b16a5c58893ff51d

[...]

PR #2551, opened 2019-06-03
Fix sitemap not excluding posts properly with max_posts setting

EkoJR:
Issue #2546

## Proposed changes

Fixes posts not being excluded properly when the max posts is set when Sitemap Indexes is enabled

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

1. On a new site, install the master branch and enable the sitemap module
2. Now create a hundred posts in the Uncategorized category
3. Next, edit three posts  as follows:

- One post should be set to Noindex
- One post should be set to Exclude from Sitemap
- One post should be removed from Uncategorized and added to a new category that is excluded from the sitemap using the new Excluded Terms option in the sitemap module

4. View the sitemap, the three posts that have been excluded above should not appear in the sitemap_post.xml
5. Now go to the sitemap settings and set pagination to 50 and view the sitemap
6. You should see two post sitemaps (sitemap_post_1 snd sitemap_post_2). View each one and check to see if the posts set as Noindex and Exclude from Sitemap are excluded.  You should see that the only post that is excluded is the one assigned to the category that has been excluded using Excluded Terms

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -4369,7 +4369,8 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 				),
 			);
 			$ex_args['fields']         = 'ids';
-			$ex_args['posts_per_page'] = $this->max_posts;
+			// This needs to be -1 so that excluding posts isn't restricted to affect posts to not be excluded properly.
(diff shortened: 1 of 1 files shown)


wpsmort:
I have tested this and confirmed that it fixes the bug.

Issue #2569, opened 2019-06-05
.ui-sortable-handle span {} affects other wp_admin meta boxes


arnaudbroes:
Reported on 6/5/2019 via e-mail by Ilkka V. 

> There's also other things that ruins the WP admin styles:
> I.E: .ui-sortable-handle span {
> ...
> }
> The CSS rules are too global!

I don't know an example of a plugin this would be conflicting with but I do see the issue so I'll restrict the class to our own options (which I should have done in the first place).

In aioseop_module.css.

PR #2570, opened 2019-06-05
Restrict .ui-sortable-handle span {} to aioseop options

arnaudbroes:
Issue #2569.

## Proposed changes

Restricts the .ui-sortable-handle span {} class to our own options as this could affect other plugin's meta boxes in wp_admin.

## Types of changes

## Checklist

## Testing instructions

Check all menus and make sure that the header of each section still has the same style that was introduced in AIOSEOP v3.0. Also, have a look at the code so you understand the change that was made.


wpsmort:
This is tested and is good.

Issue #2572, opened 2019-06-05
Missing social tab in 3.0.1


michaeltorbert:
The Main Settings and Social Settings tabs are missing:

<img width="979" alt="Screen Shot 2019-06-05 at 2 02 53 PM" src="https://user-images.githubusercontent.com/17548525/58978908-db91f580-879a-11e9-92dd-d5511617df14.png">

wpsmort:
I tracked this down to a conflict with this plugin - https://wordpress.org/plugins/accelerated-mobile-pages/

I was able to reproduce this on a new site with only AIO and that plugin installed.  The Accelerated Mobile Pages plugin has a CSS style of display: none; applied to the class .hide on line 1038 of accelerated-mobile-pages/pagebuilder/inc/admin-amp-page-builder.css

We use the class .hide on our tabs here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/ce0acebb6efd9d30c6547049bddff1c6a6e31b51/aioseop_class.php#L4334

The commit that caused this is here https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2290/commits/0dade54d6cdea0283f19321fad3dfbbe200b9c28 where we removed the style .aioseop_header_tabs.hide

PR #2577, opened 2019-06-05
Fix missing tabs when .hide is used by other plugins

wpsmort:
Issue #2572

## Proposed changes
This fixes a conflict where other plugins use the .hide class by adding back styles removed in v3.0

## Types of changes

## Checklist

## Testing instructions
- Install and activate AIO v3.0 and this plugin - https://wordpress.org/plugins/accelerated-mobile-pages/
- Activate the Social Meta module
- Go to Posts > Add New and you should see that the Main Settings and Social Settings tabs are missing
- Now install this PR and verify the tabs are back

## Further comments
@EkoJR When testing, please check whether adding this style back affects anything else.  I don't know what the reason was for removing the style in the first place.

### DIFF
--- a/css/modules/aioseop_module.css
+++ b/css/modules/aioseop_module.css
@@ -1217,6 +1217,11 @@ div.aioseop_notice a.aioseop_dismiss_link {
 	width: 98%;
 }
 
+.aioseop_tabs.hide,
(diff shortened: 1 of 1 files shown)


arnaudbroes:
This looks good.

Issue #2580, opened 2019-06-05
Fatal error: Call to undefined method DirectoryIterator::getExtension() in class-aioseop-notices.php on line 159


michaeltorbert:
Reported by Jim via email:

> Received this error today and was curious if you've received similar fatal error messages

`Fatal error: Call to undefined method DirectoryIterator::getExtension() in /home/content/70/9001570/html/wp-content/plugins/all-in-one-seo-pack/admin/class-aioseop-notices.php on line 159`

michaeltorbert:
This is caused by our use of the DirectoryIterator, which requires PHP 5.3.6.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/3.0/admin/class-aioseop-notices.php#L158

PR #2581, opened 2019-06-05
Only show notices for PHP > 5.3.6

michaeltorbert:
Issue #2580 

## Proposed changes

In 3.0, we introduced a new notices class. This class makes use of DirectoryIterator->getExtension which wasn't introduced until PHP 5.3.6.

## Types of changes

-

## Checklist

## Testing instructions
- On a fresh site, install WordPress < 5.2 (WordPress 5.2 requires PHP > 5.6)
- Install PHP 5.2.4
- Install All in One SEO Pack
- Verify the fatal error
- Then verify this PR fixes it.
- Bonus: perform this same test when upgrading from an earlier version of AIOSEOP (pre-3.0)

## Further comments

This will make notices not work at all if they're running PHP < 5.3.6, but there won't be a fatal error.

### DIFF
--- a/admin/class-aioseop-notices.php
+++ b/admin/class-aioseop-notices.php
@@ -112,6 +112,12 @@ if ( ! class_exists( 'AIOSEOP_Notices' ) ) {
 		 * @since 3.0
 		 */
 		public function __construct() {
+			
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good.

Issue #2583, opened 2019-06-06
Restrict WooCommerce notice to our own screens


michaeltorbert:
Since there are still some users with the problem of it not staying away (and I'd still like us to figure out why) we need to discuss whether to at least restrict it to our own admin screens. This may be the more considerate option anyway.

Thoughts?

arnaudbroes:
Is what I would have done. I'd either limit it to only AIOSEOP or AIOSEOP and the Edit Product & Edit Product Category screens.

wpsmort:
I agree with restricting the notice to just our admin screens. As you say, it's the considerate option.

PR #2590, opened 2019-06-06
restrict WC notice to our own screens 

michaeltorbert:
Issue #2583

## Proposed changes

Because a handful of users are experiencing problems with getting a dismissed notice to stay dismissed, we should at least restrict it to our own screens, which is the more considerate option anyway, and probably what we should have done from the start.

## Types of changes

## Checklist

## Testing instructions
- Install both AIOSEOP and WooCommerce to trigger the notice.
- Bonus: test on a site having the persistence problems.


wpsmort:
This is tested and is good.

arnaudbroes:
This is approved.

Issue #2589, opened 2019-06-06
Remove review request notice


michaeltorbert:
Since we're having the problem of dismissible notices staying gone, we probably need to preemptively disable this one until the problem is resolved.

PR #2591, opened 2019-06-06
disable review request notice 

michaeltorbert:
Issue #2589

## Proposed changes

Since some users are still having problems with dismissible notices not staying dismissed, we should temporarily disable this one from happening.

## Types of changes

- Removing old code/functionality (actually it just comments out)

## Checklist

## Testing instructions
- Make sure you can reliably trigger the notice.
- Apply this patch and make sure the notice no longer appears and that there are no messages in the error log.

## Further comments

While it would be super nice if more users would leave us 5 star reviews, we probably don't want to request it with a notice that (unintentionally) can't be dismissed. We never want to give users a bad experience, but we really really shouldn't give them a bad experience while asking for a review. 
:)

### DIFF
--- a/admin/class-aioseop-notices.php
+++ b/admin/class-aioseop-notices.php
@@ -115,7 +115,7 @@ if ( ! class_exists( 'AIOSEOP_Notices' ) ) {
 
 			// DirectoryIterator::getExtension() was added in PHP 5.3.6. We can remove this once we drop support < PHP 5.3.
 			if ( version_compare( phpversion(), '5.3.6', '<' ) ) {
-			    return false;
+				return false;
 			}
 
 			$this->_requires();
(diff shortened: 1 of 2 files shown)


arnaudbroes:
For some reason did not only not remove the notice but add an additional blank one - 
image

arnaudbroes:
As far as Grunt and Travis are concerned, I think Michael's branch is a little bit outdated, so those checks can be ignored. Grunt for instances is reporting three errors in a file he didn't change.

michaeltorbert:
@arnaudbroes  The branch shouldn’t be outdated. I just created it, and I’m pretty sure has all the changes for grunt in it.

michaeltorbert:
@arnaudbroes Good catch. Initially I was able to reproduce what you're seeing but not anymore. Can you see if you can figure out some conclusive steps to reproduce?

@EkoJR Can you help out with the best way to disable a notice?

michaeltorbert:
Until Ben comes and takes a look to see if there's a better way, my latest commit *should keep the review notice from ever showing up.

EkoJR:
Closing to https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2594

Issue #2576, opened 2019-06-05
Bug - Redirect to about page is breaking following updates


Zodiac1978:
/wp-admin/index.php?page=aioseop-about

**Describe the bug**
In a bulk update from AIOSEOP and other plugins, the updates following after AIOSEOP are failing because of the redirect to the About page.

Related #2255

**To Reproduce**
1. Go to Plugins
2. Mark the plugins to update or click update link from AIOSEOP and at least one more plugin
3. See the second one failing - error message is showing the about page

**Expected behavior**
Updates shouldn't fail.

**Screenshots**
<img width="1223" alt="aioseop" src="https://user-images.githubusercontent.com/870475/58968323-2705cd00-87b6-11e9-91e3-76f02d95cda6.png">

Zodiac1978:
![aioseop](https://user-images.githubusercontent.com/870475/58968479-66ccb480-87b6-11e9-8ab3-abc8bc8432ae.gif)

michaeltorbert:
@Zodiac1978 Thanks for reporting this, and THANKS for the great steps to reproduce + animation. 
We're looking into it now.

arnaudbroes:
I'm not able to reproduce this issue. Here's what I did:

1. Install AIOSEOP, Classic Editor, Contact Form (same plugins as user);
2. Change version number of each plugin so they can be updated.
3. Go to the database and change the value of 'aioseop_seen_about_page' in the user meta table to an earlier version.

![image](https://user-images.githubusercontent.com/18156697/58989231-cbeacf00-87e3-11e9-90b3-9e8ea1a6380a.png)

michaeltorbert:
Hi @Zodiac1978 

So far we've been unable to reproduce this. You can submit your login info to our secure form here and we'll be happy to investigate further so we can get out a fix as quickly as possible: https://semperplugins.com/contact/support-login-information/

wpsmort:
I have been able to reproduce this on a local test site.  Here are the steps to reproduce this:

1. Install old versions of multiple plugins (NOTE: the plugins should have names that place them after All in One SEO Pack in the alphabetical list on the Plugins screen)
2. Install v2.12.1 of AIOSEOP
3. Go to the Plugins screen an verify that all plugins require updates
4. Select all of the plugins and then select Update in the Bulk Actions drop down and click Apply. See screenshot below.

[...]

PR #2593, opened 2019-06-06
Don't start welcome screen if we're in Ajax

michaeltorbert:
Issue #2576 

## Proposed changes

The bug is that our Welcome Screen stuff injects itself in during a bulk-update (sometimes) and causes a one-time failure for a subsequent plugin updating.
I've put in checks for whether we're in ajax.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Testing is tricky.
- Follow the steps in the issue to reproduce the bug. Make sure you can reliably reproduce it. https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2576#issuecomment-499249701
- The problem is, when update testing, you're getting the 3.0.2 code from .org instead of the patch. We'll need to figure out a way to intercept that for testing purposes.

### DIFF
--- a/inc/aioseop_updates_class.php
+++ b/inc/aioseop_updates_class.php
@@ -72,13 +72,18 @@ class AIOSEOP_Updates {
 		$this->do_feature_updates();
 	}
 
+	/**
+	 * Calls the Welcome Screen.
+	 *
+	 * since 2.3.9.2
+	 * since 3.0 Don't fire if we're doing Ajax or not in the admin.
+	 */
 	function aioseop_welcome() {
-		if ( get_transient( '_aioseop_activation_redirect' ) ) {
+		if ( ( ! DOING_AJAX || ! defined( 'DOING_AJAX' ) ) && is_admin() && get_transient( '_aioseop_activation_redirect' ) ) {
 			delete_transient( '_aioseop_activation_redirect' );
 			$aioseop_welcome = new aioseop_welcome();
 			$aioseop_welcome->init( true );
 		}
-
 	}
 
 	/**


EkoJR:
I believe the problem is occuring at `inc/aioseop_functions.php:202` `aioseop_admin_head()`; which is hooked into `admin_head`. That is where some of the text is being printed out.

wpsmort:
I tested this on a new site and an existing site but I get the same error that I reported here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2576#issuecomment-499273942

Screen Shot 2019-06-12 at 1 23 05 PM

michaeltorbert:
@wpsmort That seems like a different error.

michaeltorbert:
@wpsmort How did you test this exactly? Clicking update will pull down the code from .org, not the code from this PR.

wpsmort:
Here's the process I'm using to test which, as you say, is not going to work because its pulling the v3.1 code from .org before updating the other plugins:

1. Install and activate old versions of five plugins on a new site
2. Download the code from this PR, rename the directory to all-in-one-seo-pack and zip it up
3. Install and activate the zipped PR code and verify that the welcome screen is displayed
4. Delete aioseop_seen_about_page from the usermeta table in the database
5. Return to the Plugins screen, select all plugins and then select Update in the dropdown and click Apply
6. The error is displayed after AIO is updated

FYI... If I don't rename the PR code in step 2 then I don't get the error when updating and the update process completes for all plugins.

michaeltorbert:
Closing this until we can figure out a way to reproduce and test outside of .org.

Issue #869, opened 2017-04-07
warning in woocommerce when setting mem in performance to 256 


michaeltorbert:
When you set the memory limit to 256 in our performance module, WooCommerce (/wp-admin/admin.php?page=wc-status&tab=status) warns "40 MB - We recommend setting memory to at least 64MB."

Reported here: https://wordpress.org/support/topic/woocommerce-warning-40-mb-we-recommend-setting-memory-to-at-least-64mb/

I've verified this issue exists. Setting it to any other value makes the warning go away.

michaeltorbert:
This has been reported again by Brandon via email. Even though this seems to be a cosmetic-only issue (ie the memory limit isn't really only 40MB) we should take a look at if and how we can resolve this.

michaeltorbert:
This has been reported again by Brandon via email. Even though this seems to be a cosmetic-only issue (ie the memory limit isn't really only 40MB) we should take a look at if and how we can resolve this.
https://github.com/woocommerce/woocommerce

wpsmort:
I've emailed Mike Jolley at WooCommerce about this.

wpsmort:
Mike responded to ask if we've ruled out that this may be related to the code here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/ef0349e9312809079e484ff356295367ff599c05/all_in_one_seo_pack.php#L197

He said that "The system status report in WooCommerce simply grabs the WP Memory Limit constant here https://github.com/woocommerce/woocommerce/blob/0e80030f0b48ed3f520401b2341c4f3a3319a356/includes/api/v2/class-wc-rest-system-status-v2-controller.php#L600 That reporting is accurate. The line below tries to get the actual memory limit but that relies on certain functions in PHP which may not be available."

I tested and found that WoCommerce only outputs the warning about 40 MB when our Raise memory limit setting in the Performance module is set to 256 MB.  If this is set to any other value then WooCommerce reports the WP memory limit as 256 MB.

michaeltorbert:
The relevant parts of our code are:

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/39c6de463eeca92bbc50729b5746d3ac519780e5/all_in_one_seo_pack.php#L134-L200

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/modules/aioseop_performance.php#L135-L147

[...]

PR #2659, opened 2019-06-15
Fix WooCommerce PHP memory limit warning

arnaudbroes:
Issue #869.

## Proposed changes

Fixes an issue where WooCommerce isn't able to properly parse the PHP memory limit All in One SEO Pack sets.

## Types of changes

## Checklist

## Testing instructions

Check the following two things:

(1) All in One SEO Pack is still able to set the PHP memory limit.
(2) The warning on WooCommerce's status page has disappeared.


wpsmort:
This is tested and is good.

Issue #2609, opened 2019-06-07
Make plugin name constant dynamic


michaeltorbert:
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/01bf171799b6b45443d0b7f4b3de2c61256b7bdf/all_in_one_seo_pack.php#L41-L42

We should just check for the AIOSEOPPRO constant and then conditionally call it whichever to make switching easier.

If Pro - name = pro
else - name = free/core

arnaudbroes:
That would be a welcome change.

PR #2661, opened 2019-06-15
Make plugin name constant dynamic

arnaudbroes:
Issue #2609.

## Proposed changes

Makes the plugin name constant dynamic so we don't have to change it each time we build Pro.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
 
Check both free and Pro to make sure that everything looks good and check a tooltip (e.g. "Register imortant events" setting in General Settings menu) to see if the plugin name is correct.


wpsmort:
This is tested and is good.

Issue #1798, opened 2018-07-19
Undefined variable: post_type in aioseop_class.php


michaeltorbert:
`Undefined variable: post_type in <b>/var/www/fournais-bender.dk/public_html/wp-content/plugins/all-in-one-seo-pack-pro/aioseop_class.php</b> on line <b>4155</b>`

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/aioseop_class.php#L4159

Reported by Henriette via email. 

For reproducing... on the customer's site, it's only showing up on the _search results_ page.

wpsmort:
@jguerine Can you please send us an Administrator login for your site using the secure form below so we can troubleshoot this:

https://semperplugins.com/contact/support-login-information/

michaeltorbert:
Reported for category archive pages: https://wordpress.org/support/topic/notice-undefined-variable-error-on-archive-page/

awgd:
If you intended to use $post_type on line 4159 of the aioseop_class then you will need to include 'global $post_type;' somewhere before it to be able to use it. Since the comment above it says it is duplicated to prevent stepping on toes, not sure if it's even needed or not.

michaeltorbert:
This code was added in https://github.com/semperfiwebdesign/all-in-one-seo-pack/commit/ebf2c686fa106e9d0aa908ec68d12ef297780b17 by @contactashish13

contactashish13:
to reproduce:

1. Default to NOFOLLOW: checked for posts
2. Use noindex for Categories: checked
3. Go to category archive page

PR: https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1825

wpsmort:
@contactashish13 I tested this PR by following your steps above.  Whilst this does fix the undefined variable notice it incorrectly changes the nofollow meta tag on category archives.  It changes the nofollow to match the setting for Posts which is wrong.

jguerine:
I checked on my website.

It appeared on the 404 pages, for which I had ticked "noindex" in the AIO options. 
I unticked that option for the 404 pages, and the error disappear.

I ticked again, and the error reappear.

It does indeed seems (to me) that it would come from that!

contactashish13:
[...]

PR #2671, opened 2019-06-17
Refactor get_robots_meta()

arnaudbroes:
Issue #1798

## Proposed changes

Fixes the following notice "undefined variable: post_type in aioseop_class.php". This happens on the category archive page when Posts are set to NOFOLLOW globally and categories are set to NOINDEX globally.

## Types of changes

## Checklist

## Testing instructions

First, in order to reproduce, do these steps:

1. Go to the General Settings, make sure Categories are NOINDEX and enable NOFOLLOW for Posts.
2. Create a Post and assign it to a category or leave it uncategorized.
3. View the relevant category page and you'll see the notice.*

*(WP_DEBUG has to be enabled of course).

Then install this PR, make sure that the notice went away and that this didn't break anything else.
You might want to play around with custom posts and taxonomies.


wpsmort:
@arnaudbroes I tested this but it changes the output of the robots meta tag on categories, tags, date and author archives to noindex,nofollow.

This should not affect the follow part of the meta tag.

arnaudbroes:
The fix for this notice uncovered a whole bunch of existing bugs. Since the current code is terrible, I've refactored the function and I'll therefore also rename this PR as the scope of it is now much bigger than just fixing a notice.

FYI, here are some of the changes I made:

* If site-wide noindexing is enabled under Settings > Reading, then All in One SEO Pack does not change the robots meta tag.
* Posts and Taxonomies were interlinked and their nofollow/noindex values were affecting each other, which shouldn't be happening. I've made the checks for both independent from each other.
* I've removed redundant code and improved readability.
* I've changed (no)index and (no)follow string values to bools as it is bad practice to use string values multiple times throughout the code and moved them to a separate helper function. 

@wpsmort you will want to extensively test this to make sure this doesn't break anything and that there are no bugs left. Since this refactors the entire function, you'll also want to go through all existing nofollow/noindex issues and see if this PR fixes any of them.

arnaudbroes:
Fixed an additional bug where date archives, author archives and search pages were using the individual noindex/nofollow setting values of the first post.

wpsmort:
@arnaudbroes This is really great work.  Unfortunately there are two bugs as follows:

- Whatever is set for the latest post affects the homepage as well.  For example, if you set Noindex on the latest post, then the homepage (when displaying posts) will also be set to Noindex.  This is a new bug.

- Noindex and Nofollow are displayed on any paginated archive when Noindex or Nofollow are set as the default for Posts.  This happens on page 2 and greater for the homepage, term archives, date archives and author archives. This is not a new bug.

Let me know if you need help reproducing these.

arnaudbroes:
@wpsmort I believe I fixed both of those bugs as well as the fatal error that was occurring with PHP 5.4. Please test again and let me know if you find anything else.

wpsmort:
This is tested and is good.

aioseop_class.php
```diff
@@ -4301,153 +4305,134 @@ function apply_description_format( $description, $post = null ) {
...
+		$tax_noindex    = '';
+		$aiosp_noindex  = '';
+		$aiosp_nofollow = '';
+
+		if (
+				! empty( $opts ) &&
+				! is_date() &&
+				! is_author() &&
+				! is_search() &&
+				! is_home()
```
EkoJR: Would these extra conditions restrict the base settings for noindex/nofollow from also being applied to categories, tags, and posts/pages?

aioseop_class.php
```diff
@@ -4301,153 +4305,134 @@ function apply_description_format( $description, $post = null ) {
...
-							( ! empty( $aioseop_options['aiosp_paginated_noindex'] ) )
-							&& $page > 1
-						) ||
-						(
-							( '' === $aiosp_noindex ) &&
-							( ! empty( $aioseop_options['aiosp_cpostnoindex'] ) ) &&
-							in_array( $post_type, $aioseop_options['aiosp_cpostnoindex'] )
-						)
-
+			if ( ! $noindex ) {
```
EkoJR: I believe this should be `if ( $noindex ) {`.

aioseop_class.php
```diff
@@ -4301,153 +4305,134 @@ function apply_description_format( $description, $post = null ) {
...
+				if ( 
+						! empty( $aioseop_options['aiosp_paginated_noindex'] ) &&
+						$page > 1 ||
+						'' === $aiosp_noindex &&
+						! empty( $aioseop_options['aiosp_cpostnoindex'] ) &&
+						in_array( $post_type, $aioseop_options['aiosp_cpostnoindex'] )
```
EkoJR: Previously this was wrapped as.

```
(	
	( ! empty( $aioseop_options['aiosp_paginated_noindex'] ) )	
	&& $page > 1	
) ||	
(	
	( '' === $aiosp_noindex ) &&	
	( ! empty( $aioseop_options['aiosp_cpostnoindex'] ) ) &&	
	in_array( $post_type, $aioseop_options['aiosp_cpostnoindex'] )	
	)
```

Was this intended? Or was it suppose to be removed (same check occurs above [line 4346])?

aioseop_class.php
```diff
@@ -4301,153 +4305,134 @@ function apply_description_format( $description, $post = null ) {
...
 				) {
-					$noindex = 'noindex';
+					$noindex = true;
 				}
+			}
+			if ( ! $nofollow ) {
```
EkoJR: I believe this should be `if ( $nofollow ) {`.

aioseop_class.php
arnaudbroes: It should be `! noindex`.

On lines 4346-4351, the code first checks whether the page has been noindexed/nofollowed via our meta box on the Edit screen; so it checks the postmeta values. If it has been, then we don't need to run subsequent checks for either noindex/nofollow (or both - see lines 4361-4364). 

Only if the page hasn't been noindexed/nofollowed individually should we check whether its type has been noindexed/nofollowed globally/site-wide in the General Settings menu.

aioseop_class.php
arnaudbroes: It should be `! nofollow`.

Same explanation as for noindex (see above).

aioseop_class.php
arnaudbroes: Good question. Those do not restrict categories, tags , posts and pages. They are needed for those archive pages because the code otherwise fetches the postmeta values from the first post on the archive page.

aioseop_class.php
arnaudbroes: That was not intended.

```
( ! empty( $aioseop_options['aiosp_paginated_noindex'] ) )	
	&& $page > 1	
)
```

This was moved up to line 4346 and should have been removed here.

arnaudbroes:
@wpsmort I made some small adjustments which shouldn't affect anything, but this has to be retested now as it's a rather big PR.

arnaudbroes:
> This seems a bit more complex than anticipated, and also seems to include more than 1 issue in this PR.

Well, to put it on your own words, this was a bit like opening the box of Pandora so a complete rewrite was needed here. There should be 4 more issues this PR will fix but I'll confirm that after it's been merged - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2698

wpsmort:
This is tested and is good.

aioseop_class.php
```diff
@@ -4301,153 +4305,130 @@ function apply_description_format( $description, $post = null ) {
...
+	 * Helper function for get_robots_meta().
+	 *
+	 * @since 3.2.0
+	 *
+	 * @param bool $noindex
+	 * @param bool $nofollow
+	 *
+	 * @return string
+	 */
+	private function aioseop_get_robots_meta_string( $noindex, $nofollow ) {
```
EkoJR: When adding methods in a class, you don't need to prefix functions/methods.

aioseop_class.php
```diff
@@ -4301,153 +4305,130 @@ function apply_description_format( $description, $post = null ) {
...
+		if ( $noindex ) {
```
EkoJR: Sorry I didn't catch this sooner ( I thought AccessLint also looked for these type of patterns ), but one coding patterns that would improve this is.

```
$index_value  = 'index';
$follow_value = 'follow';

if ( $noindex ) {
	$index_value = 'noindex';
}

if ( $nofollow ) {
	$follow_value = 'nofollow';
}
```

arnaudbroes: Definitely. Not sure why I didn't think of this.

aioseop_class.php
arnaudbroes: Good to know!

arnaudbroes:
@EkoJR this can be reviewed again.

wpsmort:
@arnaudbroes I found two instances where this isn't working as expected:

1. If you have a static Blog page set under Reading Settings then you cannot control Noindex or Nofollow either from the default settings for pages or from the page itself.  I think that expected behavior here should be that if you set Noindex or Nofollow for all pages then the this should affect the Blog page.  Similarly, you should be able to set Noindex or Nofollow on the Blog page itself.

2. Similar to above, the default settings for pages doesn't affect the WooCommerce Shop page.  The settings on the Shop page do work fine though.  This also happens if you set the Shop page as the static homepage under Reading Settings.

I'm not sure how I missed this before so let me know if you cannot reproduce this.

arnaudbroes:
@wpsmort this can be retested.

wpsmort:
@arnaudbroes I tested again and found these bugs:

1. Setting the default for Pages to noindex and/or nofollow affects the static home page.  This was not the case in previous testing and I don't believe that we want this behavior.

2. Settings the default for Pages to noindex and/or nofollow doesn't affect the static blog page.  This is the bug reported in my previous comment.  

3. I can no longer control noindex or nofollow for the WooCommerce Shop page.  Before, I could control it by editing the page, but now I can't control it either on the page or using the default for Pages.

4. Something is causing some categories to be defaulted to noindex, nofollow and I can't seem to change that behavior or figure out where that's coming from but it's definitely due to a change made since this commit https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2671/commits/6ad3b9dccf8b807deb98bb87ebffd8544a01eef8

5. I can no longer override the default noindex / nofollow for a post type by editing an individual page or post.  So if I set the default for Pages to noindex and/or nofollow and then set an individual page to index and/or follow, the page still shows noindex, nofollow.

6. I get the following warning when browsing to a custom taxonomy such as product categories:

PHP Warning:  in_array() expects parameter 2 to be array, string given in all-in-one-seo-pack/aioseop_class.php on line 4506

arnaudbroes:
@wpsmort 

1. We now ignore all noindex/nofollow settings for the homepage.
2. This should be resolved.
3. This should be resolved.
4. I wasn't able to reproduce this but I might have fixed it in (1), (2) or (3). Let me know if this is still happening.
5. This should be resolved.
6. This should be resolved.

wpsmort:
@arnaudbroes There are still a couple of bugs here as follows:

1. The static home page is now defaulting to noindex, nofollow.

2. Item 4 in my previous comments regarding certain categories is still a bug.  I traced it back to this commit https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2671/commits/ac78002cd2b720b2c383df512949fbba0bb561a2 and specifically to this code on line 4576:

```
if ( property_exists( $queried_object, 'term_id' ) ) {
	$meta = get_term_meta( $queried_object->term_id );
}
```

arnaudbroes:
@wpsmort as discussed, (1) has been fixed and (2) was just a configuration issue.

wpsmort:
@arnaudbroes There's one final bug where you can't control noindex and nofollow for the Shop page by editing the page.  The default settings for Pages does work though.

arnaudbroes:
@wpsmort this is fixed. If you look at the code change, you'll see that you don't have to test everything else again; just this since that's the only thing it affects.

wpsmort:
This is tested and is good.

Issue #746, opened 2017-02-09
XML Sitemap - Add a filter to exclude taxonomy terms by ID


wpsmort:
Requested by Hakan Mitts by email on 2/9/2017

User has requested a way to exclude only certain categories by ID from the XML sitemap.  This includes being able to exclude certain tax terms by ID.  For example, exclude only certain product categories.

michaeltorbert:
My recommendation is that this is a filter hook.

contactashish13:
@michaeltorbert @wpsmort I believe this is related to https://github.com/semperfiwebdesign/aioseop-pro/issues/240 and is resolved by https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1419

wpsmort:
@contactashish13 Yes, this is related to https://github.com/semperfiwebdesign/aioseop-pro/issues/240 and I've updated that issue with my comments.  Can you confirm whether there is a filter for this (in addition to the new option) and can you provide an example of how that filter can be used to exclude specific terms from a custom taxonomy.

contactashish13:
@wpsmort please check https://github.com/semperfiwebdesign/aioseop-pro/issues/240#issuecomment-421278189

wpsmort:
@contactashish13 Please provide examples of how to use these filters.

arnaudbroes:
After going through the code, I found we already have a filter for this, but no documentation. 
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/bdd97e94ae953ef85f894da6f4230217d5693223/modules/aioseop_sitemap.php#L4165-L4168

I've removed some unnecessary arguments and created a doc for it.

PR #2672, opened 2019-06-17
Change naming and remove redundant arguments exclude_tax_terms filter

arnaudbroes:
Issue #746

## Proposed changes

Changes name of an existing filter and removes unnecessary arguments.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

Use to code example here to test this filter - https://semperplugins.com/wp-admin/post.php?post=4958&action=edit&lang=en. 

If the filter works, please also review the documentation and publish it.

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -4163,7 +4163,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 				}
 			}
 
-			$args = apply_filters( $this->prefix . 'tax_args', $args, $page, $this->options );
+			$args = apply_filters( 'aioseop_sitemap_exclude_tax_terms', $args );
 
 			return $args;
 		}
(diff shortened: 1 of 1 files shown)


wpsmort:
As discussed, I've assigned this back to remove an extra line of code.

arnaudbroes:
@wpsmort turns out I didn't have to. Confused it with another PR.

wpsmort:
@arnaudbroes I'm getting these two errors when testing this PR.

The warning below is in reference to the line `array_push( $args['exclude'], '16' );` in the function (16 is the ID of my category)"

`PHP Warning:  array_push() expects parameter 1 to be array, null given in functions.php on line 351`

The notice below is in reference to the line `$excluded_terms = $args['exclude'];` in the function:

`PHP Notice:  Undefined index: exclude in functions.php on line 348`

arnaudbroes:
@wpsmort this was happening because of a bug in the original filter where $args['exclude'] wouldn't be instantiated when no terms are excluded in the XML Sitemap menu. I've moved it a few lines up to fix it. Good catch! Please test the filter again with the code example in the documentation.

wpsmort:
This is tested and is good.  I have published the doc here - https://semperplugins.com/documentation/aioseop-sitemap-exclude-tax-terms/

EkoJR:
@arnaudbroes is going to double check his records to see if anyone may have used the old filter name for excluding terms. Although it wasn't documented, there may be a chance the filter is being used by those who came across it; who are knowledgeable in coding.

arnaudbroes:
@EkoJR no one is using it yet so I'm sending it back to you for final approval.

Issue #1614, opened 2018-03-19
Replace parse_url with wp_parse_url


michaeltorbert:
Just for the sake of being up to date, we should replace PHP's parse_url with WP's wp_parse_url, now that AIOSEOP's minimum requirement is WP 4.4.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/410ad40cc9c8a3640ad2be33fb16feab42b908e3/inc/aioseop_functions.php#L821

https://developer.wordpress.org/reference/functions/wp_parse_url/

"PHP 5.4.7 expanded parse_url()’s ability to handle non-absolute url’s, including schemeless and relative url’s with :// in the path. This function works around those limitations providing a standard output on PHP 5.2~5.4+.

Secondly, across various PHP versions, schemeless URLs starting containing a ":" in the query are being handled inconsistently. This function works around those differences as well."

EkoJR:
This is part of the WPCS PR changes that occured from #2084

Sitemap where areas are marked as `// TODO Change to wp_parse_url().`

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L2380

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L2768

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L2772

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L3625

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L3640

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L3648

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a7f01e2289ee6e16a276e529d91f156ab65c6fc3/modules/aioseop_sitemap.php#L3649

[...]

PR #2690, opened 2019-06-25
Replace parse_url with wp_parse_url 

arnaudbroes:
Issue #1614

## Proposed changes

Replace the parse_url() function with the built-in wp_parse_url() function. 

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

Have a look at the code changes and see where this function has been changed. You'll just want to check if everything still works fine, mainly the sitemap.

EDIT: in order to test if schemeless images are still being processed correctly, which is what is_image_url_valid() is meant for, follow these steps:

1. Create three new posts. 

- The first post should have an image tag with an absolute source URL such as "http://example.org/image1.jpg". 
- The second post should have an image tag with a source URL of the following format: "//example.org/image2.jpg".
- The third post should have an image tag with a source URL of the following format: "/image3.jpg".

2. Go to the XML Sitemap and make sure that each of these posts has 1 image in the sitemap.

## Further comments

I checked with Xdebug to make sure that the result is the same for each instance.

### DIFF
--- a/inc/aioseop_functions.php
+++ b/inc/aioseop_functions.php
@@ -919,7 +919,7 @@ if ( ! function_exists( 'aioseop_mrt_exclude_this_page' ) ) {
 			if ( $url === null ) {
 				$url = $_SERVER['REQUEST_URI'];
 			} else {
-				$url = parse_url( $url );
+				$url = wp_parse_url( $url );
 				if ( ! empty( $url['path'] ) ) {
 					$url = $url['path'];
 				} else {
--- a/inc/aiosp_common.php
+++ b/inc/aiosp_common.php
@@ -148,7 +148,7 @@ class aiosp_common {
 		if ( 0 !== strpos( $url, 'http' ) && '/' !== $url ) {
 			if ( 0 === strpos( $url, '//' ) ) {
 				// for //<host>/resource type urls.
-				$scheme = parse_url( home_url(), PHP_URL_SCHEME );
+				$scheme = wp_parse_url( home_url(), PHP_URL_SCHEME );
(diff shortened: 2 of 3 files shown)


wpsmort:
This is tested and is good.

michaeltorbert:
@arnaudbroes @EkoJR Since we're using wp_parse_url, which in 4.7+ handles non-absolute URLs, do we still need to call absolutize_url() here?
https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/5f119116455062cfdf11818ed11dad7ab62c1eac/modules/aioseop_sitemap.php#L4054

arnaudbroes:
@michaeltorbert interestingly, there were some variables there that aren't being used. wp_parse_url() also wasn't needed in that function because the argument that is passed on to the function is already parsed by wp_parse_url() elsewhere, so I removed that redundant code as well.

@wpsmort can you test this again, and specifically make sure that the image indexing for our sitemap still functions correctly?

michaeltorbert:
@arnaudbroes The specific functionality in is_image_url_valid() you changed/removed needs to be tested. Not just images in general.

arnaudbroes:
I've added absolutize_url() back on second thought since the unit test for schemeless images is failing for one specific schemeless image URL format ( "/image.jpg") even though the image is still being included in the sitemap on fresh and existing test sites. The format fails for this function but might still be processed and picked up somewhere else in the code. The other redundant code is still removed though.

@wpsmort I'm updating the testing instructions above.

michaeltorbert:
+1 for unit tests

wpsmort:
This is tested and is good.

Issue #2526, opened 2019-05-29
% sign in title breaks Preview Snippet


wpsmort:
**Describe the bug**
Placing multiple % signs in the title of a post or page breaks the Preview Snippet.  

NOTE: It does not break the placeholder text in the Title field or the doc title tag  in the source code.

**To Reproduce**
In order to reproduce this issue, follow these steps:

1. Install All in One SEO Pack on a new test site
2. Create a new Post and set a title without a %. You'll see that the Preview Snippet is fine
3. Now add two % signs to the title with some text in between each % sign, for example:

Testing 100% and another 100%

You'll see the Preview Snippet doesn't show any of the text between the two % signs.  See the screenshot below.

**Expected behavior**
Text between percent signs should be displayed correctly in the Preview Snippet.

**Screenshots**
<img width="801" alt="Screen Shot 2019-05-29 at 12 31 27 PM" src="https://user-images.githubusercontent.com/17548525/58574348-c6065400-820d-11e9-87f9-a2749e37da37.png">

**Version Info:**
 - WordPress version: 5.2.1
 - All in One SEO Pack version: 2.12.1
 - PHP version: 7.2.0

Arnaud believes that this is related to #1408.

wpsmort:
@arnaudbroes Please test back to see what version this bug first occurred in.

arnaudbroes:
I believe this is related to #1408.

Issue #1408, opened 2017-12-12
Double site title on posts page


michaeltorbert:
Reported here: https://semperplugins.com/support/troubleshooting-all-in-one-seo-pack-pro/posts-page-preview-snippet-shows-site-title-twice/

Preview snippet of posts page is showing page title | site title | site title

We need to determine whether this bug has always been there, or when it was introduced.

arnaudbroes:
@michaeltorbert first appeared in 2.2.4. I found this in the changelog:

> Added placeholder text to show auto generated titles and descriptions on the edit post/page screen

arnaudbroes:
Might be related (double vertical bar) - https://wordpress.org/support/topic/double-separator-in-wordpress-title/

arnaudbroes:
I've cloned Marcos site to my personal development server and will look into this server.

arnaudbroes:
Reported again here - https://wordpress.org/support/topic/title-shown-twice-in-preview-snippet-on-posts-page/

michaeltorbert:
@arnaudbroes Can we reproduce this yet?

wpsmort:
Reported again by Marcos by email today.

wpsmort:
I have been able to reproduce this on a new site by following these steps:

1. Set up WordPress with Reading Settings > Your homepage displays set to A static page and a Homepage and Posts page selected
2. Install the current version of AIOSEOP (v2.12.1) and activate it
3. Edit the Posts page and you should see the double site titles in the Preview Snippet

PR #2705, opened 2019-07-04
Fix preview snippet bugs

arnaudbroes:
Issue #1408 & #2526.

## Proposed changes

Fixes two bugs for the preview snippet title.

(1) Double % signs in the title break the title (#2526)
(2) Double site title on Posts page (#1408)

The former was happening because there was some code that was deleting everything in between two percentage signs (including the signs); the latter was happening because the title for the Posts page is already handled by get_aioseop_title().

In the meanwhile, I've also cleaned up the phpDocs and refactored a little bit.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

For #2526:
Add two % signs to the title of a new Post with some text in between each % sign and make sure the preview snippet displays the title correctly.

For #1408:
Set a static Posts page under Settings > Reading, navigate to it and make sure the preview snippet doesn't include the site title twice.


arnaudbroes:
@EkoJR can you confirm that lines 3417-3424 which I deleted weren't used by anything?

wpsmort:
This is tested and is good.

EkoJR:
@arnaudbroes It appears that code is being used, and I believe it is a default to fallback on when there is no title.

To reproduce.
1. From the admin menu, under a post type, click on 'Add New'
2. Go to AIOSEOP Title, and type anything. Then delete it.
3. The title should still show a title, but doesn't show any title.

PR2526-before

Changes to...
PR2526-after

aioseop_class.php
```diff
@@ -1022,48 +1022,54 @@ function custom_output_option( $buf, $args ) {
...
 		$title        = $info['title'];
 		$description  = $info['description'];
 		$keywords     = $info['keywords'];
 		$url          = $info['url'];
 		$title_format = $info['title_format'];
 		$category     = $info['category'];
-		$w            = $info['w'];
-		$p            = $info['p'];
+		$wp_query     = $info['w'];
+		$post         = $info['p'];
```
EkoJR: Thank you for changing the names on these. This is much better!

aioseop_class.php
```diff
@@ -1137,31 +1143,39 @@ public function get_title_format( $args ) {
...
+		 * Runs after we have processed the title format for the snippet preview is.
+		 *
+		 * @since 3.0.0
 		 */
 		do_action( 'aioseop_after_format_title' );
 
 		return $title_format;
 	}
 
+	function get_preview_snippet_title_helper( $title_format ) {
```
EkoJR: Missing phpDoc for the function.

aioseop_class.php
arnaudbroes: Thanks, I agree! I originally had no idea what I was looking at and had to browse through the code to find what these variables are.

arnaudbroes:
@EkoJR incredible catch (since that slipped past Steve and me during testing)! 

I was able to reproduce the issue and fix it, although it wasn't caused by the removal of those lines. This was happening because the JavaScript on the Edit page affects the part of the preview snippet title that is within the span element. I had moved the entire preview snippet title inside the span element,. The static parts (basically all macros besides %post_title%/%page_title%) had to be moved out of the span element again so that they remain static and aren't modified on user input.

arnaudbroes:
@wpsmort this needs to be retested. You'll now also want to test that the static parts of the title (all macros and input in the relevant title format field, e.g. %site_title%, with the exception of the page/post title) are not affected when the title field is modified and then emptied,

That sounds abstract, but here's an example: if the title format is `%current_month% | %post_title% | %site_title`, then only the middle part (%post_title%) should change when the title field is modified or emptied; the other two parts should remain unaffected.

wpsmort:
This is tested and is good.

Issue #2126, opened 2018-12-28
Static pages included twice in sitemap


wpsmort:
Reported here - https://semperplugins.com/support/troubleshooting-all-in-one-seo-pack-pro/sitemap-lists-same-url-twice-with-two-different-dates-of-updates/

The WooCommerce Shop page is included twice in the XML Sitemap and can the two entries may have different Priority, Frequency and Last Change date.

1. Install WooCommerce and AIOSEOP on a new site
2. Ensure the default pages have been created by WooCommerce so there is a Page called Shop
3. Activate the XML Sitemap module and view the sitemap
4. You should see the Shop page listed twice, see screenshot below.

I verified that the Shop page is included once as a page (one entry goes away if you uncheck Pages next to Post Types) and is included once as the post type archive for Products (one entry goes away if you uncheck Products next to Post Types).

Tested with:
- WordPress version: 5.0.2
- AIOSEOP version: 2.10.1
- PHP version: 7.2.0

<img width="1083" alt="screen shot 2018-12-28 at 6 29 58 pm" src="https://user-images.githubusercontent.com/17548525/50530798-2c584080-0acf-11e9-9b18-4842751c59e8.png">

wpsmort:
@arnaudbroes Can you test back through the old versions of AIO and see if this has always been the case or is new.

arnaudbroes:
@wpsmort traced this back to https://github.com/semperfiwebdesign/all-in-one-seo-pack/tree/6a40943844702930faa44cb4cc757f3b5a44c124.

arnaudbroes:
@wpsmort I'm no longer able to reproduce this anymore with the newest versions of All in One SEO Pack and WooCommerce nor with older versions (nor with the commit I traced it back to). 

Can you confirm this?

wpsmort:
@arnaudbroes I've set up a new site on your Cloudways server where the problem is present - https://wordpress-281813-887299.cloudwaysapps.com

arnaudbroes:
Whilst testing the PR for this issue Steve found that, if configured, the static homepage and blog page are also included twice.

arnaudbroes:
So we are also including the homepage twice; once in the Additional Pages index and another time in the Pages index. I'd suggest that we do the same as for #2126 which is to leave it under Additional Pages if it is a post archive page and leave it under Pages if it is a static page.

PR #2720, opened 2019-07-11
Don't include static pages twice

arnaudbroes:
Issue #2126

## Proposed changes

Fixes an issue where the WooCommerce shop page is included twice in the sitemap.
Also bases the timestamp for the WooCommerce shop page on the last modified product.

## Types of changes

## Checklist

## Testing instructions

1. Install this PR.
2. Install WooCommerce, complete the initial wizard and make sure the Shop page has been created.
3. Add a WooCommerce Product.
4. Check the sitemap and make sure the WooCommerce shop page is no longer included twice.
5. Check that the shop page's timestamp is based on the last modified product.


arnaudbroes:
@wpsmort the plugin was including the Shop page once as a Page and another time as a Product Archive page. I've removed it as a Product Archive page since it seems more logical to include it as a Page. 

An issue that arises from this is that the Last Change timestamp is based on when the Page was last changed; not on the latest change to a Product. I don't think that's a big issue here but I just wanted to let you know.

wpsmort:
@arnaudbroes This is tested and is good.  

I agree that in the case of the Shop page, because there is an actual page then it should be listed in the pages sitemap instead of the product sitemap.  However, I do think that the Last Change date should be based on the date of the last change to the products.

arnaudbroes:
@wpsmort this should do the trick.

arnaudbroes:
@wpsmort the Last Change date for the WooCommerce shop page is now based on the last modified product and the homepage and posts page are removed from the Additional Pages index and can be found under Pages if they are both static.

In addition to testing the WooCommerce shop page, you'll want to test that (1) the Additional Pages index is not listed on the root sitemap index page if both the homepage/posts page are static; (2) it is listed when any additional pages have been specified in the XML Sitemap menu.

wpsmort:
@arnaudbroes There are a couple of bugs with this PR, they are:

1. When sitemap indexes is enabled then the addl-sitemap is not included in the index even when there are URLs added in the Additional Pages section
2. When sitemap indexes is enabled and you navigate to the pages-sitemap then the following PHP notice is displayed:

woocommerce_get_page_id is <strong>deprecated</strong> since version 3.0! Use wc_get_page_id instead. in wp-includes/functions.php on line 4435

I've attached the full stack trace from the debug log.
[debug.log](https://github.com/semperfiwebdesign/all-in-one-seo-pack/files/3392748/debug.log)


3. When sitemap indexes is not enabled then the last change date for the Shop page is the date the page was last modified and not the date of the last product.

arnaudbroes:
@wpsmort the first issue is is something that is fixed in https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2721; the other two issues have been addressed.

wpsmort:
This is tested and is good.

arnaudbroes:
> Instead of using is_plugin_active(), which depends on knowing the directory name, it's better to use function_exists() or class_exists().
> In this case, I'd probably use something like:
> 
> `if ( class_exists( 'woocommerce' ) ) `

Good to know, I'll keep that in mind. 

Made the requested change and also added a TODO in the meanwhile.

arnaudbroes:
@wpsmort this needs a quick retest just to confirm that the timestamp for the shop page is still updated.

wpsmort:
This is tested and is good.

modules/aioseop_sitemap.php
```diff
@@ -2204,9 +2204,35 @@ public function get_sitemap_index_filenames() {
...
+		 * This will not be the case if there is both a static homepage/posts page and there are no additional pages specified.
+		 *
+		 * @since 3.2.0
+		 *
+		 * @return bool
+		 */
+		private function does_addl_sitemap_contain_urls() {
+			$is_addl_pages = $this->option_isset( 'aiosp_sitemap_addl_pages' );
+			if ( ! $is_addl_pages &&
+				( 0 !== get_option( 'page_on_front' ) ) &&
```
EkoJR: Should this be an `||`? Just in case one static page setting is used and not the other.

modules/aioseop_sitemap.php
```diff
@@ -3149,6 +3178,57 @@ public function get_addl_pages() {
...
+		 * Removes URLs from a sitemap page. This is used both for indexes and pages within indexes.
+		 *
+		 * @since 3.2.0
+		 *
+		 * @param array $pages
+		 * @param array $pages_to_remove
+		 * @return array $pages
+		 */
+		private function remove_urls_from_sitemap_page( $pages, $pages_to_remove ) {
+			$count = count( $pages );
```
EkoJR: This operation could potentially improve processing by changing it to. 

```
foreach ( $pages_to_remove as $page_url_to_remove ) {
	$key = array_search( $page_url_to_remove, array_column( $pages, 'loc' ) );
	if ( $key ) {
		unset( $pages[ $key ] );
	}
}
```

OR

```
foreach ( $pages_to_remove as $page_url_to_remove ) {
	if ( in_array( $page_url_to_remove, array_column( $pages, 'loc' ) ) ) {
		unset( $pages[ array_search( $page_url_to_remove, array_column( $pages, 'loc' ) ) ] );
	}
}
```

It would make debugging a little quicker though with the fewer loops.

modules/aioseop_sitemap.php
```diff
@@ -4284,6 +4373,87 @@ public function get_all_post_priority_data( $include = 'any', $status = 'publish
 
 			$links = $this->get_prio_from_posts( $posts, $this->get_default_priority( 'post', true ), $this->get_default_frequency( 'post', true ) );
 			$links = array_merge( $links, $this->get_archive_prio_from_posts( $posts ) );
+
+			$is_sitemap_indexes_disabled = empty( $this->options['aiosp_sitemap_indexes'] );
+			if ( $is_sitemap_indexes_disabled || ( ! $is_sitemap_indexes_disabled && 'page' === $include ) ) {
```
EkoJR: The `$include` variable can also be an array. Might want to also add `|| in_array( 'page', $include )`...but I suppose `$is_sitemap_indexes_disabled` usually handles multiple $includes.

Which would also mean $include isn't always a string in the phpDocs, and wasn't updated sometime in the past.

modules/aioseop_sitemap.php
```diff
@@ -4284,6 +4373,87 @@ public function get_all_post_priority_data( $include = 'any', $status = 'publish
...
+		private function get_prio_freq_static_homepage( $links ) {
+			if ( 0 === get_option( 'page_on_front' ) ) {
+				return $links;
+			}
+			$prio = $this->options['aiosp_sitemap_prio_homepage'];
+			$freq = $this->options['aiosp_sitemap_freq_homepage'];
+
+			$homepage_url = get_site_url() . '/';
+			$count = count( $links );
+			for ( $i = 0; $i < $count; $i++ ) {
```
EkoJR: This For loop could be reduced with either...

```
$test_key = array_search( $homepage_url, array_column( $links, 'loc' ) );
$test_keys = array_keys( array_column( $links, 'loc' ), $homepage_url );
```

modules/aioseop_sitemap.php
```diff
@@ -4284,6 +4373,87 @@ public function get_all_post_priority_data( $include = 'any', $status = 'publish
...
+		 * @return array $links
+		 */
+		private function update_woocommerce_shop_timestamp( $links ) {
+			if ( ! aioseop_is_woocommerce_active() ) {
+				return $links;
+			}
+
+			$shop_page_url = get_permalink( wc_get_page_id( 'shop' ) );
+			$count = count( $links );
+			for ( $i = 0; $i < $count; $i++ ) {
```
EkoJR: This For loop can be reduced by using either

```
$test_key = array_search( $shop_page_url, array_column( $links, 'loc' ) );
$test_keys = array_keys( array_column( $links, 'loc' ), $shop_page_url );
```

modules/aioseop_sitemap.php
arnaudbroes: No, this function is meant to check whether the addl sitemap file should be included in the root sitemap file. If there is no static homepage AND no static posts page AND no additional pages specified in the XML Sitemap menu, then there should be no addl sitemap file. If either of those conditions is true then it should be included.

modules/aioseop_sitemap.php
arnaudbroes: That won't be needed since, like you've said yourself, sitemap indexes only contain one type of post (or taxonomy).

modules/aioseop_sitemap.php
arnaudbroes: Done.

modules/aioseop_sitemap.php
arnaudbroes: Done.

modules/aioseop_sitemap.php
arnaudbroes: We'll leave this as is as discussed since I can't get this to work.

arnaudbroes:
Oops, shouldn't have tagged those commits with a hashtag. x)

Anyway, this can be reviewed again @EkoJR.

Issue #2534, opened 2019-05-30
Run validator everywhere AIOSEOP shows up


wpsmort:
The Validity Checker plugin for Chrome is a good place to start, but we may want to check out some others as well.

wpsmort:
The W3C audit is done and the results can be found here - https://docs.google.com/spreadsheets/d/1KCuWxeU0cRYXoGoU_HKflHNqH9B7zIzhrH0qY9tXLjI/edit?usp=sharing

I've resolved everything that I can resolve and will reassign this so that the remaining items can be resolved.

PR #1412, opened 2017-12-13
Error with Gutenburg

contactashish13:
#1410

PR #2740, opened 2019-07-23
Improvements to HTML / CSS for markup validation

wpsmort:
Issue #2534

## Proposed changes
This introduces numerous improvements to our HTML and CSS markup in order to pass the W3C validation.  

## Types of changes
- Improves existing code

## Checklist

## Testing instructions
Test all screens in all browsers, paying specific attention to the following:

 - Width of the settings screens when viewed in various browser widths
 - The plugin title in the About box on the settings screens
 - The link to Github in the About box on the settings screens
 - The bullet list in the About box on the settings screens
 - The contents of the Support box on the settings screens
 - The links in the Support box on the settings screens
 - Closing and opening the meta boxes on the settings screens
 - Saving settings on the settings screens
 - Layout of the options on the settings screens
 - The Additional Pages table on the Sitemap module
 - The table on the Robots module
 - The fields with character counters in the Home Page Settings section on the Open Graph module
 - The OG image on the Add New Post / Edit Post screen
 - The Debug This Post button on the Add New Post / Edit Post screen

Verify all CSS and JavaScript is loading correctly.  Check the code changes and let me know if you have any questions.


arnaudbroes:
This is good and can be code reviewed.

modules/aioseop_opengraph.php
```diff
@@ -1816,7 +1813,7 @@ function get_all_images( $options = null, $p = null ) {
 			$images = array();
 			if ( ! empty( $image ) ) {
 				foreach ( $image as $k => $v ) {
-					$images[ $v ] = '<img height=150 src="' . $v . '">';
+					$images[ $v ] = '<img alt="" height=150 src="' . $v . '">';
```
EkoJR: If you have the image ID available, you could use the image caption ( `wp_get_attachment_caption( $image_id );` ) or image title `get_the_title( $image_id )` for the alt value.

modules/aioseop_opengraph.php
```diff
@@ -364,7 +363,6 @@ public function init() {
 					'name'       => __( 'Home Description', 'all-in-one-seo-pack' ),
 					'default'    => '',
 					'type'       => 'textarea',
-					'class'      => 'aioseop_count_chars',
```
EkoJR: I just realized the operation for adding the class occurs at `admin/aioseop_module_class.php:2661` [link](https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/a9f2bb6b607c6b6720ff6d702acd0628b6119304/admin/aioseop_module_class.php#L2296), and it was just something I hacked up at the time to get it running and it's based on the 'count' value existing.

Do you think it would be more simplified to keep the class in the setting's configs, rather than having to search where it is being added?

EkoJR:
@arnaudbroes This also appears to have a merge conflict.

modules/aioseop_opengraph.php
wpsmort: @EkoJR That's a little beyond my skillset. If you're able to provide the code I can add it.

modules/aioseop_opengraph.php
wpsmort: @EkoJR That class is also used for the Home Description field on the General Settings screen which presumably uses the same code you link to.  If that's the case then we should leave as is for now, but you may want to open an issue to review and improve this.

modules/aioseop_opengraph.php
EkoJR: @wpsmort I took a look at it, and it doesn't appear to be possible. No ID is given, and the image default to `all-in-one-seo-pack/images/default-user-image.png`, which isn't part of a site's database to get the additional data.

modules/aioseop_opengraph.php
EkoJR: Actually, that looks like a commit Ashish made...I was thinking it may have been me https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/1412. At first, I was wondering why it was still working when it was removed, and I believe Ashish might have been unsure where to add the CSS class

Well, I did intend for it to be added to the default_options configs. Since that's where most of the layout configs are, which makes it easier to locate, and allows more control (in case there is a setting where we need the CSS/JS changed or removed from it).

wpsmort: @EkoJR I think we can leave as is for now then.

wpsmort:
@EkoJR Is this now ready to merge or do you need to code review again?

EkoJR:
@wpsmort @arnaudbroes Opps, I thought I was dealing with a PR from Arnaud this morning

Issue #2769, opened 2019-08-02
PHP Fatal error: Call to undefined function boolval in aioseop_class.php on line 4407


michaeltorbert:
Reported here: https://wordpress.org/support/topic/problem-after-updating-plugin-2/#post-11792754

PR #2770, opened 2019-08-02
change boolvar to bool for < php 5.5

michaeltorbert:
Issue #2769 

## Proposed changes

In 3.2, we introduced the boolvar() function, which requires PHP 5.5 or greater. I've replaced it with (bool) which should accomplish the same thing for our purposes.

## Types of changes

## Checklist

- [ x] I've selected the appropriate branch (ie x.y rather than master).
- [ x] I've created an appropriate title, descriptive of what the PR does.

## Testing instructions
- In AIOSEOP 3.2, with schema enabled, you should get a fatal error on the front-end when running PHP < 5.5.
- Test that this PR no longer gives a fatal error on the front-end with PHP < 5.5.
- Test that the schema info is output to the front end.

## Further comments

For code review, make sure that this accomplishes the same thing.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4404,7 +4404,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 		}
 
 		// Handle Schema.
-		if ( ! empty( $aioseop_options['aiosp_schema_markup'] ) && boolval( $aioseop_options['aiosp_schema_markup'] ) ) {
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good.  I confirmed I could reproduce the fatal error and that this PR fixes it.

michaeltorbert:
We need to go ahead and get this out.

Issue #2766, opened 2019-08-02
Illegal string offset in schema/graphs/graph-organization.php


arnaudbroes:
Reported here - https://wordpress.org/support/topic/problem-after-updating-plugin-2/

Warning: Illegal string offset 'width' in /home/upwardsp/public_html/wp-content/plugins/all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 114

Warning: Illegal string offset 'height' in /home/upwardsp/public_html/wp-content/plugins/all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 115

michaeltorbert:
The fix itself is pretty simple, but so far I've been unable to reproduce the warning. I tried setting $logo_meta to false in both PHP 5.3 and 7.3.

wpsmort:
I have access to a cloned user site where I can reproduce this.  I have identified this as a conflict with the Composer theme from innwithemes.com.  It only happens when Use Schema.org Markup is Enabled and Person or Organization is set to Organization.

Here's the full set of warnings and notices generated:

Warning: Illegal string offset 'width' in all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 114
Notice: Uninitialized string offset: 0 in all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 114
Warning: Illegal string offset 'height' in all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 115
Notice: Uninitialized string offset: 0 in all-in-one-seo-pack/inc/schema/graphs/graph-organization.php on line 115

wpsmort:
@EkoJR I've got a cloned site for you to troubleshoot with.

toxpal:
> I have identified this as a conflict with the Composer theme from innwithemes.com

You are talking about my cloned site then. However, be aware this issue might not be related to specific theme because the first user who reported this issue also included URL of his site. It's upwardspiralblog.com and it runs a different theme but still has the same problem.

EkoJR:
@toxpal It is possible it isn't isolated to a single theme. Theme's that hook into customizer, and possibly uses its own settings as well, but the value returned isn't a numeric attachment ID. When it retrieves the meta data for the image when it is expecting an ID, it will return empty values. Which then creates the PHP illegal string offset error.

vagelisp:
[...]

PR #2775, opened 2019-08-02
 Fix illegal string offset in schema graph organization

EkoJR:
Issue #2766

## Proposed changes

PHP error with an array variable not being set; which was produced by the $logo_id not returning an int image ID.

## Types of changes

## Checklist

## Testing instructions

Was not able to reproduce, but I speculate the following...

Requires a custom theme that hooks into the customizer, but uses its own settings. Add the image to the theme settings (not the customizer). Fetching the logo ID does not return a numeric value as expected, and then fetched meta data with a value that is expected to be an attachment ID.

Note: Testing may have to be done on the clone site.

## Further comments

Composer theme by Innwithemes is known to trigger the issue.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4404,7 +4404,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 		}
 
 		// Handle Schema.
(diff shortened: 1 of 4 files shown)


wpsmort:
This is tested and is good.

Issue #2781, opened 2019-08-04
array_key_exists() expects parameter 2 to be array, boolean given in aioseop_class.php on line 4735


wpsmort:
Reported here https://wordpress.org/support/topic/error-message-767/

Also reported by Jim Douglas by email today.

wpsmort:
Awaiting access to the sites so we can troubleshoot and reproduce.

wpsmort:
Reported by Carole Asselin by email today.

arnaudbroes:
This was happening because BuddyPress user profiles are considered as some kind of post type but there's actual post meta data tied to them which is why get_post_meta() is returning false instead of an array.

PR #2787, opened 2019-08-05
Fix array_key_exists() expects array parameter notice in aioseop-class.php

arnaudbroes:
Issue #2781

## Proposed changes

Fixes an issue where array_key_exists() does not receive an array as argument (in aioseop-class.php).

## Types of changes

## Checklist

## Testing instructions

@wpsmort as discussed on Skype. Just install the PR and make sure this fixes the issue with BuddyPress forum profiles. Also check that this doesn't affect other post types and taxonomies.


wpsmort:
This is tested and is good.

Issue #2805, opened 2019-08-09
PHP notice due to unsaved options


boonebgorges:
Please fill out this form as completely as possible so that we can help you more quickly.

- [x] I understand that this is a development repository and not a support forum. This is a bug report and is not a support request.
 - [x] I've followed the [troubleshooting steps](https://semperplugins.com/documentation/how-to-troubleshoot-issues-with-all-in-one-seo-pack)  to determine if there's a theme/plugin conflict. 

**Provide the relevant URL(s) here**

**Describe the bug**
Before plugin settings are initially saved, plugin throws PHP notices because of `array_key_exists()` checks in `check_recently_added_modules()`. See https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/189d80c5d37694b73175954ea221ff292ff076eb/aioseop_class.php#L5701

The best solution may be to have a central function for fetching the plugin options, a function which would ensure that proper defaults are in place. An easier fix is simply to change the line in question to:

```php
$options = get_option( 'aioseop_options', array() );
```

to ensure that the return value will be in an array.

Thanks!

michaeltorbert:
@boonebgorges Thanks for the report and input. 
The best way to make sure your issue is resolved and included in the next release is to submit a PR. Feel free to submit one if you like, which will have the added bonus of listing you in the credits of All in One SEO Pack. ;)

michaeltorbert:
Although I'm not sure where check_recently_added_modules() is... Is that a different plugin you're using or am I overlooking something?

michaeltorbert:
@boonebgorges We'll also need steps to reproduce.

wpsmort:
@boonebgorges Can you provide us with the steps to reproduce this PHP notice?

boonebgorges:
Not exactly. My guess is that it's a race condition that is hard to trigger consistently, as it requires two nearly-simultaneous requests of a specific type. See "Steps to Reproduce" at https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2807

PR #2807, opened 2019-08-09
Bugfix/avoid php notice

boonebgorges:
See #2805.

Issue #

## Proposed changes

See #2805 

## Steps to reproduce

I can't figure out reliable steps to reproduce. If the option is totally absent, then the schema update routines in `AIOSEOP_Updates` will reinstall the options before the 'shutdown' hook is fired.

My guess is that there's a race condition here, of the following form:
1. The plugin options are unsaved
2. Two AJAX requests are triggered at roughly the same time. These AJAX requests come from the front end, but go to wp-admin/admin-ajax.php, so `is_admin()` is true (and your 'shutdown' callback is fired).
3. The first of these requests triggers the 'version_updates' routine, but it doesn't have a chance to save the options to the database before the second request reaches the 'shutdown' hook. Thus, the `get_option()` call in that second request turns up nothing.

## Types of changes

## Checklist

## Testing instructions
- Don't assume the tester knows the entire backstory of the issue, and don't force him/her to decipher the code to try and figure out what -it's doing or how to test it.
- Do provide step by step instructions on how to test. 
- Do note things to watch out for.
- Do note what aspects the tester should try and break.

## Further comments

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -5698,7 +5698,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 	 */
 	function check_recently_activated_modules() {
 		global $aioseop_options;
-		$options = get_option( 'aioseop_options' );
+		$options = get_option( 'aioseop_options', array() );
 		$modules_before = array();
 		$modules_now    = array();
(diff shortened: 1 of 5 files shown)


arnaudbroes:
This can't be tested before we've reproduced the issue at hand.

Issue #2797, opened 2019-08-07
General - Noindex for paginated content does not apply to paginated homepage


michaeltorbert:
Reported here: https://wordpress.org/support/topic/the-robots-meta-tag-does-not-work/

> in the plugin version 3.2.2, the noindex meta tag on the pagination page stopped working. When switching from the main page of the site to the page pagination meta tag noindex no.
> 
> https://vellisa.ru/page/2
> 
> In the plugin settings, this option is enabled.
> 
> http://i.piccy.info/i9/12c2443fab24df6ce75d804a9778904b/1565170889/14923/1329154/2019_08_07_123818.png

and here: https://wordpress.org/support/topic/noindex-for-pagination-pages/

> In version 3.2.2. robots (noindex) meta tag parameters for pagination pages in HTML code are not displayed, although they are specified in the settings.

wpsmort:
I've reproduced the first issue here where noindex is no longer added to the paginated homepage.  

I have not been able to reproduce the second issue regarding paginated pages so I've requested more info from the person who reported this.

I need to some research into noindex for paginated homepages to see what the best practice is.

wpsmort:
@arnaudbroes As discussed today, we need to fix the problems with noindex / nofollow on paginated content that we introduced in v3.2.

PR #2815, opened 2019-08-15
Fix noindex for paginated content

arnaudbroes:
Issue #2797

## Proposed changes

Fix an issue where noindex for paginated pages doesn't affect the homepage and other archive pages.

## Types of changes

## Checklist

## Testing instructions

Make sure that the "Noindex paginated pages/posts" setting does not affect the first page of the homepage, but does affect page 2 and onward. Same goes for the static posts page and other pages.


wpsmort:
@arnaudbroes This works great for the homepage but needs to also work for the static Posts page.  

If you set a static home page and a static posts page then the Noindex / Nofollow paginated pages needs to be output on page 2 and higher of the posts (Blog) page.

arnaudbroes:
@wpsmort I had a look at v3.1.1 as discussed and noticed that "noindex for paginated posts/pages" seemed to affect paginated content everywhere so I just went with that. That might not have been the intended behaviour (although I also went back a few more versions to see if anything was different) but that's how it used to be. We might want to consider renaming that setting to "noindex for paginated content" though.

wpsmort:
@arnaudbroes There's a typo on line 4643, you used 'aiosp_paginated_nofollow' for the noindex option.

wpsmort:
This is tested and is good.  I fixed an error in the code where nofollow was incorrectly used for noindex.

Issue #2389, opened 2019-05-07
Improve travis config


michaeltorbert:
Looking at some other configurations, I wonder if there are things we can do to improve ours.

For example:
https://travis-ci.org/Automattic/jetpack/builds/529421646/config
https://travis-ci.org/woocommerce/woocommerce/builds/529406948/config
https://travis-ci.org/easydigitaldownloads/easy-digital-downloads/builds/529410324/config

They have coding standards checks, PHP 7.3, WP nightly, etc.

EkoJR:
PR was intended to get travis operational again. There is still room for improvements.

arnaudbroes:
@EkoJR this should be done in this release.

PR #2820, opened 2019-08-19
Test Travis yml

EkoJR:
Issue #2389

## Proposed changes

Improve Travis CI matrix builds to include PHP 7.3, multiple WP versions supported, and improve code & documentation.

## Types of changes

- Improves existing code

## Checklist

- [ ] I've created an appropriate title, descriptive of what the PR does.

## Testing instructions
No testing required.

## Further comments


arnaudbroes:
@EkoJR this is great work. I'd suggest that we also remove both tests that don't support PHPUnit from the list with allowed failures as both aren't failing and I'm also wondering whether we should just go ahead and remove the multisite test since it has never worked before?

EkoJR:
@arnaudbroes I was tempted to remove the PHP versions unsupported by PHPUnit, but kept it more or less as a placeholder since we support those versions. I would lean towards removing those 2 PHP version tests though.

The multisite test, I believe should remain even though it has mostly been a placeholder. At some point I would like to get unit testing support restored to multisite, and then I could add various tests for that.

arnaudbroes:
This is good.

PR #2828, opened 2019-08-25
Merge 3.2.4 into 3.3

arnaudbroes:


arnaudbroes:
Ben and I agreed this could be merged and was likely failing because of old phpcs rules.

Issue #2847, opened 2019-08-29
Dev Env - Change order Travis matrix and add PHP 5.2 and 5.3 tests


arnaudbroes:
We should move the lowest versions and Grunt to the top so that we know faster when a build fails.

PR #2848, opened 2019-08-29
Improve Travis configuration

arnaudbroes:
Issue #2847

## Proposed changes

Improves our Travis configuration:

- Adds support for PHP 5.2 and PHP 5.3
- Move most critical tests to top of our build matrix in a separate stage

Note: additional stages other than the default "test" stage seem to run the PHP 7.3/WP_VERSION=latest test at the beginning and I have no idea how to disable this. Support for stages in combination with a build matrix seems to be rather poor at the moment. However, this isn't a bad thing for the time being as it doesn't hurt to run our unit tests in that stage as well.

## Types of changes

- Improves existing code

## Checklist


arnaudbroes:
@EkoJR here are some changes for Travis. I've added support for PHP 5.2 and 5.3, added a "critical" stage so that we don't have to wait until all tests are run to see if the PR needs work and I've also removed support for WP <4.9 so that our builds don't take as long. WP 4.9 is now our lowest supported version.

EkoJR:
@arnaudbroes I thought you wanted PHP >= 5.5 removed.

This also has a duplicate test with `PHP 7.3 WP_VERSION=latest`.

 Also, this makes it so we do have to wait, and the tests take a little bit longer as well.

That is why `fast_finish: true` is in travis.yml, and it seems you would have better luck with `allow_failures` for skipping the wait, but that kind of defeats the purpose of it. Unfortunately, I couldn't find anything in Travis that would allow skipping other builds.

I did try to expedite Grunt runs, but was unaware of Travis stages. I do think it would be better to make the Grunt Task Runner run first since it does take the longest and there are concurrent runs.

With PHP 5.2 & 5.3, I would suggest moving it down to `allow_failures` so `fast_finish` can take effect and PHPUnit also isn't supported with those versions.

michaeltorbert:
Don't we want to know if 5.2/5.3 fail?

> With PHP 5.2 & 5.3, I would suggest moving it down to allow_failures

arnaudbroes:
> I thought you wanted PHP >= 5.5 removed.

No, 5.2-5.5 are still supported till WP<5.1 so we have to keep supporting it for now. 

> This also has a duplicate test with PHP 7.3 WP_VERSION=latest.

I'm not sure why that's happening (I tried to remove it), but I'm guessing that it's caused by the fact that stages aren't fully supported for matrices yet. However, I'm fine with it since that already runs our unit tests in the first stage.

> That is why fast_finish: true is in travis.yml, and it seems you would have better luck with allow_failures for skipping the wait, but that kind of defeats the purpose of it. Unfortunately, I couldn't find anything in Travis that would allow skipping other builds.

So the idea behind these improvements is that we just run a set of "critical" tests first so we can immediately see (as the developer) whether the PR needs to be adjusted. I don't mind that much that the full build takes longer now. That's fine; this is more to get a quick idea of what's going on.

> I did try to expedite Grunt runs, but was unaware of Travis stages. I do think it would be better to make the Grunt Task Runner run first since it does take the longest and there are concurrent runs.

That's a good idea. I'll move that one up.

> With PHP 5.2 & 5.3, I would suggest moving it down to allow_failures so fast_finish can take effect and PHPUnit also isn't supported with those versions.

So as mentioned by Michael, we don't those to fail so I can't put them back in there. I did disable PHPUnit for them; that's just something we'll have to live with unfortunately.

EkoJR:
@michaeltorbert 5.2 and 5.3 will never fail because there is no PHPUnit. So, there is nothing happening on Travis.

arnaudbroes:
> @michaeltorbert 5.2 and 5.3 will never fail because there is no PHPUnit. So, there is nothing happening on Travis.

I'll probably remove them once https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2851 is implemented. 

@EkoJR so what's the point of having all those environment tests then. Does it only run our unit tests in a different environment each time?

EkoJR:
@arnaudbroes Yea, those run code, but PHP 5.2 & 5.3 haven't done anything.

arnaudbroes:
@EkoJR I removed the PHP 5.2 & 5.3 tests again.

EkoJR:
@arnaudbroes I found the problem to the duplicate test. Using a hyphen (-) indicates a seperate test, and when the critical stage was initiated, it was also set as it's own test, which fell back on the default tests values (PHP 7.3 & WP Latest being the initial defaults).

EkoJR:
I noticed for some odd reason. This takes longer than it did. Even though number of tests have been reduced, and grunt is expedited to prevent the last test from prolonging approvals.

arnaudbroes:
> I noticed for some odd reason. This takes longer than it did. Even though number of tests have been reduced, and grunt is expedited to prevent the last test from prolonging approvals.

I noticed two of the tests ran for an abnormal amount of time so I wouldn't worry about it. In any case, the total duration doesn't matter that much anymore now that the first three tests quickly tell us if we need to fix anything. 

Also, good job on getting removing that duplicate test!

Issue #2739, opened 2019-07-23
Bug - Date field not supported in Safari


wpsmort:
The date type for a form input field is not supported by Safari so the field gets rendered as a text input field.  This came up when running HTML validation on all our screens.  It specifically affects the Additional Pages section in the XML Sitemap.

<img width="612" alt="Screen Shot 2019-07-23 at 3 19 16 PM" src="https://user-images.githubusercontent.com/17548525/61740700-4607f080-ad5d-11e9-960c-e5bab52cb6b4.png">

wpsmort:
I've found a way to fix this using just CSS.

wpsmort:
Here's a screenshot of the newly styled date picker in Safari:

<img width="698" alt="Screen Shot 2019-08-29 at 6 02 10 PM" src="https://user-images.githubusercontent.com/17548525/63980161-ccb09b80-ca88-11e9-9403-2fdd60e594b2.png">

arnaudbroes:
Microsoft Edge:

![image](https://user-images.githubusercontent.com/18156697/64074385-d9acd500-ccaa-11e9-9108-cdf319d24e6e.png)

Mozilla Firefox:

![image](https://user-images.githubusercontent.com/18156697/64074398-f6e1a380-ccaa-11e9-9e4c-e5dcbf1d867c.png)

Internet Explorer:

![image](https://user-images.githubusercontent.com/18156697/64074409-242e5180-ccab-11e9-8297-8a5338be5abf.png)

Google Chrome:

![image](https://user-images.githubusercontent.com/18156697/64074428-49bb5b00-ccab-11e9-93ce-1e0dd2ce3fd4.png)

@wpsmort all look good except for Google Chrome, but that seems to be pre-existing. Does it look the same in Chrome on MacOS, and if so, could we improve that too?

arnaudbroes:
For Chrome I'm specifically referring to the lack of a border of the right side of the box and that there's two action buttons in the date field; one to manually increase or decrease the number; the other to open the datepicker.

PR #2853, opened 2019-08-29
Improved display of date picker in Safari

wpsmort:
Issue #2739

## Proposed changes
The jQuery UI date picker doesn't look good in Safari so this adds CSS to improve the look.

## Types of changes
- Improves existing code

## Checklist

## Testing instructions
- Go to All in One SEO > XML Sitemap > Additional Pages and check the styling of the date picker in the Last Modified option
- Verify the styling in all modern browsers (specifically IE and Safari)
- Perform the same tests with the XML Sitemap and Video Sitemap in Pro

## Further comments
I've tested this in mobile Safari and it does not affect the look or behavior of the date field.  You should verify this does not affect mobile browsers on Android devices.


arnaudbroes:
@wpsmort 

Microsoft Edge:

image

Mozilla Firefox:

image

Internet Explorer:

image

Google Chrome:

image

@wpsmort all look good except for Google Chrome, but that seems to be pre-existing. Does it look the same in Chrome on MacOS, and if so, could we improve that too?

arnaudbroes:
For Chrome I'm specifically referring to the lack of a border of the right side of the box and that there's two action buttons in the date field; one to manually increase or decrease the number; the other to open the datepicker.

wpsmort:
@arnaudbroes It looks fine in Chrome on Mac (the right border is present):

Screen Shot 2019-09-03 at 12 00 43 PM

Can you tell why the border is missing in Chrome on Windows?

wpsmort:
I discussed this with Arnaud and it turns out that the problem with the right border missing in Chrome on Windows was due to a browser plugin and did not happen if he tested in Incognito mode.

Testing is therefore good in all browsers.

Issue #2524, opened 2019-05-29
UX - Review accessibility of new tooltips


michaeltorbert:
As an example: https://a11y.nicolas-hoffmann.net/simple-tooltip/

PR #2856, opened 2019-08-30
Improve accessibility of tooltip help

wpsmort:
Issue #2524

## Proposed changes
This improves the accessibility of our tooltip help for users who use a screen reader.  This PR makes three changes to the UI:

- It adds tabindex to the help icon so that you can tab to the help icon so that a screen reader can read the help text
- It adds the option name to the help text so that there is context to the help (which option the help is relevant to)
- It removes the colon from the end of all option labels (some have a colon some don't) so that our option labels are standardized and so that the option name in the help text can have a colon to add a pause when using screen readers

## Types of changes
- Improves existing functionality (UI)
- Improves existing code

## Checklist

## Testing instructions
- Install a screen reader or enable the screen reader feature in your browser or OS
- Tab through all settings to verify you can tab from the help icon to the option input
- Verify that as you tab to the help icon, the screen reader should read the option name followed by the help text 
- Verify that the colon has been removed from all option labels so that the standard for all options is no colon at the end of the label

## Further comments
A note about removing the colon from labels:  We've not had a standardized format for option labels in the past so some labels ended with a colon and some did not (older labels from pre v2.0 had colons).  

This not only adds a standard label format for all labels but it also allows for the colon to be added in the help text which results in a pause between a screen reader reading the option name and the option text which is good for usability.

### DIFF
--- a/admin/aioseop_module_class.php
+++ b/admin/aioseop_module_class.php
@@ -2788,7 +2788,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Module' ) ) {
 			if ( 'none' != $opts['label'] ) {
 				$tmp_help_text = $info->get_help_text( $name );
 				if ( isset( $tmp_help_text ) && ! empty( $tmp_help_text ) ) {
-					$display_help = '<a tabindex="0" class="aioseop_help_text_link" style="cursor: help;" title="%s"></a><label class="aioseop_label textinput">%s</label>';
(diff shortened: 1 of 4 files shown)


arnaudbroes:
@wpsmort I made a small change where I added an H4 to make it look better for our regular users. Screen readers should ignore it. Let me know what you think.

wpsmort:
This is tested and is good.  The change made by Arnaud is good.

Issue #2868, opened 2019-09-05
Enable schema.org markup by default


michaeltorbert:
Can anyone thing of a compelling reason why we shouldn't?

wpsmort:
We originally disabled this because our Schema output wasn't very good and it was causing conflicts.  Now that our Schema output is very good, we should enable it by default.

PR #2871, opened 2019-09-05
Enable Schema by default

wpsmort:
Issue #2868

## Proposed changes
This changes the default setting for Use Schema.org Markup to Enabled. This only affects new installs.

## Types of changes
- Improves existing code

## Checklist

## Testing instructions
- Install this on a fresh site and confirm that the Use Schema option is enabled by default
- Install this on an existing site and reset the General Settings to default and confirm that the Use Schema option is enabled by default
- Install this on an existing site with the current live version and where Use Schema is disabled, confirm that the option is not changed


arnaudbroes:
Same as #2878 applies here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2878#issuecomment-529127141

arnaudbroes:
This has been tested and can be code reviewed. I checked both the General Settings and database options after I had installed the plugin on a new site.

Issue #2876, opened 2019-09-06
Bug - Robots meta tag doesn't have correct closing tag


Babene03:
In the Sourcecode of the Site, is no END Tag **>** from the Meta Tag

`<meta name="robots" content="noindex,nofollow"`

michaeltorbert:
I'm unable to reproduce this behavior, which means it's likely a problem specific to your site. Please use these troubleshooting steps to narrow down a plugin/theme conflict. https://semperplugins.com/documentation/how-to-troubleshoot-issues-with-all-in-one-seo-pack/

If you're able to reproduce it on your site with all plugins other than AIOSEOP active, and with the unmodified default theme active, then please post the link to your site.

michaeltorbert:
Thank you for reporting this. I've been able to reproduce this and have verified it is a bug.
I've assigned it to be fixed for the next major release.

michaeltorbert:
@Babene03 
Thank you for reporting this. I've been able to reproduce this and have verified it is a bug.
I've assigned it to be fixed for the next major release.

michaeltorbert:
The problem is on https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/8b5b6a6960af511fdd786d5a42b8d768044b05dd/aioseop_class.php#L4219

Also, and this should probably be its own issue, we seem to be doing robots meta in two places. https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/8b5b6a6960af511fdd786d5a42b8d768044b05dd/aioseop_class.php#L4341
We should probably look into improving that.

michaeltorbert:
As this was broken in 3.2.2, we should include it in 3.2.6.

Babene03:
I think the issue is

`printf( "<meta name=\"robots\" content=\"%s\" >\n", esc_attr( $robots_meta ) );`

michaeltorbert:
@Babene03 Correct, which is also what I've posted above. ;)

PR #2878, opened 2019-09-06
Add closing tag to robots meta

michaeltorbert:
Issue #2876 

## Proposed changes

We're missing the closing tag after the robots meta.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Create a post.
- On the edit post screen, check "Disable on this page/post"
- On the edit post screen, check NOINDEX and/or NOFOLLOW.
- Verify that you see the robots meta tag without the closing tag in the source code of the post.
- Apply the fix.
- Verify it fixes it.

## Further comments

As described in https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2876#issuecomment-528972928 we inadvertently removed the closing tag in our robots meta. 
It's important to note that we have two places in the code where the robots meta tag could be generated. This affects the less likely one, so when testing, make sure you've first reproduced the problem on the page where you're testing the fix.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4216,7 +4216,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			$robots_meta = apply_filters( 'aioseop_robots_meta', $this->get_robots_meta() );
 
 			if ( ! empty( $robots_meta ) && 'index,follow' !== $robots_meta ) {
-				printf( '<meta name="robots" content="%s"', esc_attr( $robots_meta ) ) . " />\n";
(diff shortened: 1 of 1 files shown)


arnaudbroes:
I made a small change ("echo sprintf" instead of "printf") because the closing tag and new line character weren't being printed. I then tested this and confirmed this fixes the issue. 

@wpsmort can you confirm again that the fix works and then code review this?

arnaudbroes:
Also, the build failures can be ignored as this PR targets the 3.2.6 branch which doesn't include our Travis improvements/fixes. I ran Grunt and our unit tests on this in my local development environment.

wpsmort:
I've confirmed this works and reviewed the code and it looks good.

Issue #2756, opened 2019-08-01
Restrict message about sitemap posts per file to sites with the module active


michaeltorbert:
Reported here: https://wordpress.org/support/topic/notice-to-avoid-problems-with-your-xml-sitemap/

PR #2892, opened 2019-09-10
 Trigger sitemap notice only when XML Sitemap module is active

arnaudbroes:
Issue #2756

## Proposed changes

Fixes an issue where our sitemap notice is triggered even when the XML Sitemap module isn't enabled.

## Types of changes

## Checklist

## Testing instructions

Enable the XML Sitemap module, then enable Sitemap Indexes and set the amount of pages per index to something larger than 1,000. The notice should appear (if this is on a fresh test site). Now dismiss the notice and disable the XML Sitemap module again. Then go to the database, delete aioseop_notices and make sure the notice isn't triggered again. 

I believe this only happens when people have previously used our XML Sitemap module and the their default amount of pages was set to something larger than 1,000.


wpsmort:
@arnaudbroes I tested and found that this is not removing the message completely.  It removes the text from the message box but leaves the box there.  See screenshot that shows the empty message box:

Screen Shot 2019-09-10 at 11 19 32 AM

arnaudbroes:
@wpsmort this can be retested.

wpsmort:
@arnaudbroes I tested this again and the notice is being displayed when the Sitemap module is deactivated.  I think the problem is that the code you've added is in modules/aioseop_sitemap.php and that won't run if the Sitemap module is deactivated.

wpsmort:
Arnaud and I tested this together and couldn't get the notice to reappear when the Sitemap module is activated and Sitemap Indexes is disabled, or it's enabled and number of pages per sitemap is set to over 1000.

arnaudbroes:
Closing this as this was just a low effort to fix this quickly. In retrospect, the whole notice class/code probably needs a bit of work.

Issue #2863, opened 2019-09-03
make bool ready for php upgrade


michaeltorbert:
In a recent version, we had to change boolval() to (bool) to accomodate older versions of PHP.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/e780901af2ef8b690c53fbe7e87ab960e76852bf/aioseop_class.php#L4408

We should do something to support older versions of PHP, but then make it obvious when it's time for us to drop support.

Possible choices:

`if (!function_exists('boolval')) {
        function boolval($val) {
                return (bool) $val;
        }
}`

`if ( version_compare( PHP_VERSION, '5.5', '>=' ) ) {`

michaeltorbert:
@EkoJR @arnaudbroes Any input?

arnaudbroes:
The latter is more clear to me.

michaeltorbert:
It also probably makes it more obvious when dropping PHP support, since we just have to search for version_compare.

> The latter is more clear to me.

PR #2894, opened 2019-09-10
Make bool ready for php upgrade

michaeltorbert:
Issue #2863

## Proposed changes

This uses both the pre and post PHP 5.5 bool/boolvar(), and makes it easy to find for when we drop support for older versions of PHP.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Make sure schema shows up when checked and doesn't when unchecked.


wpsmort:
This is tested and is good.

Issue #2913, opened 2019-09-17
Add filter hook to disable schema.org markup


michaeltorbert:
This way someone can programmatically disable it as needed.

PR #2915, opened 2019-09-17
Add filter hook to disable schema.org markup

michaeltorbert:
Issue #2913

## Proposed changes

Adds an API filter hook to programmatically disable our schema.org markup.

## Types of changes

- Adds new API hooks

## Checklist

## Testing instructions
- By default this conditional is true. If you do nothing, our schema markup should be outputted as normal.
- Returning false should cause schema to not load. (You can either make a callback function or just use __return_false.) 
- Feel free to get fancy with it, and only add the filter for certain post types, a certain post, etc.
- Our schema output should still load anywhere else on the site, not specified in the above step.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4460,8 +4460,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 		 *
 		 * @since 3.2.8
 		 *
-		 * @param string
-		 * @return string
+		 * @return boolean
 		 */
 		if ( apply_filters( 'aioseop_disable_schema', '__return_true' ) ) {
 			// Handle Schema.
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4453,16 +4453,28 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			echo "$meta_string\n";
 		}
 
-		// Handle Schema.
-		if ( version_compare( PHP_VERSION, '5.5', '>=' ) ) {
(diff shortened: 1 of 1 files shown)


arnaudbroes:
This has been tested and is good.

arnaudbroes:
FYI, the remaining two Grunt issues can be ignored and are addressed in another PR.

aioseop_class.php
```diff
@@ -4453,16 +4453,27 @@ function wp_head() {
...
+		/**
+		 * The aioseop_disable_schema filter hook.
+		 *
+		 * Used to disable schema.org output programatically.
+		 *
+		 * @since 3.2.8
+		 *
+		 * @return boolean
+		 */
+		if ( apply_filters( 'aioseop_disable_schema', '__return_true' ) ) {
```
EkoJR: Couple issues here.

1. Using `'__return_true'` remains a string, that option is for `add_filter()` and doesn't work the same for `apply_filter()`. When testing the hook, it was not a boolean, but instead was `'__return_true'`.  This would be better as a boolean (which seems to be the intent).
2. Would it be better to reverse the boolean so it makes more sense to other users as `add_filter( 'aioseop_disable_schema', '__return_true' );` rather than `add_filter( 'aioseop_disable_schema', '__return_false' );`?

I was thinking something like this would work. `if ( ! apply_filters( 'aioseop_disable_schema', false ) ) {`

michaeltorbert: 1) Good catch. I'll change that. 
2) Also a good catch. I had it backwards. :)

michaeltorbert:
@arnaudbroes @EkoJR Can you both test and code review again?

EkoJR:
Grunt was also clean.

arnaudbroes:
I've tested this and code looks good too.

Issue #2914, opened 2019-09-17
Chrome 77 and Gutenberg conflict


michaeltorbert:
We should have a version check, as I think this will be fixed in WordPress 5.3. Once we drop support for WP < 5.3, we can remove the code.

Background: 
In Chrome 77 and Gutenberg for WP (5.0-5.2), metaboxes overlap with post text. It's not an AIOSEOP bug, but people think it is.

https://github.com/WordPress/gutenberg/issues/17406

PR #2921, opened 2019-09-18
Fix conflict between Chromium v77.0.3865.75 and Gutenberg

arnaudbroes:
Issue #2914

## Proposed changes

Fixes a conflict between Chromium browsers and the Gutenberg editor that causes meta boxes to overlap with the content.

The PR also includes Grunt fixes which are needed for 3.2.8; else Travis CI will fail.

## Types of changes

## Checklist

## Testing instructions

Install a browser based on Chromium v77.0.3865.75 as well as All in One SEO Pack from WordPress.org. You'll notice that our meta box overlaps with the content. Next, install this PR and confirm that it fixed this.

## Further comments

While this isn't an issue we caused, users thinks we did, so that's why we're adding this fix.


michaeltorbert:
@arnaudbroes When was this bug introduced? WP 5.2?
Should version_compare() be 5.2 < x < 5.3, assuming it's fixed in 5.3?

arnaudbroes:
> @arnaudbroes When was this bug introduced? WP 5.2?
> Should version_compare() be 5.2 < x < 5.3, assuming it's fixed in 5.3?

No, because it wasn't introduced in Gutenberg; it's a Chromium issue. So what we'll want to keep track of is whether Gutenberg will still include a fix for this in 5.3 now that Chromium fixed this on their end.

michaeltorbert:
@arnaudbroes Ok I'll rephrase. Does this bug affect WP < 5.2?

arnaudbroes:
> @arnaudbroes Ok I'll rephrase. Does this bug affect WP < 5.2?

Yes it does.

michaeltorbert:
@arnaudbroes Then it should be checking for WP < 5.3 (assuming WP fixes it then).

michaeltorbert:
I've tested this on WP 5.0 and 5.3. After confirming the bug on each, have confirmed the fix on each as well.

Issue #2662, opened 2019-06-16
Better controls over welcome page


michaeltorbert:
Probably in the updates file, we should make it explicit for when the welcome page appears... ie after major releases.

arnaudbroes:
Related to https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/2631.

arnaudbroes:
Requested again here - https://wordpress.org/support/topic/annoying-splash-screen-when-updating/

PR #2922, opened 2019-09-18
Only show Welcome Screen on major version updates

michaeltorbert:
Issue #2662

## Proposed changes

Currently we show the welcome screen on every update, which can be a bit annoying for some users. This truncates everything except the major version part of a version.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Testing is a bit tricky. You'll need to make sure you can reliably reproduce the welcome screen on demand.
- Open your favorite db tool and in aioseop_options in the options table you should see the latest checked version at the end. This is from our db upgrade code in aioseop_updates_class.php.
-Also look for aioseop_seen_about_page. This is the latest version of the about page someone has seen.
-You can delete or modify these numbers to manipulate your environment.
-1.2.3, 1.2.3.4, 1.2 should all have the same result. If you've seen any of them, you don't need to see it again.
-Make sure nothing breaks it, for instance lots of numbers, few numbers, etc.

## Further comments

Ideally both @EkoJR and @arnaudbroes can do a code review, since if we get this wrong it could be very very annoying.

### DIFF
--- a/admin/display/welcome.php
+++ b/admin/display/welcome.php
@@ -106,7 +106,8 @@ if ( ! class_exists( 'aioseop_welcome' ) ) {
 				return;
 			}
 
-			if ( ( AIOSEOP_VERSION === $seen ) || ( true !== $activate ) ) {
+			// Compare the major versions so we don't show the welcome screen on minor versions.
+			if ( ( get_major_version( AIOSEOP_VERSION ) === get_major_version( $seen ) ) || ( true !== $activate ) ) {
 				return;
 			}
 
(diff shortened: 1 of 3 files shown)


arnaudbroes:
FYI, the remaining two Grunt issues can be ignored and are addressed in https://github.com/semperfiwebdesign/all-in-one-seo-pack/pull/2921.

arnaudbroes:
@michaeltorbert 

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/c4a21d4536a157011ddd4577f89328f27f7e181c/admin/display/welcome.php#L109-L112

Here you'll either need to instantiate the aiosp_common class since that's where that function is currently located or even better would be to move it to aioseop_functions. Now it's causing a fatal error. Apart from that, this is working great.

Issue #2939, opened 2019-09-26
Parse error: syntax error, unexpected T_FUNCTION, expecting T_PAAMAYIM_NEKUDOTAYIM 


michaeltorbert:
Reported here: https://wordpress.org/support/topic/bug-report-parse-error-compat-gutenberg-php/

> Parse error: syntax error, unexpected T_FUNCTION, expecting T_PAAMAYIM_NEKUDOTAYIM in /home/d31336/cyklavandra.se/wp-content/plugins/all-in-one-seo-pack/inc/compatability/compat-gutenberg.php on line 28

We inadvertently introduced a PHP 5.3 requirement in 3.2.8.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/d514d55aacfe9c054e081e1344174741b9cf6606/inc/compatability/compat-gutenberg.php#L28

PR #2940, opened 2019-09-26
remove anon function for php 5.2 compat #2939

michaeltorbert:
Issue #2939

## Proposed changes

In 3.2.8 we added an anonymous function. This removes that for PHP 5.2 compatibility. 
I've also prefixed the functions.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- First, verify the fatal error in PHP 5.2
- Make sure it's fixed with this patch

Also, we need to make sure the Gutenberg/Chrome fix still works.
- First, verify the bug exists in Chrome 77.
- Test that this patch still fixes it: specifically in PHPs 5.2, 5.3, 7.x, and in WPs 5.1, 5.2.

### DIFF
--- a/inc/compatability/compat-gutenberg.php
+++ b/inc/compatability/compat-gutenberg.php
@@ -21,11 +21,20 @@
  *
  * @return void
  */
-function gutenberg_fix_metabox() {
+function aioseop_gutenberg_fix_metabox() {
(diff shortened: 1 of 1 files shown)


EkoJR:
I've tested this with multiple PHP versions on chrome 77, and this looks good.

arnaudbroes:
This has been tested. I was able to reproduce the original issue and confirmed that this fixes it (as well as that it still fixes the CSS issue). Code also looks good.

Issue #2938, opened 2019-09-26
Refactor operations in All_in_One_SEO_Pack_Opengraph::add_meta()


EkoJR:
**Is your feature request related to a problem? Please describe.**

In `All_in_One_SEO_Pack_Opengraph::add_meta()` at the bottom of the function/method there is a variable variables operation used initiate variables within an array.

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/d514d55aacfe9c054e081e1344174741b9cf6606/modules/aioseop_opengraph.php#L1575

This kind of concept makes it difficult to track down variables, and is prone to errors if any variable within the function gets renamed or deleted Changing this also reduces the amount of warnings (unused variables) from 74 down to 44.

There is also a number of other problems that need cleaning up.

1. Renaming foreach variables from single characters to a proper name that better defines its use.
2. Change to group operations related to each other. Currently some operations are spread out within the function/method.
3. Unnecessarily using a filter to access a method within the same class, and the filter has only a single use. Changing to use the method directly makes it easier to track/inspect.
4. Add missing Filter Documentation.

**Describe the solution you'd like**
Change the variable variables concept to be initiated directly by those variables.

**Additional context**

PR #2941, opened 2019-09-26
Refactor operations in All_in_One_SEO_Pack_Opengraph::add_meta()

EkoJR:
Issue #2938

## Proposed changes

Refactors operations in `All_in_One_SEO_Pack_Opengraph::add_meta()` to fix a number problems that made tracking variables and debugging difficult.

1. Refactors variable variables to be direct variables.
2. Renaming foreach variables from single characters to a proper name that better defines its use.
3. Change to group operations related to each other. Currently some operations are spread out within the function/method.
4. Unnecessarily using a filter to access a method within the same class, and the filter has only a single use. Changing to use the method directly makes it easier to track/inspect.
5. Add missing Filter Documentation.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
Code refactor, so no changes to expected output. Testing would primarily be on the frontend.

1. Enable the Opengraph Module.
2. Ensure CTPs are enabled on content you are testing (ex. Post and Pages).
3. Any options that may effect output can be changed as desired.
4. (Optional) Go to edit post, and modify any Opengraph settings.
5. Go to frontend and check for any errors or missing data.

## Further comments

### DIFF
--- a/modules/aioseop_opengraph.php
+++ b/modules/aioseop_opengraph.php
@@ -1526,42 +1526,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Opengraph' ) ) {
 			);
 
 			// TODO Remove when `$tmp_meta_slug` is removed from 'aiosp_opengraph_meta' filter.
-			$meta_keys = array(
-				'facebook' => array(
-					'og:type'                => 'type',
-					'og:title'               => 'title',
-					'og:description'         => 'description',
-					'og:url'                 => 'url',
-					'og:site_name'           => 'sitename',
-					'og:image'               => 'thumbnail',
-					'og:image:width'         => 'width',
-					'og:image:height'        => 'height',
-					'og:video'               => 'video',
-					'og:video:width'         => 'videowidth',
-					'og:video:height'        => 'videoheight',
-					'fb:admins'              => 'key',
-					'fb:app_id'              => 'appid',
-					'article:section'        => 'section',
-					'article:tag'            => 'tag',
-					'article:published_time' => 'published_time',
-					'article:modified_time'  => 'modified_time',
-					'article:publisher'      => 'publisher',
-					'article:author'         => 'author',
-				),
-				'twitter'  => array(
-					'twitter:card'        => 'card',
-					'twitter:site'        => 'site',
-					'twitter:creator'     => 'creator',
-					'twitter:domain'      => 'domain',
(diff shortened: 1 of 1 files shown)


wpsmort:
@EkoJR I tested this and found one problem.  If the site is set to HTTPS, the og:image:secure_url and og:video:secure_url meta tags are not correctly set and the following notices get output when viewing the site:

Undefined index: thumbnail_1 in aioseop_opengraph.php on line 1667
Undefined index: thumbnail_1 in aioseop_opengraph.php on line 1568
Undefined index: video_1 in aioseop_opengraph.php on line 1667
Undefined index: video_1 in aioseop_opengraph.php on line 1568

The meta tags looks like this (i.e. they are missing the URL for the image/video and the property is incorrect):

<meta property="thumbnail_1" content="og:image:secure_url" />
<meta property="video_1" content="og:video:secure_url" />

To reproduce this, follow these steps:

1. Create a new site that has HTTPS enabled
2. Install the PR code and activate the Social Meta module
3. Visit the front end and you should see the notices

EkoJR:
Nice catch @wpsmort 

I added changes for the an SSL enabled site. Should be good now.

wpsmort:
This is tested and is good.

modules/aioseop_opengraph.php
```diff
@@ -1502,60 +1465,55 @@ function add_meta() {
...
-					'title'             => 'twitter:title',
-					'description'       => 'twitter:description',
-					'twitter_thumbnail' => 'twitter:image',
+					'twitter:card'        => $card,
+					'twitter:site'        => $site,
+					'twitter:creator'     => $creator,
+					'twitter:domain'      => $domain,
+					'twitter:title'       => $title,
+					'twitter:description' => $description,
+					'twitter:image'       => $twitter_thumbnail,
```
arnaudbroes: This code occurs three times. Wouldn't it be better to pull this into one or two separate functions?

arnaudbroes:
@EkoJR just one small comment.

modules/aioseop_opengraph.php
EkoJR: I believe you are referring to the reference array that occurs two times. It was intentionally done that way to make it painfully obvious to remove it when the filter hook is changed.

You are right though. I was split on the idea and you thought of it too, so I'll change it.

arnaudbroes:
This is good now.

PR #2947, opened 2019-10-02
Only output meta descriptions if autogenerate descriptions is enabled

### LINKED ISSUE #2439: Excerpt printed in description when "Autogenerate Descriptions" is disabled
When "Autogenerate Descriptions" is disabled, the plugin outputs the entire excerpt (if the page has one) into the description.

*Steve has confirmed this as well.

OlegApanovich:
There just simple inconsistency between 'Autogenerate Descriptions' and 'Use Content For Autogenerated Descriptions' settings in advanced settings block.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -2901,7 +2901,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			return $description;
 		}
 		if ( ! $description ) {
-			if ( empty( $aioseop_options['aiosp_skip_excerpt'] ) ) {
+			if ( ! empty( $aioseop_options['aiosp_generate_descriptions'] ) && empty( $aioseop_options['aiosp_skip_excerpt'] ) ) {
 				$description = $post->post_excerpt;
 			}
 			if ( ! $description && isset( $aioseop_options['aiosp_generate_descriptions'] ) && $aioseop_options['aiosp_generate_descriptions'] ) {
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -2901,7 +2901,7 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			return $description;
 		}
 		if ( ! $description ) {
-			if ( ! empty( $aioseop_options['aiosp_generate_descriptions'] ) && 'on' === $aioseop_options['aiosp_generate_descriptions'] && empty( $aioseop_options['aiosp_skip_excerpt'] ) ) {
+			if ( empty( $aioseop_options['aiosp_skip_excerpt'] ) ) {
 				$description = $post->post_excerpt;
 			}
 			if ( ! $description && isset( $aioseop_options['aiosp_generate_descriptions'] ) && $aioseop_options['aiosp_generate_descriptions'] ) {
@@ -2911,8 +2911,6 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 						$content = aioseop_do_shortcodes( $content );
 					}
 					$description = $content;
-				} else {
-					$description = $post->post_excerpt;
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


michaeltorbert:
Here's the full PR writeup:

Issue #2439 

## Proposed changes

Checks if autogenerate descriptions is enabled before applying the post excerpt.

## Types of changes

- Improves existing code

## Checklist

- [ ] I've created an appropriate title, descriptive of what the PR does.

## Testing instructions
- Verify that, when autogenerate descriptions is disabled, the excerpt is used in the description.
- After applying the patch, verify that the excerpt is no longer used.
- Make sure no other autogenerate descriptions functionality is broken

arnaudbroes:
I've reproduced the issue and confirmed that this fixes it. This can be code reviewed.

Failure of the build can be ignored as the user's fork doesn't have the updated Travis configuration file in it. I ran Grunt on it locally to make sure there are no issues.

EkoJR:
Changes needed to be applied elsewhere, and PHPUnit was also failing prior to the changes I had made. 

Currently, Grunt is failing due to an update to WPCS.

EkoJR:
PR #2951 (Apply new grunt changes) fixes where grunt was failing at.

arnaudbroes:
@EkoJR I think you might have misunderstood the original issue because the code changes you made broke this again (while @OlegApanovich fixed it). 

To reproduce:

1. Create a Post, add some content and add something into the Excerpt field.
2. Disable Autogenerated Descriptions in the General Settings menu.
3. View the source code. You'll notice that the excerpt is used while it shouldn't output anything.

EkoJR:
@arnaudbroes Excerpts aren't auto generated content, and you need to check "Use Content For Autogenerated Descriptions" in order to skip excerpts (or don't add anything to the excerpt). The content has mostly a fallback since both AIOSEOP Descriptions and Excerpts are usually intended effects because they are manually added by the users, and the post content isn't always reliable for various reasons.

I created an issue to rename the option setting since there is confusion both to users and backend ( 
Change aiosp_skip_excerpt setting label OR db name #2950 ). Conceptually though, I don't think excerpts is part of the auto-gen content, and is an intended effect when a description is manually added.

The change I made was to prevent the excerpt still being added (with "Use Content For Autogenerated Descriptions" checked) if no description was auto-generated from content and . PHPUnit was also failing at the time, which ultimately made me rethink the intent of this PR.

One final note though, I wonder if another issue/feature should be made to adjust the priority of what description source to use, because I can see others preferring auto-gen desc. from content over an excerpt and/or AIOSEOP Desc..

arnaudbroes:
@EkoJR I think you might be confused as to what the expected behaviour of the description and Post Excerpt is but I believe @wpsmort can clear things up.

EkoJR:
@arnaudbroes That's why I mentioned the PHPUnit tests that made me rethink the changes, cause that was a unit test that I had written for truncating strings. I remember being told specifically that Excerpts are manual descriptions and should not be truncated. Conceptually, changing excerpts to be considered auto-gen would itself be fundamentally skewed.

Here was some mentions of manual descriptions (both AIOSEOP Desc. and Excerpts were considered manual). https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/1395#issuecomment-355035771 ...but I remember most of the conversations occuring on Slack because the amount of trouble the PR was causing.

What seems to be more of a problem is the UX on the General Settings page. On the backend it's also polarized.

I did mark this as Needs Discussion for a meeting since there seems to be different thoughts on this.

EkoJR:
There's also a good possibility users do not want to use AIOSEOP Descriptions and Auto-Gen content, and rely on WP Excerpts for printing descriptions. If they don't add an Excerpt, they will end up with Auto-Gen content. Which they may not want with the way it generates on their site.

OlegApanovich:
I added new commits to PS trying to pay attention on commit @EkoJR  about strict comparator with 'aiosp_generate_descriptions' option and fix unit tests that was broken by new fixis.

wpsmort:
This is tested and is good.

Issue #2080, opened 2018-11-25
API - aioseop_attachment_size filter doesn't always work


michaeltorbert:
Reported here: https://wordpress.org/support/topic/image-size-for-sharing-on-social-media/

From looking at the code, the filter only fires on get_the_image_by_attachment()
I *think* that the image settings have to be configured to attachment, and then on the post itself, don't select an image.

onetarek:
While looking at the function `get_the_image_by_attachment()` I have found an unnecessary line of code
`$alt   = trim( strip_tags( get_post_field( 'post_excerpt', $id ) ) );`

https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/master/admin/aioseop_module_class.php#L1627

michaeltorbert:
Needs reproducing instructions.

wpsmort:
Steps to reproduce:

1. Create a post with an attached image, an image in the content, and a featured image
2. Go to All in One SEO > Social Meta > Image Settings and set the OG:Image Source as Default Image, check the box for Use Default If No Image Found and upload a default image that's large enough so that a large, medium and thumbnail size image are created
3. View the source code of the post and verify that the og:image is the full size default image
4. Use the filter to set the returned image size as thumbnail
5. View the source code of the post and see if the og:image is displaying the thumbnail
6. Go back to the Image Settings and repeat these steps with the other selections in the Select OG:Image Source drop down

By following these steps above, you'll see that the only images affected by the filter are the first attached image and the first image in content.

The filter needs to also affect the default image, featured image and first available image.

wpsmort:
Issue moved to [awesomemotive/aioseop-pro #901](https://github.com/awesomemotive/aioseop-pro/issues/901) via [**ZenHub**](https://www.zenhub.com/)

PR #2979, opened 2019-10-25
Add aioseop_attachment_size filter to all image settings

OlegApanovich:
…ta module (#2080)

Issue #2080 

## Proposed changes

Extend aioseop_attachment_size filter to other image options in social meta module.

## Types of changes

_Delete those that don't apply_

- Improves existing code

## Checklist

- [x ] I've selected the appropriate branch (ie x.y rather than master).


michaeltorbert:
Maybe I'm missing something but this seems to do nothing other than remove our own API filter hooks.

OlegApanovich:
It is not removing filter hooks @michaeltorbert It is replaced them to another location where they can cover all image settings of "Select OG:Image Source" options. Right now filter work only for a 'first attached image' option, but after I replaced filter to another location and add additinal function filter began to work to other options too.
In comment to issue @wpsmort sayed "By following these steps above, you'll see that the only images affected by the filter are the first attached image and the first image in content. The filter needs to also affect the default image, featured image and first available image." I think he said it because thought it difficult to extend filter to options that parse the content searching img tag. But my PR cover all possible setting options

michaeltorbert:
@OlegApanovich Yes, I hadn't seen the second file, which is why I already deleted my comment.

EkoJR:
In the past, we've experience performance issues with converting attachment urls to post ids, and there's been a few related issues concerning how attachment urls operate. There's a function that is already available that serves the same purpose as your `aioseop_get_attachment_id( $url )` in the most optimal way that was found possible.

It is `aiosp_common::attachment_url_to_postid( $url )` in `inc/aiosp_common.php`.

Looking at your query_args, I believe that was one of the queries I had tested. Unfortunately, using any meta query search with attachments has big impact on large sites.

OlegApanovich:
@EkoJR  in my last commit I replace my new function with aiosp_common::attachment_url_to_postid( $url )
Also I improve it trying to escape performance issue that function produce on large sites.
I test it on a 20k post wordpress installation and existing call of this function in
AIOSEOP_Graph_Organization::get_logo_id() has huge improvement in a first invocation when cash and transient data do not exist yet. I add some benchmarks in attached files.
Screenshot from 2019-11-08 10-00-29
Screenshot from 2019-11-08 10-07-00

arnaudbroes:
@EkoJR let me know what your thoughts are on this. I believe you were still reluctant on approving this.

EkoJR:
I did a comparison on the benchmarks, and it seems like that the changes added degrades the performance. The initial load is the main concern (without being able to use transient data), and found that the Time is nearly doubled and the Memory increases by a good margin.

Originally, the code was designed to resolve user reports with timeout errors. At the time, it was only the Sitemap that would use the function, and it still heavily relies on it; it's also what I used to test on. I had also found any meta_query would significantly degrade performance, which I had mentioned as well.

The 2nd query I haven't fully tested yet, but seems interesting. It's still heavier than query_1, but I need to load a site with more edited images in use (it was only called twice on current benchmarks), and compare benchmarks with the code that was commented out. As a public release, query_2 should remain disabled though.

## Total Time/Memory

The initial load times seem to be...

Time: 91.19% (degrade) increase.
Memory: 11.95% (degrade) increase.

### Parent/Base Commit

Time | Memory
---- | ------
67,511,710 | 1,239,914,456 
66,606,753 | 1,236,776,296

### PR

Time | Memory
---- | ------
129,073,847 | 1,388,032,088
66,911,419 | 1,236,918,992

## Screenshots

### Parent/Base Commit

#### Initial Load

Master 01 Time
Master 01 Memory

#### Reload (2nd load)

Master 02 Time
Master 02 Memory

### PR Branch

#### Initial Load

PR 01 Time
PR 01 Memory

#### Reload (2nd load)

PR 02 Time
PR 02 Memory

OlegApanovich:
Yes @EkoJR  there defenetely some perfomence degradetion in All_in_One_SEO_Pack_Sitemap::get_images_from_post() in my previos commit I do not paid attention on it and I try to improve it in my last commit that I added today.

In our benchmarks key function right now is aiosp_common::set_transient_url_postids()  that has heavy database query that load all url for all existent attachments in one array.
Right now we use aiosp_common::set_transient_url_postids() function in 2 places:
1. All_in_One_SEO_Pack_Sitemap::get_images_from_post()
2. AIOSEOP_Graph_Organization::get_logo_id()

And in my first commit I add use of aiosp_common::set_transient_url_postids() in third place to resolve initial issue for I created it pull requers.
3. All_in_One_SEO_Pack_Opengraph::add_meta()

Right now I think we already use aiosp_common::set_transient_url_postids() function improperly cos we have two different task when we use in
1. All_in_One_SEO_Pack_Sitemap::get_images_from_post()
2. AIOSEOP_Graph_Organization::get_logo_id()

In our first place with sitemap we have mass action when we need all urls and in a second place with social media we need only one url to convert to id.
I think it is too expensive to load all attachment urls in one query even with static cache and transient functionality in later use.
I think we must divide using aiosp_common::set_transient_url_postids() function on two separate 
1. aiosp_common::convert_image_url_to_id_bulky()
2.aiosp_common::convert_image_url_to_id_individually()

Firt we will use for a single url to id attchment query and second for a multiple url to id attchment query and we must add optionality to function that will be determine 1 or 2 query use for this task ( first query for manin images, second query for custom sizes )

All of it I already implemented in my last commit.
Below I had some benchmarks that demonstrate performance improvement especially for already existing AIOSEOP_Graph Organization::get_logo_id() function.


All_in_One_SEO_Pack_Sitemap::get_images_from_post() - **master branch**
**Initial Load**
master-sitemap

All_in_One_SEO_Pack_Sitemap::get_images_from_post() - **pr branch**
**Initial Load 1 query**
branch-1query-sitemap

All_in_One_SEO_Pack_Sitemap::get_images_from_post() - **pr branch**
**Initial Load 2 query**

branch-2query-sitemap

AIOSEOP_Graph_Organization::get_logo_id() - **master branch**
**Initial Load**

opengraph-initload-master

AIOSEOP_Graph_Organization::get_logo_id() - **pr branch**
**Initial Load**

opengraph-initload-branch

Even if you think that use second query in sitemap case is not rational we can deactivate it just set second option to false. Right now in my last commit I set it as true.

Please review my pr with my last commit again.

arnaudbroes:
@OlegApanovich I'm going to close your PR for now because we're currently in the midst of cleaning up things and we'll probably revisit the Social Meta module in its entirety in the near future, but it's linked to the issue so we'll definitely have a look at this again when we get to it.

Issue #2980, opened 2019-10-26
i18n - Make strings untranslatable again


arnaudbroes:
I'm rather sure these should not be translatable - https://github.com/semperfiwebdesign/all-in-one-seo-pack/blob/8c85891893863c38a5b37e14077cc50499192b01/inc/schema/graphs/graph-webpage.php#L90-L100

Specifically the arguments in the `get_the_date` function.

PR #2987, opened 2019-10-28
Make strings untranslatable 

michaeltorbert:
Issue #2980

## Proposed changes

We inadvertently made some schema.org output translatable. This PR fixes that.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions
- Test to make sure the output is both correct and unchanged in schema.org for date archives (year, month, and day).


wpsmort:
This is tested and is good.

arnaudbroes:
@wpsmort I had to make a small change here because the first part of each line should be translatable; the second part had to be made untranslatable again. I might have explained badly during our call. In any case, this will just need a quick retest. Switch to another language and make sure the strings are still localized.

wpsmort:
This is tested and is good.

Issue #2937, opened 2019-09-26
Schema - Slow query with thousands of users


michaeltorbert:
Reported here: https://wordpress.org/support/topic/slow-query-on-shop-with-thousands-of-users/

> I’m running a large shop with around 12000 registered customers and we’re investigating slow db performance. We’ve installed “query monitor” and one of the slow queries is linked to your plugin.
Is there any way to optimise this query to run more efficiently on a large shop?
Screenshot: https://www.dropbox.com/s/5jndd0y4k8y28eg/Bildschirmfoto%202019-09-26%20um%2013.13.31.png?dl=0
Query is in
wp-content/plugins/all-in-one-seo-pack/aioseop_class.php:1093

		$user_args = array(
			'role__in' => array(
				'administrator',
				'editor',
				'author',
			),
			'orderby'  => 'nicename',
		);
		$users = get_users( $user_args );

![fZbbMYVA](https://user-images.githubusercontent.com/17364998/65694792-6a759700-e044-11e9-87f6-cddf998fc0df.png)

JulianMar:
Is there a workaround for this?

notmike101:
Another example of long query times for this on a shop with nearly 6 thousand users.

![image](https://user-images.githubusercontent.com/7198316/66776855-3db2e380-ee7c-11e9-87e3-d7993ab8c6e9.png)

If anyone knows of any fixes for this, please let me know.

jmock:
Same here. We have 140,000 users and I had to disable the plugin because it was melting our database this morning.

leonelectro:
The WP database architecture makes these checks for users of certain roles expensive. You can mitigate this problem with caching, but you still have to generate the cache values regularly. You could do that in a cron.

This code seems to be pulling a list of admins for the sake of default options on every request. Perhaps the defaults can be generated once on activation or only on the wp-admin side on a particular configuration page.

A temporary fix would be to stay on version 3.1.x. After 9 patches to 3.2 and a month gone by, it's straining the concept of temporary.

michaeltorbert:
I’m assigning this to 3.3. @arnaudbroes or @EkoJR can one of you have a look?

arnaudbroes:
I've confirmed the behaviour.

michaeltorbert:
At a minimum, limit this code to the the settings page.

wpsmort:
[...]

PR #2995, opened 2019-10-30
Force manual input for Schema.org Person when > 20 users

arnaudbroes:
Issue #2937

## Proposed changes

Fixes an issue where a query for the Schema.org markup settings slows down website performance. 
The code for this now only runs when the user is on a plugin page in the admin panel (couldn't limit it to our toplevel page specifically) and when there is a max. of 20 users.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

First, make sure you have 20 users (a few of them Admin, Editor or Author) and then go to the General Settings page and check that the dropdown for Schema.org Person Name is populated with these users.

Then, add another user and make sure that the plugin now only allows manual input.

## Further comments

Obviously, the user treshhold can be changed if needed.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -1126,23 +1126,35 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 			),
 		);
 
-		$user_args = array(
-			'role__in' => array(
-				'administrator',
-				'editor',
-				'author',
-			),
-			'orderby'  => 'nicename',
-		);
-		$users     = get_users( $user_args );
+		global $pagenow;
+		if ( 'admin.php' === $pagenow ) {
+			// Person's Username setting.
+			$this->default_options['schema_person_user']['initial_options'] = array(
+				0  => __( '- Select -', 'all-in-one-seo-pack' ),
+				-1 => __( 'Manually Enter', 'all-in-one-seo-pack' ),
+			);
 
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good.

arnaudbroes:
Also, let's leave the issue open after this is merged so we can improve this in 3.4.

aioseop_class.php
```diff
@@ -1126,23 +1126,35 @@ function __construct() {
...
-		// Person's Username setting.
-		$this->default_options['schema_person_user']['initial_options'] = array(
-			0  => __( '- Select -', 'all-in-one-seo-pack' ),
-			-1 => __( 'Manually Enter', 'all-in-one-seo-pack' ),
-		);
-		foreach ( $users as $user ) {
-			$this->default_options['schema_person_user']['initial_options'][ $user->ID ] = $user->data->user_nicename . ' (' . $user->data->display_name . ')';
+			global $wpdb;
+			$user_count = $wpdb->get_var( "SELECT COUNT(*) FROM $wpdb->users" );
+			if ( 20 < $user_count ) {
```
EkoJR: Are we sure 20 is large enough? I was thinking 100, or at minimum 50.

arnaudbroes: 50 sounds good. I've updated it.

Issue #3023, opened 2019-11-12
Bug- Full content appearing in Preview Snippet and Description placeholder


michaeltorbert:
Reported here: https://wordpress.org/support/topic/error-autogenerate-descriptions/

> when i activate Autogenerate Descriptions, it generate the complete post in description box NOT the first 160 characters! see the screenshots below:
> screen 1: https://yadi.sk/i/DZDOJDD_zasxzg
> screen 2: https://yadi.sk/i/1vgN-mCm9qkJaA
> 
>

arnaudbroes:
The character count is based/relies on the preview snippet so this would be an issue with the latter.

arnaudbroes:
Reported again here - https://wordpress.org/support/topic/autogenerate-descriptions-error/#post-12149551

wpsmort:
@arnaudbroes I've now been able to get access to a site to troubleshoot and reproduce this.

The full content from a post appears in the Preview Snippet and Description field placeholder only when using the Classic block. This only happens with certain type of formatted content.

This bug first occurs in AIO v3.3.2.

Here are the steps to reproduce:

1. Install AIO v3.3.2 on a new site and enable Autogenerate Descriptions
2. Create a new post using the Block editor
3. Add the Classic block and enter a large amount of text
4. Publish the post and check the Preview Snippet and Description field placeholder, both should show the content truncated to around 160 characters which is correct behavior
5. Edit the post and add some additional text to the Classic block and style it as a blockquote
6. Check the Preview Snippet and Description field placeholder, both should show the full content not truncated.

This only seems to happen with certain formatted text, such as blockquotes, aligned (left, center, right) text, headings and preformatted text.

PR #3064, opened 2019-11-26
Improve regex pattern to remove new lines

arnaudbroes:
Issue #3023

## Proposed changes

This improves the regex pattern to remove new lines.

## Types of changes

- Improves existing code

## Checklist

## Testing instructions

1. Reproduce the original issue as seen here- https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/3023#issuecomment-558285178. You'll want to add at least 2 paragraphs into the Classic Editor block.
2. Install this PR and see if it fixes it. 

## Further comments

@wpsmort this was rather hard to troubleshoot but I believe the regex pattern wasn't able to filter the new lines that the Classic Editor block uses. 

Also during test, note that the Preview Snippet will only update after you unfocus/leave the Classic Editor block and click anywhere else. This behaviour is normal and expected.

### DIFF
--- a/js/admin/aioseop-preview-snippet.js
+++ b/js/admin/aioseop-preview-snippet.js
@@ -165,7 +165,7 @@ jQuery(function($){
 		// Remove all HTML tags.
 		content = content.replace(/(<[^ >][^>]*>)?/gm, '');
 		// Remove all line breaks.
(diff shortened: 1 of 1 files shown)


wpsmort:
This is tested and is good.

Issue #3090, opened 2019-12-26
Bug - addl-sitemap returns 404 when homepage is set to latest posts and static pages are not cleared


arnaudbroes:
When the homepage is set to display the latest posts, but the static pages below it aren't cleared, the addl-sitemap file returns a 404 error because the homepage is listed under pages. 

Should be an easy fix (I wrote the original code).

PR #3105, opened 2020-01-10
Fix addl-sitemap 404 when static pages are not cleared

arnaudbroes:
Issue #3090

## Proposed changes

Fixes an issue where the addl-sitemap file returns a 404 error when the homepage is set to display the latest posts but the static pages are still set as well.

## Types of changes

## Checklist

## Testing instructions

1. Install the latest version of the plugin and reproduce the original issue by first setting a static homepage and blogpage, and then setting the homepage back to latest posts. 
Do not clear the dropdown selectors for the static pages. If they are automatically cleared (this may happen), then go to the database via PhpMyAdmin > wp_options table and set "page_for_posts" and "page_on_front" to the ID of two random pages.

2. Install the PR, do the same as the above and confirm the addl-sitemap no longer returns a 404 page but instead lists the homepage.

You'll also want to play around with additional pages (set in XML Sitemap menu) and scenarios where just one of the two static pages is set since I had to rewrite the entire function somewhat.

### DIFF
--- a/modules/aioseop_sitemap.php
+++ b/modules/aioseop_sitemap.php
@@ -2207,8 +2207,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Sitemap' ) ) {
 
 			$files = apply_filters( 'aioseop_sitemap_index_filenames', $files, $prefix, $suffix );
 
-			// Remove Additional Pages index if all pages are static and no extra pages are specified.
-			if ( ! $this->does_addl_sitemap_contain_urls() ) {
+			if ( ! $this->should_addl_sitemap_exist() ) {
 				$page_to_remove = array( get_site_url() . '/addl-sitemap.xml' );
 				$files          = $this->remove_urls_from_sitemap_page( $files, $page_to_remove );
 			}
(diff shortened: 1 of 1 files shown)


wpsmort:
@arnaudbroes I found two instances that need fixing.

The first is when you set a static Posts page but don't set a static Home page.  Here's the steps to reproduce:

1. Set the Your homepage displays to A static page and leave the Homepage dropdown blank and set the Posts page dropdown to a page
2. Make sure you have Sitemap Indexes enabled and view the sitemap index, you'll see that addl-sitemap.xml isn't there
3. Try browsing to addl-sitemap.xml and it should come up and display the home URL
4. Disable Sitemap Indexes and you should see the home URL.

The second is the same as above but setting the Your homepage displays to Your latest posts when there's a page ID set for page_for_posts.  Here's the steps to reproduce:

1. Set the Your homepage displays to A static page and leave the Homepage dropdown blank and set the Posts page dropdown to a page
2. Set the Your homepage displays to Your latest posts and verify that there's a valid page ID in the  page_for_posts option name in the database
3. Repeat steps 3 and 4 above and you'll get the same result.

arnaudbroes:
@wpsmort this can be tested again.

wpsmort:
This is tested and is good.

Issue #3080, opened 2019-12-17
Bug - Duplicate robots meta tag inside robots meta tag when SEO for Post is disabled


arnaudbroes:
Reported here - https://wordpress.org/support/topic/be-careful-with-product-pages/
 
To reproduce this, disable SEO for the Post and toggle either NOINDEX, NOFOLLOW or both.

arnaudbroes:
Reported again here - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/3095

I'm bumping this to 3.3.5 since it's been reported a few times now.

wpsmort:
Reported again by Romain by email today.

PR #3106, opened 2020-01-10
Fix nested robots meta when SEO is disabled

arnaudbroes:
Issue #3080

## Proposed changes

Fixes a bug where one robots meta tag is nested inside another when SEO for the Post is disabled.

## Types of changes

## Checklist

## Testing instructions

1. To reproduce this, disable SEO for a Post and toggle either NOINDEX, NOFOLLOW or both. Check the source code. You'll notice one robots meta tag is nested inside another.

2. Install the PR and confirm the fix.

### DIFF
--- a/aioseop_class.php
+++ b/aioseop_class.php
@@ -4382,10 +4382,10 @@ class All_in_One_SEO_Pack extends All_in_One_SEO_Pack_Module {
 		if ( ! $this->is_page_included() ) {
 
 			$aioseop_robots_meta = new AIOSEOP_Robots_Meta();
-			$robots_meta         = $aioseop_robots_meta->get_robots_meta();
+			$robots_meta         = $aioseop_robots_meta->get_robots_meta_tag();
 
 			if ( ! empty( $robots_meta ) ) {
-				echo sprintf( '<meta name="robots" content="%s"', esc_attr( $robots_meta ) ) . " />\n";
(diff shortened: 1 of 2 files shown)


wpsmort:
This is tested and is good.

aioseop_class.php
```diff
@@ -4382,10 +4382,10 @@ function wp_head() {
 		if ( ! $this->is_page_included() ) {
 
 			$aioseop_robots_meta = new AIOSEOP_Robots_Meta();
-			$robots_meta         = $aioseop_robots_meta->get_robots_meta();
+			$robots_meta         = $aioseop_robots_meta->get_robots_meta_tag();
 
 			if ( ! empty( $robots_meta ) ) {
-				echo sprintf( '<meta name="robots" content="%s"', esc_attr( $robots_meta ) ) . " />\n";
+				echo $robots_meta;
```
EkoJR: I'm surprised this didn't get flagged as needing an `esc_*` function (even though there is one in `AIOSEOP_Robots_Meta::get_robots_meta_tag_helper()`). Not sure if this should be changed or not.

arnaudbroes: No need since it's already being escaped in `AIOSEOP_Robots_Meta::get_robots_meta_tag_helper()`. Otherwise we're just doing it twice.

Issue #3102, opened 2020-01-09
Bug - Preview Snippet crashes IE11 due to browser bug


arnaudbroes:
Reported here - https://wordpress.org/support/topic/ie11-crashes-on-post-page/

There's a bug in IE11 where the browser constantly triggers any input event on an input or textarea field that has a placeholder set. This ultimately causes the browser to crash.

I have reproduced the issue in IE11 and confirmed that changing the `input` event to the `keyup` event fixes it.

PR #3107, opened 2020-01-11
Fix input event bug in IE11

arnaudbroes:
Issue #3102

## Proposed changes

Fixes a critical bug where IE11 crashes when an input event is triggered. 

## Types of changes

## Checklist

## Testing instructions

1. Reproduce the issue in IE11 by navigating to any Post/Page and trying to type in the editor. The page should immediately freeze up. @wpsmort check in on Team Lead for cross-browser testing tool. Worst case scenario we can do a call and I can show it to you.

2. Install this PR and confirm the fix.


wpsmort:
This is tested and is good.

Issue #1656, opened 2018-04-16
Social Meta - OG: & Twitter:description still generated while autogeneration is disabled


arnaudbroes:
Reported here and confirmed by me - https://github.com/semperfiwebdesign/all-in-one-seo-pack/issues/739. I'm guessing that the code for the 'Use Content For Autogenerated OG Descriptions' setting is running regardless or whether or not that setting is activated/deactivated. I'll troubleshoot to see if this was always the case or to see when it broke.

artevodesign:
Hi! Any development on this issue, please?

arnaudbroes:
@artevodesign no updates yet.

michaeltorbert:
@arnaudbroes See when this broke.

arnaudbroes:
@wpsmort the above code (see commit) fixes this issue but the plugin still outputs a OG:description when there is an excerpt (when"Autogenerate descriptions" is disabled). 

Is that the expected behaviour or should there be no OG:description similar to how there is no description unless the user entered one? I'd go with the latter but I'd like your thoughts.

wpsmort:
@arnaudbroes The current behavior is correct.  There's no setting for Autogenerate OG Description in the Social Meta module and it's shouldn't use the Autogenerate Descriptions setting in the General Settings.  Here's why we did it this way:

Social media sites such as Facebook need some basic OG tags in order for a share to be displayed.  Without these necessary tags, when someone shares content, Facebook may return an error and not display the share.  When we developed the Social Meta module, we decided that we'd provide the minimum required OG tags so that sharing would work.  This meant that we had to output an OG:description.  We decided that whatever happened, we'd display this based off of the SEO description, excerpt or content.  

Later on, we added the Use Content For Autogenerated OG Descriptions option to mirror the same option in General Settings.  This was because of WooCommerce and the way they have a Product Short Description (Excerpt) and Product Content and many users wanted to choose which to use for social sharing.  

FYI... I've updated the doc here to make it clearer what we're doing - https://semperplugins.com/documentation/social-meta-settings-individual-pagepost-settings/#description

arnaudbroes:
Awesome. Submitted the PR.

arnaudbroes:
I turned out to be incorrect; the current behaviour is the expected behaviour so this doesn't need a PR.

PR #3118, opened 2020-02-03
Don't generate OG:description from content when setting is disabled 

arnaudbroes:
Issue #1656

## Proposed changes

Fixes an issue where the OG:description still uses the content when the "Use content for autogenerated OG:descriptions" setting is disabled.

## Types of changes

## Checklist

## Testing instructions

1. Install the latest version of the plugin, enable the Social Meta menu and make sure the "Use content for autogenerated OG:descriptions" setting is disabled.
2. Create a new post with some content. Check the source code and confirm that the post has an OG:description with the content of the post in it.

Then, install this PR and confirm this is no longer the case and that the setting's value is respected. When the post has an excerpt, it should still have an OG:description like we discussed.

### DIFF
--- a/modules/aioseop_opengraph.php
+++ b/modules/aioseop_opengraph.php
@@ -1177,7 +1177,7 @@ if ( ! class_exists( 'All_in_One_SEO_Pack_Opengraph' ) ) {
 
 					$description = $post->post_excerpt;
 
-					if ( $this->options['aiosp_opengraph_generate_descriptions'] || empty( $description ) ) {
+					if ( $this->options['aiosp_opengraph_generate_descriptions'] && empty( $description ) ) {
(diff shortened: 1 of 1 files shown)


wpsmort:
@arnaudbroes This has changed the login so it no longer matches the correct behavior.  Now the OG description is only output if Use Content For Autogenerated OG Descriptions is enabled.

Here's the correct behavior matrix:

Autogenerate Descriptions: **Disabled** = OG description is output based on 1, 2 and 3 below
Autogenerate Descriptions: **Enabled** = OG description is output based on 1, 2 and 3 below

1. OG description uses manually entered SEO description
2. If there's no manually entered SEO description, then the OG description uses the Excerpt
3. If there's no Excerpt, then the OG description uses the Content

If Use Content For Autogenerated OG Descriptions is enabled then the OG description is output based on 1 and 2 below

1. OG description uses manually entered SEO description
2. If there's no manually entered SEO description, then the OG description uses the Content

Issue #2905, opened 2019-09-13
New Feature - Add nofollow + title attribute field for links in Gutenberg Editor


arnaudbroes:
As announced here - https://webmasters.googleblog.com/2019/09/evolving-nofollow-new-ways-to-identify.html

I believe we should be able to do this by adding custom format actions to the editor toolbar.

michaeltorbert:
We should add nofollow to this. Also, we should decide whether to do it as a button or in the little carrot dropdown which has open in a new tab, or both.
<img width="468" alt="Screen Shot 2019-09-13 at 9 44 42 AM" src="https://user-images.githubusercontent.com/17364998/64867606-b881a800-d60b-11e9-8f77-017fdd214752.png">

arnaudbroes:
I'm bumping this for now since there's no proper way to add this at the moment (we'd have to unregister the current link format, adjust it and then register it again AND ship a bunch of Gutenberg code), but I'm subscribed to the issue in the Gutenberg repo. I might also look into adding a filter myself and submitting it as a PR.

arnaudbroes:
https://github.com/WordPress/gutenberg/issues/17556

PR #3138, opened 2020-02-15
Add support for "nofollow" and "sponsored" link attributes

arnaudbroes:
Issue #2905 

## Proposed changes

This PR allows users to add the `nofollow` & `sponsored` attributes (and `title` attribute for the Classic Editor) to links/anchor tags.

As the scripts for the Gutenberg Editor are written in ESNext, this PR also adds support for Webpack & Babel.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist

## Testing instructions

Both the Classic Editor and Gutenberg Editor have to be tested.

**Classic Editor**

1. Install and activate the Classic Editor plugin, together with this PR;
2.  Create or edit a Post, write some text and add a link. Click on the gear icon to trigger the modal. You should now be able to add a `title`, `nofollow` and `sponsored` attribute. Test these options and check the source code;
3. Do the same but instead of adding a link to existing text, click on a blank spot and add a link through the anchor button in the toolbar and define the anchor text inside the modal;
4. Test (3) & (4) in the HTML editor (Text tab). It's possible that a console error occurs that is unrelated to our plugin (a bug in the Classic Editor). This can be ignored.

**Gutenberg Editor**

1. Deactivate the Classic Editor;
2. Create or edit a Post, write some text and add a link. Clicking on the gear icon should reveal two additional options to add a `nofollow`and `sponsored` attribute. Note that there is no field for a `title` attribute in the Gutenberg Editor. Test these options and check the source code;
3. Create a Classic Editor block and do (3) and (4) of the Classic Editor testing instructions again.

## Further comments

I ended up spending a bit more time on this since I had to fix some bugs in the original script for the Classic Editor.


wpsmort:
@arnaudbroes I found three problems with this as follows:

Classic editor:
When you're typing a link in the classic editor and you highlight some test and click the Link icon and type the name of a post or page then select it, if you then click the settings icon, the URL field has been wiped.  If I then type the name of the page again and select it from the results, and then click Add Link nothing gets added.

Block editor:
Similar to above, if you select text in a Paragraph block and type a URL or a page name, and then try to set the nofollow or sponsored toggle it wipes the URL or the toggle.  You have to apply the URL first, then go back and select the URL settings.

Block editor:
When you convert a Classic Editor block to a Paragraph block , the link settings don't follow over so when you look at the link setting, neither the nofollow or sponsored options are toggled.

arnaudbroes:
@wpsmort as discussed, the third issue would be rather difficult to add support for but I'll see if I can fix the other two issues in the original scripts.

arnaudbroes:
@wpsmort I fixed the first two issues. Third one is beyond the scope of this PR and technically very difficult to implement (due to the lack of filters in Gutenberg) as discussed via Skype.

I didn't notice during our call but I found out (with regards to the first issue) that the URL was being copied, but to the search field instead of the URL field. I've made sure it's copied to both fields now as I thought that made the most sense. If you believe that the user input shouldn't be copied to the search field, then let me know and I'll quickly edit that.

The second issue was caused by a function call in the original script that was lacking the nofollow & sponsored state parameters.

wpsmort:
This is tested and is good.

arnaudbroes:
@EkoJR any idea why Travis isn't able to locate Grunt on this PR?

EkoJR:
@arnaudbroes 

message: 'Unexpected token: name «inputField», expected: punc «;»',
filename: 'aioseop-count-chars.js',
line: 13,
col: 5,
pos: 251

arnaudbroes:
@wpsmort @benjaminprojas this can be tested again! :-)

wpsmort:
@arnaudbroes I found two bugs after testing this again:

The first is in the classic editor and happens when you have two links in the same text.  The second link overrides the first link.  See screencast here - https://a.supportally.com/eQrgKC

The second is in the Classic block in Gutenberg and happens again when you have two links in the same text.  When adding the second link, the first link is removed.  And when clicking on the second link that also gets removed.  See screencast here - https://a.supportally.com/h83eiF

arnaudbroes:
@wpsmort this can be tested again. Fixed the first issue manually and the second by updating the rest of the script to the latest version in WP Core.

wpsmort:
This is tested and is good.

wpsmort:
@arnaudbroes Does this need any additional code review or is it good to merge?

arnaudbroes:
@wpsmort this is good to go.

benplum:
How can this be disabled? Your custom scripts are breaking other plugins that also modify the modal.

arnaudbroes:
@benplum You can disable it for the Classic Editor using this filter - 

```
add_action( 'wp_enqueue_editor', 'aioseop_dequeue_script', 9999999, 0 );

function aioseop_dequeue_script() {
	wp_dequeue_script( 'aioseop-link-attributes-classic-editor' );
}
```

benplum:
Thanks, this helps. Can it be disabled globally? ACF link fields that also use the modal are also being modified.

arnaudbroes:
> Thanks, this helps. Can it be disabled globally? ACF link fields that also use the modal are also being modified.

If this the same modal that the Classic Editor uses, then it should also be disabled for ACF. Is it still showing?

Issue #3122, opened 2020-02-06
New Feature - Sitemap priority setting for individual posts


arnaudbroes:
https://app.asana.com/0/1160816225937945/1160816225938024

Issue #2730, opened 2019-07-19
Do Not Override no longer fetches default prio/freq for taxonomies


arnaudbroes:
> When you override the priority and frequency for a specific taxonomy, then the priority and frequency for all other taxonomies is empty. Here's the steps to reproduce:
> 
> Go to the XML Sitemap settings and in the Priorities section, choose "Select Individual" next to Taxonomies
> Now select "0.1" next to Categories Taxonomy
> Next, in the Frequencies section, choose "Select Individual" next to Taxonomies
> Now select "Hourly" next to Categories Taxonomy and click Update Sitemap
> View the sitemap and you should see that all categories are set based on the settings above but all other taxonomies are missing the priority and frequency.

This does not apply to CPTs.

PR #3140, opened 2020-02-17
Sitemap priority/frequency for individual posts/terms

arnaudbroes:
Issue #3122 & #2730

## Proposed changes

Adds support for overriding the global or default priority/frequency values for individual posts and taxonomy terms.

Also fixes a bug where taxonomies weren't able to fetch their default priority/frequency values, refactors some code and removes a few unused functions.

## Types of changes

- New feature (non-breaking change which adds functionality)
- Improves existing code

## Checklist

## Testing instructions

1. Download this PR, build Pro with the latest 3.3.5 version, install & activate it and then add a few different CPTs that are assigned to a few custom taxonomies (WooCommerce Product Categories for instance).
2. Go to the Edit Post screen and override the priority/frequency. Then check the sitemap and confirm these are the values that are displayed. Do the same for the custom taxonomies.
3. Go to the XML Sitemap module and override the default values with a different value. At this point, the individual priority/frequency values should still take precedence. If you remove these by setting the fields to "Do Not Override", the plugin should fallback on the global values.
4. Remove the global values and confirm that the default values are used (check code snippets below to see what these are). Also check this for taxonomies. Previously, taxonomies displayed two blank fields. They should now be populated with the default values (#2730).
5. Make sure the tooltips are populated with the help text in the metabox.

`public function get_default_priority() {
			$defaults = array(
				'homepage'   => '1.0',
				'blog'       => '0.9',
				'sitemap'    => '0.8',
				'post'       => '0.7',
				'archive'    => '0.5',
				'author'     => '0.3',
				'taxonomies' => '0.3',
			);`

`		public function get_default_frequency() {
			$defaults = array(
				'homepage'   => 'always',
				'blog'       => 'daily',
				'sitemap'    => 'hourly',
				'post'       => 'weekly',
				'archive'    => 'monthly',
				'author'     => 'weekly',
				'taxonomies' => 'monthly',
			);`


wpsmort:
@arnaudbroes This works fine for taxonomies but does not work for post types.  There are two problems as follows:

1.  When I override only the Priority, there's nothing output for the Frequency.  See screenshot below.

2.  When I override only the Frequency, there's no change.  The Frequency only gets overridden as long as the Priority is overridden too.  See screenshot below.

3.  For custom post types with an archive, if you override the very first post in the post type then it also overrides the post type archive too.  See screenshot below.

Screen Shot 2020-02-17 at 4 50 52 PM
Screen Shot 2020-02-17 at 4 51 30 PM
Screen Shot 2020-02-17 at 4 59 09 PM

arnaudbroes:
@wpsmort as discussed, (1) and (2) were caused by a typo and are fixed.

(3) was a pre-existing bug that has now also been fixed. 
All archive pages (post, date, author) should be using the default values. Post archive pages should be using the last modified timestamp of the last modified post; date and author archives don't have this yet so I'm opening a separate issue for that - #3142

We'll discuss the other issues later during the L10.

wpsmort:
This is tested and is good.

modules/aioseop_sitemap.php
```diff
@@ -3497,153 +3467,176 @@ public function get_prio_calc( $date, $stats ) {
...
-			return $archives;
+			return $date_archives;
 		}
 
 		/**
-		 * Get Archive Priority from Posts
+		 * Returns a post archive sitemap entry if the post type supports one.
 		 *
-		 * Generate sitemap priority data for archives from an array of posts.
+		 * @since   ?
```
benjaminprojas:
```suggestion
```

modules/aioseop_sitemap.php
```diff
@@ -3497,153 +3467,176 @@ public function get_prio_calc( $date, $stats ) {
 		}
 
 		/**
-		 * Get Date Archive Priority from Posts
+		 * Returns all date archive entries for the sitemap.
 		 *
-		 * Generate sitemap priority data for date archives from an array of posts.
+		 * @since
```
benjaminprojas:
```suggestion
```

modules/aioseop_sitemap.php
```diff
@@ -3765,26 +3770,26 @@ public function get_comment_count_stats( $posts ) {
 		}
 
 		/**
-		 * Get Priority from Posts
+		 * Gets the last modified timestamp, priority and frequency for posts.
 		 *
-		 * Generate sitemap priority data from an array of posts.
-		 *
-		 * @since ?
+		 * @since   ?
```
benjaminprojas:
```suggestion
```

modules/aioseop_sitemap.php
```diff
@@ -4500,72 +4522,35 @@ public function set_post_args( $args ) {
 		}
 
 		/**
-		 * Get Data Archive Priority Data
+		 * Returns all sitemap entries for a custom post type.
 		 *
-		 * Return sitemap data for date archives.
+		 * @since   ?
```
benjaminprojas:
```suggestion
```

modules/aioseop_sitemap.php
```diff
@@ -4636,6 +4623,56 @@ private function get_prio_freq_static_homepage( $links ) {
...
+		 */
+		private function get_prio_freq_static_blogpage( $links ) {
+			$blogpage_id = (int) get_option( 'page_for_posts' );
+			$permalink   = get_permalink( $blogpage_id );
+
+			if ( 0 === $blogpage_id || 'page' !== get_option( 'show_on_front' ) ) {
+				return $links;
+			}
+
+			$blogpage_index = array_search( $permalink, array_column( $links, 'loc' ) ); // phpcs:ignore PHPCompatibility.FunctionUse.NewFunctions.array_columnFound
```
benjaminprojas: I am pretty sure array_column is a PHP 5.5+ feature. We will need to test with our minimum supported version, but I think we should just swap it out.

arnaudbroes: @benjaminprojas correct, but we have a compatibility function for this - https://github.com/awesomemotive/all-in-one-seo-pack/blob/caecc441411a4812f036afe0df72f70668c0b430/inc/compatibility/php-functions.php#L8-L25

benjaminprojas: ok cool

aioseop_class.php
```diff
@@ -5222,81 +5245,67 @@ function display_category_metaboxes( $tax ) {
...
 			$awmp_edit = $_POST['aiosp_edit'];
 		}
+
 		if ( isset( $_POST['nonce-aioseop-edit'] ) ) {
 			$nonce = $_POST['nonce-aioseop-edit'];
 		}
 
 		if ( isset( $awmp_edit ) && ! empty( $awmp_edit ) && wp_verify_nonce( $nonce, 'edit-aioseop-nonce' ) ) {
+
+			if ( ! ( ! empty( $this->options['aiosp_can'] ) ) ) {
```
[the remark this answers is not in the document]
arnaudbroes: My eyes!!!

PR #3167, opened 2020-04-08
Security and performance improvement: rel="noopener noreferrer"

netchild:
## Proposed changes

For all external href with target="_blank" to add
rel="noopener noreferrer". The reason for this patch is performance
and security, you can find more info at
    https://developers.google.com/web/tools/lighthouse/audits/noopener
and
    https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/

## Types of changes

- Improves existing code

## Checklist

- [X] I've created an appropriate title, descriptive of what the PR does.
- [X] If I added, removed or modified any translatable strings, I updated the commonstrings.php file (free repository) accordingly or else opened an issue for it.


wpsmort:
I'm closing this PR because WordPress already adds the noopener and noreferrer attributes to all external links.

netchild:
Can you please provide a reference to wordpress docs or whatever which suggests this is the case? If I look at the source code of the admin page of the plugin in my WP-instance, I see no such behavior:
`<li><a href='https://semperplugins.com/all-in-one-seo-pack-pro-version/?utm_source=WordPress&utm_campaign=liteplugin&utm_medium=admin-menu'><span class='upgrade_menu_link'>Upgrade to Pro</span></a></li></ul></li>`

I also see no such behavior on all my external links in a blog post.

wpsmort:
It was added in this Trac ticket - https://core.trac.wordpress.org/ticket/37941
Here's the function doc - https://developer.wordpress.org/reference/functions/wp_targeted_link_rel/

netchild:
Thanks for the reference, but as you can see in my example from above, the output of a real instance as it arrives in the browser is different. In the example we have target=_blank, but no rel="noopener noreferrer". The WP instance I used to get this example is the most recent released WP version and the most recent released all-in-one-seo-pack plugin without modifications. So the link to the pro-version is not having the properties you suggest it should have.

PR #3207, opened 2023-02-14
Fix for nullable cacheTableName in Database class

powernic:
# Proposed changes
AIOSEO\Plugin\Common\Models\Model
Php 8.1. When calling the `getColumns `method, we get an error, because the table for which we are executing the query is not specified.
`Deprecated: stripos(): Passing null to parameter #1 ($haystack) of type string is deprecated in /wp-content/plugins/all-in-one-seo-pack/app/Common/Utils/Database.php on line 1647`

# Types of changes
Bugfix (non-breaking change which fixes an issue)

### DIFF
--- a/app/Common/Models/Model.php
+++ b/app/Common/Models/Model.php
@@ -452,7 +452,8 @@ class Model implements \JsonSerializable {
 
 			// Let's set the columns that are available by default.
 			$table   = aioseo()->core->db->prefix . $this->table;
-			$results = aioseo()->core->db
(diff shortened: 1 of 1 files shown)


arnaudbroes:
Thanks for this contribution, @powernic!

We have already fixed this notice elsewhere in our 4.3.1 update that is going out on Thursday.

This PR only fixes one place where it happens rather than the cause of it, but it's nevertheless still a good change so I'll go ahead and merge this in.

PR #3212, opened 2023-07-28
Added hooks to change post ID for third party compatibility

mi5t4n:
Added hooks to change the post ID for third-party compatibility.  The compatibility with WooCommerce is hard coded, which doesn't allow room for other plugin compatibility. This has been inspired by two other SEO plugins Yoast Seo and Rank Math. I have included their code links as well.

1. Rank Math: https://github.com/rankmath/seo-by-rank-math/blob/76e87a34fb3a4f325c396251a65d19fd71fac6b5/includes/class-post.php#L97-#L130

2. Yoast : https://github.com/Yoast/wordpress-seo/blob/78df60708b2a56387e63cd42c0d0451a5a116e71/src/helpers/current-page-helper.php#L76-#L96

### DIFF
--- a/app/Common/Traits/Helpers/WpContext.php
+++ b/app/Common/Traits/Helpers/WpContext.php
@@ -172,6 +172,22 @@ trait WpContext {
 	public function getPost( $postId = false ) {
 		$postId = is_a( $postId, 'WP_Post' ) ? $postId->ID : $postId;
 
+		/**
+		 * Filter: Allow to change post ID.
+		 *
+		 * @see https://github.com/rankmath/seo-by-rank-math/blob/76e87a34fb3a4f325c396251a65d19fd71fac6b5/includes/class-post.php#L97-#L130
+		 *
+		 * @since x.x.x
+		 *
+		 * @param int $postId Post ID.
+		 */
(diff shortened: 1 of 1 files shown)


arnaudbroes:
@mi5t4n I'm not necessarily opposed to this change, but do you mind giving us an example of a use-case or third-party plugin that would require this hook?

mi5t4n:
@arnaudbroes We have an LMS plugin called [Masteriyo](https://wordpress.org/plugins/learning-management-system/), where the courses list page is similar to the shop page of WooCommerce. We have set a WP post page as an archive page of courses, due to which the SEO meta tags created by All In One Seo don't render because it thinks it is an archive page. We have a hook to override the post ID, which All In One has done for the WooCommerce shop page. It will help us make our plugin compatible with All In One SEO.

arnaudbroes:
Got it, makes sense. I'll go ahead and add this in for our 4.4.5 update which will be released at the end of this month.

arnaudbroes:
@mi5t4n I've made the change in our private repo. As mentioned above, we'll be pushing this out in 4.4.5.

I named the hook `aioseo_get_post_id` to make more consistent to our codebase and I also moved it down a few lines, but it should still work fine.

I'll go ahead and close this PR now but feel free to let me know if you have any questions.

